PR #1, opened 2014-12-18
Added support for services and clients

esteve:
This adds the structures needed for services in the middleware interface.

### DIFF
--- a/ros_middleware_interface/include/ros_middleware_interface/functions.h
+++ b/ros_middleware_interface/include/ros_middleware_interface/functions.h
@@ -38,9 +38,9 @@ bool receive_response(const ClientHandle& client_handle, void * ros_response);
 
 ServiceHandle create_service(const NodeHandle& node_handle, const rosidl_generator_cpp::ServiceTypeSupportHandle & service_type_support_handle, const char * service_name);
 
-bool take_request(const ServiceHandle& service_handle, void * ros_request);
+bool take_request(const ServiceHandle& service_handle, void * ros_request, void * ros_request_header);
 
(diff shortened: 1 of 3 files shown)


ros_middleware_interface/include/ros_middleware_interface/functions.h
```diff
@@ -27,8 +28,19 @@ GuardConditionHandle create_guard_condition();
...
 void trigger_guard_condition(const GuardConditionHandle& guard_condition_handle);
 
-void wait(SubscriberHandles& subscriber_handles, GuardConditionHandles& guard_condition_handles, bool non_blocking);
+void wait(SubscriberHandles& subscriber_handles, GuardConditionHandles& guard_condition_handles, ServiceHandles& service_handles, bool non_blocking);
 
+ClientHandle create_client(const NodeHandle& node_handle, const rosidl_generator_cpp::ServiceTypeSupportHandle & service_type_support_handle, const char * service_name);
+
+void send_request(const ClientHandle& client_handle, const void * ros_request);
+
+bool receive_response(const ClientHandle& client_handle, void * ros_response, long timeout);
```
dirk-thomas: We will design this API asynchronously for now. So the `timeout` argument should not be necessary.

We have to verify that the following example is still viable:

* user land asynchronous callback gets called by an executor
* then calls a synchronous service call (without having another executor waiting)
* also shouldn't affect "the rest" of the system

esteve: I added a new API function (take_response) that works asynchronously, the executor will take responses only when there is data on the responses data reader.

PR #4, opened 2015-03-06
Add support for take* calls

esteve:
`rmw_take*` functions now take a `bool` argument (taken) to signal if they actually took incoming data.

Connects to ros2/rmw_connext#3

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -75,7 +75,7 @@ rmw_destroy_subscription(rmw_subscription_t * subscription);
 
 RMW_PUBLIC
 rmw_ret_t
-rmw_take(const rmw_subscription_t * subscription, void * ros_message);
+rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * taken);
 
 RMW_PUBLIC
 rmw_client_t *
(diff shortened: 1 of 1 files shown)


wjwwood:
Before we had discussed that instead of a boolean just setting the pointers to `NULL` to indicate that nothing was available to take, leaving the return code to indicate when an error had really occurred. Is there a reason you reintroduced a boolean? I'm not necessarily opposed to having the bool, but it seems you have enough information without it.

esteve:
That would require passing a pointer to a pointer because the related structures (request, response) are created before calling the `rmw_take*` functions. I thought passing a `bool` would make the API less convoluted.

wjwwood:
Ok, makes sense. We need to review all of our memory allocation and ownership, at which point we might decide to do this differently, but right now I think this is the best solution.

+1

dirk-thomas:
If the functions sets the pointer to zero the caller must always keep a second variable to not loose the pointer to a e.g. allocated instance. Also the API could pass a pointer which is const itself but allow only to modify the value inside the instance addressed by the pointer.

Therefore I would consider the extra boolean argument to be the better API.

tfoote:
+1

PR #7, opened 2015-03-11
Add allocators for clients and services

esteve:
Add allocators for clients and services

@dirk-thomas @tfoote @wjwwood


wjwwood:
+1 the changes look ok.

However, this stuff is fairly untested, and some more thought may need to be given to the way I've set these up.

tfoote:
+1

dirk-thomas:
lgtm

esteve:
Thanks.

PR #11, opened 2015-04-28
export type support for rmw implementation

dirk-thomas:
This is necessary to build libraries / executable against messages / services within the same package.

@esteve @tfoote @wjwwood Please review.


wjwwood:
lgtm

esteve:
+1

PR #14, opened 2015-06-18
fix rmw error strings

dirk-thomas:
Without the patch if a library like `rmw_connext_cpp` sets the error string another library like `rclcpp` will not be able to retrieve the error string.

Please review.

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -1,11 +1,15 @@
 cmake_minimum_required(VERSION 2.8.3)
 
-project(rmw NONE)
+project(rmw C)
 
 find_package(ament_cmake REQUIRED)
 
+include_directories(include)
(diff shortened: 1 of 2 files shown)


esteve:
+1

rmw/include/rmw/error_handling.h
```diff
@@ -20,26 +20,18 @@ extern "C"
 {
 #endif
 
-#include <string.h>
-
 #include "macros.h"
 #include "visibility_control.h"
 
-RMW_THREAD_LOCAL char * __rmw_error_string = 0;
+RMW_THREAD_LOCAL char * __rmw_error_string;
```
wjwwood: Shouldn't this be removed from the header all together?

dirk-thomas: Very good point. Done.

PR #15, opened 2015-06-19
Added basic qos for readers and writers

esteve:
Connects to ros2/ros2#66

### DIFF
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -1,0 +1,64 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROS_MIDDLEWARE_INTERFACE_ROS_MIDDLEWARE_INTERFACE_QOS_PROFILES_H_
+#define ROS_MIDDLEWARE_INTERFACE_ROS_MIDDLEWARE_INTERFACE_QOS_PROFILES_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include "types.h"
+
+const static rmw_qos_profile_t rmw_qos_profile_sensor_data =
+{
+  RMW_QOS_POLICY_KEEP_LAST_HISTORY,
+  5,
+  RMW_QOS_POLICY_BEST_EFFORT
+};
+
+const static rmw_qos_profile_t rmw_qos_profile_parameters =
+{
+  RMW_QOS_POLICY_KEEP_LAST_HISTORY,
+  1000,
+  RMW_QOS_POLICY_RELIABLE
+};
+
+const static rmw_qos_profile_t rmw_qos_profile_default =
+{
+  RMW_QOS_POLICY_KEEP_ALL_HISTORY,
+  10,
+  RMW_QOS_POLICY_RELIABLE
+};
+
+const static rmw_qos_profile_t rmw_qos_profile_services_default =
+{
+  RMW_QOS_POLICY_KEEP_ALL_HISTORY,
+  10,
+  RMW_QOS_POLICY_RELIABLE
+};
+
+const static rmw_qos_profile_t rmw_qos_profile_parameter_events =
+{
+  RMW_QOS_POLICY_KEEP_ALL_HISTORY,
+  1000,
+  RMW_QOS_POLICY_RELIABLE
+};
+
+#if __cplusplus
+}
+#endif
+
+#endif  /* ROS_MIDDLEWARE_INTERFACE_ROS_MIDDLEWARE_INTERFACE_QOS_PROFILES_H_ */
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -33,6 +33,7 @@ extern "C"
 #include <rosidl_generator_c/message_type_support.h>
 #include <rosidl_generator_c/service_type_support.h>
 
+#include "qos_profiles.h"
 #include "types.h"
 #include "visibility_control.h"
 
@@ -63,7 +64,7 @@ rmw_create_publisher(
   const rmw_node_t * node,
   const rosidl_message_type_support_t * type_support,
   const char * topic_name,
-  size_t queue_size);
+  const rmw_qos_profile_t & qos_policies = rmw_qos_profile_default);
 
 RMW_PUBLIC
 RMW_WARN_UNUSED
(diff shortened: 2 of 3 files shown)


esteve:
Linux buildfarm:

http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/155/

rmw/include/rmw/types.h
```diff
@@ -94,6 +94,11 @@ typedef struct RMW_PUBLIC rmw_request_id_t
   int64_t sequence_number;
 } rmw_request_id_t;
 
+enum rmw_qos_policy_t {
+  RELIABLE,
+  BEST_EFFORT
```
dirk-thomas: These constants should be prefixed with something like `RMW_QOS_POLICY_`.

rmw/include/rmw/rmw.h
```diff
@@ -52,7 +52,8 @@ rmw_create_publisher(
   const rmw_node_t * node,
   const rosidl_message_type_support_t * type_support,
   const char * topic_name,
-  size_t queue_size);
+  size_t queue_size,
+  const rmw_qos_policy_t qos_policy=RELIABLE);
```
dirk-thomas: `const` should not be used for primitive types like ints.

Same below.

esteve:
Linux buildfarm:

http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/156/

dirk-thomas:
How does that work when two entities use different options? Does it make sense to cover that in a test?

rmw/include/rmw/types.h
```diff
@@ -95,6 +95,60 @@ typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
...
+  .depth = 10,
+  .reliability = RMW_QOS_POLICY_RELIABLE
+};
+
+const static rmw_qos_policy_t rmw_qos_parameter_events =
+{
+  .history = RMW_QOS_POLICY_KEEP_ALL_HISTORY,
+  .depth = 1000,
+  .reliability = RMW_QOS_POLICY_RELIABLE
+};
```
dirk-thomas: Should those policies be defined somewhere else? A separate package seems overkill but at least a separate header (e.g. `policies.h`)?

I am also not sure about all the values. Why a depth of 5 for sensor data? Why is the default "keep all"?

esteve: I'll move the QoS profiles elsewhere, does `qos_profiles.h` sound good? Values are currently a bit arbitrary, I'm going to run tests and see which ones make sense. In any case, I can change the default ones now to something else that you think make more sense.

dirk-thomas: `qos_profiles.h` sounds good to me.

Should the variables then all start with `rmw_qos_profiles_`?

esteve:
Linux buildfarm:

http://ci.ros2.org/job/ros2_batch_ci_linux/116/

rmw/include/rmw/qos_profiles.h
```diff
@@ -0,0 +1,64 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
```
dirk-thomas: Wrong year.

rmw/include/rmw/types.h
```diff
@@ -95,6 +95,25 @@ typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
   int64_t sequence_number;
 } rmw_request_id_t;
 
+enum rmw_qos_reliability_policy_t
```
dirk-thomas: The other enum uses the macro `RMW_PUBLIC_TYPE` - this one does not.

esteve: Fixed.

rmw/include/rmw/qos_profiles.h
esteve: 2014 was a great year! :-)

But fixed.

rmw/include/rmw/qos_profiles.h
```diff
@@ -0,0 +1,64 @@
...
+#define ROS_MIDDLEWARE_INTERFACE_ROS_MIDDLEWARE_INTERFACE_QOS_PROFILES_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include "types.h"
+
+const static rmw_qos_policy_t rmw_qos_profile_sensor_data =
```
dirk-thomas: Does this needs any visibility macros for Windows?

rmw/include/rmw/types.h
```diff
@@ -95,6 +95,25 @@ typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
...
+  RMW_QOS_POLICY_BEST_EFFORT
+};
+
+enum RMW_PUBLIC_TYPE rmw_qos_history_policy_t
+{
+  RMW_QOS_POLICY_KEEP_LAST_HISTORY,
+  RMW_QOS_POLICY_KEEP_ALL_HISTORY
+};
+
+typedef struct RMW_PUBLIC_TYPE rmw_qos_policy_t
```
wjwwood: This is a trivial comment, but does it make more sense for the name to be plural? Maybe `rmw_qos_policies_t`?

wjwwood:
+1

tfoote:
+1

rmw/include/rmw/types.h
esteve: I was actually going to rename this to `rmw_qos_profile_t` because it's a set of QoS policies and to make it consistent with the filename `qos_profiles.h`, thoughts?

rmw/include/rmw/qos_profiles.h
esteve: This PR fails on Windows because of the struct initialization syntax, I'll fix it and check if it needs the visibility macros:

http://ci.ros2.org/job/ros2_batch_ci_windows/101/console

rmw/include/rmw/types.h
wjwwood: Well, the question is: whether or not a profile is any set of QoS's or if it is a particular (or even preset) set of QoS?

If it's the former then I think that's fine, but since the individual enum's are called a `policy`, then I'd say `rmw_qos_policies_t` may be more appropriate.

esteve: I'd say it's the latter, since it also includes the history depth, which is not a QoS policy per se.

dirk-thomas: I would agree with @wjwwood that it kind of is: http://community.rti.com/rti-doc/510/ndds/doc/html/api_cpp/group__DDSHistoryQosModule.html

esteve: @wjwwood was referring to each individual enums named `policy`, whereas depth is defined as a `size_t`. Anyway, if `rmw_qos_policies_t` sounds more natural, that's fine to me.

wjwwood: Either is fine with me, thanks @esteve.

esteve: I renamed all references to `*qos_profile*/*qos_policy*` to `*qos_policies*`.

rmw/include/rmw/rmw.h
```diff
@@ -34,6 +34,7 @@ extern "C"
 #include <rosidl_generator_c/service_type_support.h>
 
 #include "types.h"
+#include "qos_policies.h"
```
dirk-thomas: Should this be one line above for alphabetical order?

dirk-thomas:
lgtm (beside the minor comment)

rmw/include/rmw/types.h
esteve: After going through the code a few times, I'm starting to think that a more appropriate name would either be qos_profile or qos_policy_set

wjwwood: Those are also fine by me, since they both imply plurality.

PR #17, opened 2015-06-23
add flag to ignore local publications

dirk-thomas:
Required for ros2/ros1_bridge#1

Please review.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -68,7 +68,8 @@ rmw_create_subscription(
   const rmw_node_t * node,
   const rosidl_message_type_support_t * type_support,
   const char * topic_name,
-  size_t queue_size);
+  size_t queue_size,
+  bool ignore_local_publications);
 
 RMW_PUBLIC
 rmw_ret_t


tfoote:
+1, though this feels very application specific and I wonder if this is something we should consider going around the standard API for the bridge instead of extending the standard API

wjwwood:
+1. I think this will be a necessary feature to expose to the user. However, I agree that it maybe better exposed to the users through a more generic "filtering" scheme or even provide the necessary tools to the user so the subscriber can do the filtering (caller id).

This particular option name is also somewhat vague, for instance does it mean block communications to the same node/participant? or does that include other nodes in the same process?

This pokes at another issue we have to nail down, is there a direct one to one mapping between ROS nodes and DDS participants. It may be necessary to break with that design to accomplish our intra-process goals.

Do you plan to open a pull request for `rmw_opensplice`, `rcl`, and `rclc`?

At some point I think we'll need to setup a settings struct for each of the primitives. Using a similar patter to DDS, where you get the "default", mutate it, and then pass it to the `create_*` function. Fancier things could be done in `rclcpp` like using a named parameter idiom.

dirk-thomas:
I am currently working on the `rmw_opensplice` PR. I don't plan to touch `rcl` / `rclc` since the bridge does not use them and there is no test coverage for them atm.

wjwwood:
Tested or not we shouldn't let them diverge. I'll do the pull requests for them.

wjwwood:
- `rcl`: https://github.com/ros2/rcl/pull/3
- `rclc`: https://github.com/ros2/rclc/pull/3

Issue #16, opened 2015-06-22
convert rmw_set_error_string to use file and line citations


tfoote:
This will make debugging much easier. 

From discussion on https://github.com/ros2/rmw_connext/pull/43

PR #18, opened 2015-06-29
changes to support proper life cycle of rmw objects

wjwwood:
The most notable change is that I've added the node handle as an argument to the destroy function for publishers and subscribers (it will be needed with server and client as well, but not until other things are fixed).

Otherwise I also added a function to "demangle" c++ symbols so that I can print nicer error messages in the rmw implementations. However, to do this I had to change the "error_handling" file and therefore the library to C++. That's should be ok though since the API remains C only.

Connects ros2/ros2#69

### DIFF
--- a/rmw/include/rmw/impl/config.h
+++ b/rmw/include/rmw/impl/config.h
@@ -1,4 +1,4 @@
-// Copyright 2015 Open Source Robotics Foundation, Inc.
+// Copyright 2014 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,33 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <stdlib.h>
-#include <string.h>
+#ifndef RMW_RMW_IMPL_CONFIG_H_
+#define RMW_RMW_IMPL_CONFIG_H_
 
-#include <rmw/error_handling.h>
+#ifndef RMW_AVOID_MEMORY_ALLOCATION
+// Default to off.
+// TODO(wjwwood): document these configuration flags.
+#define RMW_AVOID_MEMORY_ALLOCATION 0
+#endif
 
-RMW_THREAD_LOCAL char * __rmw_error_string = 0;
+#ifndef RMW_REPORT_ERROR_HANDLING_ERRORS
+// Default reporting of error handling errors based on real-time settings.
+// Turn this off to avoid allocation of memory in error handling funcs.
+#define RMW_REPORT_ERROR_HANDLING_ERRORS 1
+#endif
 
-void
-rmw_set_error_string(const char * error_string)
-{
-  if (__rmw_error_string) {
-    free(__rmw_error_string);
-  }
-  __rmw_error_string = strdup(error_string);
-}
-
-const char *
-rmw_get_error_string()
-{
-  return __rmw_error_string;
-}
-
-const char *
-rmw_get_error_string_safe()
-{
-  if (!__rmw_error_string) {
-    return "error string not set";
-  }
-  return __rmw_error_string;
-}
+#endif  /* RMW_RMW_IMPL_CONFIG_H_ */
--- a/rmw/include/rmw/impl/cpp/macros.hpp
+++ b/rmw/include/rmw/impl/cpp/macros.hpp
@@ -1,0 +1,100 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RMW_RMW_IMPL_CPP_MACROS_HPP_
+#define RMW_RMW_IMPL_CPP_MACROS_HPP_
+
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include <rmw/error_handling.h>
+#include <rmw/impl/config.h>  // For RMW_AVOID_MEMORY_ALLOCATION
+#include <rmw/impl/cpp/demangle.hpp>  // For demangle.
+
+// *INDENT-OFF* (prevent uncrustify from using four space indention here)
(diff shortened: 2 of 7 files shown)


esteve:
+1

tfoote:
+1

jacquelinekay:
Linux Jenkins build failure, also experienced on my machine:

http://ci.ros2.org/job/ros2_batch_ci_linux/12/console

Console output:
```
/home/rosbuild/ci_scripts/workspace/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/functions.cpp: In function 'rmw_guard_condition_t* rmw_create_guard_condition()':
/home/rosbuild/ci_scripts/workspace/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/functions.cpp:42:64: error: must #include <typeinfo> before using typeid
     std::string("caught C++ exception ") + rmw_demangle(typeid(exception).name()) + \
                                                                ^
/home/rosbuild/ci_scripts/workspace/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/functions.cpp:634:3: note: in expansion of macro 'TRY_PLACEMENT_NEW'
   TRY_PLACEMENT_NEW(guard_condition->data, guard_condition->data, goto fail, DDS::GuardCondition)
   ^
/home/rosbuild/ci_scripts/workspace/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/functions.cpp: In function 'rmw_ret_t rmw_destroy_guard_condition(rmw_guard_condition_t*)':
/home/rosbuild/ci_scripts/workspace/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/functions.cpp:56:25: error: must #include <typeinfo> before using typeid
     rmw_demangle(typeid(exception).name()) + ": " + exception.what() \
                         ^
/home/rosbuild/ci_scripts/workspace/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/functions.cpp:658:3: note: in expansion of macro 'TRY_DESTRUCTOR'
   TRY_DESTRUCTOR(dds_guard_condition->, GuardCondition, result = RMW_RET_ERROR)
   ^
make[2]: *** [CMakeFiles/rmw_opensplice_cpp.dir/src/functions.cpp.o] Error 1
make[1]: *** [CMakeFiles/rmw_opensplice_cpp.dir/all] Error 2
make: *** [all] Error 2
```

wjwwood:
Thanks, I was just getting around to running the CI jobs. I'll the pr's with fixes and start new CI jobs.

rmw/include/rmw/impl/cpp/macros.hpp
```diff
@@ -0,0 +1,62 @@
...
+#define TRY_PLACEMENT_NEW(Destination, BufferForNew, FailureAction, Type, ...) try { \
+  Destination = new(BufferForNew) Type(__VA_ARGS__); \
+} catch(const std::exception & exception) { \
+  rmw_set_error_string(( \
+    std::string("caught C++ exception ") + rmw_demangle(typeid(exception).name()) + \
+    " constructing Type: " + exception.what() \
+  ).c_str()); \
+  FailureAction; \
+} catch(...) { \
+  rmw_set_error_string("caught unknown C++ exception constructing Type"); \
```
dirk-thomas: Should this put the exception type and `Type` into the error string?

wjwwood: Yes, I will fix that.

wjwwood:
Update:

I Change the rmw library back to pure C and I moved the demangle function into a template function in a header, `rmw/impl/cpp/demangle.hpp`. This was necessary because I needed to have a C++ return type (std::string).

I also added a new macro to make the comparison between rmw implementation ids consistent.

I also introduced a new overridable definition, `RMW_AVOID_MEMORY_ALLOCATION`, which is mostly for the benefit of some uncommitted changes I have which introduce an option to have full length error strings with allocation or a limit on error string length with truncation if it exceeds that limit. For this pr, I use it because demangle will use allocation, so it's disabled if we want to avoid that. `RMW_AVOID_MEMORY_ALLOCATION` defaults to off.

rmw/src/error_handling.c
```diff
@@ -21,6 +22,9 @@ RMW_THREAD_LOCAL char * __rmw_error_string = 0;
 void
 rmw_set_error_string(const char * error_string)
 {
+  if (__rmw_error_string) {
+    free(__rmw_error_string);
+  }
```
dirk-thomas: Should we add a `rmw_clear_error_string` function which can be called during shutdown to ease memory debugging?

wjwwood: I have that in my uncommitted changes, I'll probably put those into a pull request to address https://github.com/ros2/rmw/issues/16

rmw/include/rmw/impl/cpp/macros.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include <rmw/error_handling.h>
+#include <rmw/impl/config.h>  // For RMW_AVOID_MEMORY_ALLOCATION
+#include <rmw/impl/cpp/demangle.hpp>  // For demangle.
+
+// *INDENT-OFF* (prevent uncrustify from using four space indention here)
+#define TRY_PLACEMENT_NEW(Destination, BufferForNew, FailureAction, Type, ...) try { \
```
dirk-thomas: Should all these macros be prefixed with `RMW_`?

wjwwood: I suppose I could do that now that they're in a header, but on the other hand, this should only ever be included by cpp files and the header has an include guard.

I'll think about doing it.

rmw/include/rmw/impl/config.h
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RMW_RMW_IMPL_CONFIG_H_
+#define RMW_RMW_IMPL_CONFIG_H_
+
+#ifndef RMW_AVOID_MEMORY_ALLOCATION
+// Default to off.
+// TODO(wjwwood): document these configuration flags.
+#define RMW_AVOID_MEMORY_ALLOCATION 0
```
dirk-thomas: How do we intend to deploy this? Or let users use it? It seems like a high barrier if the user needs to rebuild the rmw_implementation for that.

wjwwood: For now that's the most efficient and sure fire way to enforce the affects of this macro. I don't expect it's necessary since it really only affects functions used in error handling. If it is swimmingly used we could make it more complicated and provide an argument to rmw_init or use some other mechanism to change the behavior.

wjwwood: Commonly* (phone autocorrect)

jacquelinekay:
+1

dirk-thomas:
+1 to squash and merge.

PR #21, opened 2015-07-21
add set error msg macro and improve error api

wjwwood:
I've replaced the usage of `rmw_set_error_string` with a macro `RMW_SET_ERROR_MSG` which sets the error state with a new function `rmw_set_error_state(msg, file, line)`. I've tried to use the term error message (msg) when referring to the message the user sets and error string when referring to the formatted error message which includes the file and line number.

I also added some documentation to the error api functions and added some utility functions. I had started an implementation that avoided all memory allocation by using static error storage and truncating messages and file paths if necessary, but I just pushed that to my fork for now. We can revive that if we decide we need alloc free error setting and printing.

Connects to #16

### DIFF
--- a/rmw/include/rmw/error_handling.h
+++ b/rmw/include/rmw/error_handling.h
@@ -20,20 +20,92 @@ extern "C"
 {
 #endif
 
+#include <stddef.h>
+
 #include "macros.h"
 #include "visibility_control.h"
 
+/// Struct which encapsulates the error state set by RMW_SET_ERROR_MSG().
+typedef struct rmw_error_state_t {
+  const char * message;
+  const char * file;
+  size_t line_number;
+} rmw_error_state_t;
+
+/// Set the error message, the file it occurred in, and the line on which it occurred.
+/**
+ * This is not meant to be used directly, but instead via the RMW_SET_ERROR_MSG(msg) macro.
+ *
+ * The error_msg parameter is copied into the interal error storage and must be null terminated.
+ * The file parameter is not copied, but instead is assumed to be a global as it should be set to
+ * the __FILE__ preprocessor literal when used with the RMW_SET_ERROR_MSG macro.
+ * It should also be null terminated.
+ *
+ * \param error_msg The error message to set.
+ * \param file The path to the file in which the error occurred.
+ * \param line_number The line number on which the error occurred.
+ */
 RMW_PUBLIC
 void
-rmw_set_error_string(const char * error_string);
+rmw_set_error_state(const char * error_msg, const char * file, size_t line_number);
 
+/// Set the error message, automatically appending the current file and line number.
+/**
+ * If an error message was previously set, and rmw_reset_error was not called
+ * since, and this library was built with RMW_REPORT_ERROR_HANDLING_ERRORS set
+ * on, then previously set error message will be printed to stderr.
+ * Error state storage is thread local and so all error related functions are also thread local.
+ *
+ * \param msg The error message to be set.
+ */
+#define RMW_SET_ERROR_MSG(msg) rmw_set_error_state(msg, __FILE__, __LINE__);
+
+/// Return true if the error has been set and has not been reset since, otherwise false.
+RMW_PUBLIC
+bool
+rmw_error_is_set();
+
+/// Return a struct with the error message and the file and line number on which is was set.
+/**
+ * The returned pointer will be NULL if no error has been set in this thread.
+ *
+ * The returned pointer is valid until RMW_SET_ERROR_MSG, rmw_set_error_state, or rmw_reset_error
+ * are called in the same thread.
+ *
+ * \return A pointer to the current error state struct.
+ */
+RMW_PUBLIC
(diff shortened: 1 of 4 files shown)


rmw/include/rmw/error_handling.h
```diff
@@ -23,18 +23,76 @@ extern "C"
 #include "macros.h"
 #include "visibility_control.h"
 
+/// Struct which encapsulates the error state set by RMW_SET_ERROR_MSG().
```
dirk-thomas: Why does the comment start with three slashes?

rmw/include/rmw/error_handling.h
```diff
@@ -23,18 +23,76 @@ extern "C"
...
+typedef struct rmw_error_state_t {
+  const char * message;
+  const char * file;
+  int line_number;
+} rmw_error_state_t;
+
+/// Sets the error message, the file it occurred in, and the line on which it occurred.
```
dirk-thomas: Why is this line not part of the docblock?

rmw/include/rmw/error_handling.h
```diff
@@ -23,18 +23,76 @@ extern "C"
 #include "macros.h"
 #include "visibility_control.h"
 
+/// Struct which encapsulates the error state set by RMW_SET_ERROR_MSG().
+typedef struct rmw_error_state_t {
+  const char * message;
+  const char * file;
+  int line_number;
```
dirk-thomas: Is `__LINE__` defined as `int`? Otherwise would something unsigned make more sense, e.g. `size_t`?

rmw/include/rmw/error_handling.h
```diff
@@ -23,18 +23,76 @@ extern "C"
...
+} rmw_error_state_t;
+
+/// Sets the error message, the file it occurred in, and the line on which it occurred.
+/**
+ * This is not meant to be used directly, but instead via the RMW_SET_ERROR_MSG(msg) macro.
+ *
+ * The error_msg parameter is copied into the interal error storage and must be null terminated.
+ * The file parameter is not copied, but instead is assumed to be a global as it should be set to
+ * the __FILE__ preprocessor literal when used with the RMW_SET_ERROR_MSG macro.
+ * It should also be null terminated.
```
dirk-thomas: Should the docblock also document the parameters and the return value?

rmw/include/rmw/error_handling.h
wjwwood: That's the "C++ style" brief for doxygen: http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html

That's is to say, not javadoc style nor qt style. I guess we don't have a set away we've choosen to do it so I took the c++ way.

rmw/include/rmw/error_handling.h
wjwwood: It's the brief. I guess the Python way? I'm not sure it matters that much, as long as it makes sense when you read it. I'll drop the pluralization in this case I guess.

rmw/include/rmw/error_handling.h
wjwwood: Yeah I guess so, I stopped because I wasn't sure which style we wanted to use. I was most concerned about getting some of those notes down out of my head.

rmw/include/rmw/error_handling.h
wjwwood: It probably should be a size_t, I'll update it.

rmw/src/error_handling.c
```diff
@@ -12,33 +12,138 @@
...
     return "error string not set";
   }
+  if (!__rmw_error_string) {
+    format_error_string(__rmw_error_string, __rmw_error_state);
+  }
   return __rmw_error_string;
 }
+
+void
+rmw_reset_error_string()
```
jacquelinekay: Are there plans to use this function anywhere? I noticed that it never gets called.

jacquelinekay:
Looks good to me.

rmw/src/error_handling.c
wjwwood: If you ever wanted to catch an error, handle it, and then continue you would need to. 

Also this is a good catch I should call it after checking it in set error state.

wjwwood:
Ok, comments should be addressed now.

rmw/src/error_handling.c
```diff
@@ -12,33 +12,139 @@
...
 rmw_get_error_string_safe()
 {
+  if (!rmw_error_is_set()) {
+    return "error not set";
+  }
   if (!__rmw_error_string) {
-    return "error string not set";
+    format_error_string(__rmw_error_string, __rmw_error_state);
   }
   return __rmw_error_string;
```
dirk-thomas: Can this just call `rmw_get_error_string`?

dirk-thomas:
+1 aside the last comment.

jacquelinekay:
+1

rmw/src/error_handling.c
wjwwood: Yes, updated.

PR #22, opened 2015-07-21
warning if caller does not use result of rmw function

dirk-thomas:
To prevent future regressions in using the rmw API.


wjwwood:
Neat. +1, there seems to be a Windows equivalent: http://stackoverflow.com/a/22755542/671658

dirk-thomas:
Good point. Added the Windows case in 64130ce62ca78b0a87603f06fbea2690450f089f (without being able to test it for now).

PR #25, opened 2015-07-24
introduce time struct, add timeout duration to rmw_wait

jacquelinekay:
Connects to ros2/ros2#73

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -178,7 +178,7 @@ rmw_wait(
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
-  bool non_blocking);
+  rmw_time_t * wait_timeout);
 
 #if __cplusplus
 }
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -28,6 +28,7 @@ extern "C"
 typedef int rmw_ret_t;
 #define RMW_RET_OK 0
 #define RMW_RET_ERROR 1
+#define RMW_RET_TIMEOUT 2
 
 typedef struct RMW_PUBLIC_TYPE rmw_node_t
 {
@@ -94,6 +95,12 @@ typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
   int8_t writer_guid[16];
   int64_t sequence_number;
 } rmw_request_id_t;
+
+typedef struct RMW_PUBLIC_TYPE rmw_time_t
+{
+  uint64_t sec;
+  uint64_t nsec;
+} rmw_time_t;
 
 #if __cplusplus
 }


rmw/include/rmw/rmw.h
```diff
@@ -178,7 +178,7 @@ rmw_wait(
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
-  bool non_blocking);
+  rmw_time_t * wait_timeout);
```
esteve: I think `wait_timeout` doesn't need to be a pointer here, we can just pass copies of `rmw_time_t`

rmw/include/rmw/types.h
```diff
@@ -19,7 +19,6 @@
 extern "C"
 {
 #endif
-
```
wjwwood: nitpick: leave the space, it matches the other files.

rmw/include/rmw/rmw.h
wjwwood: +1

wjwwood:
+1

rmw/include/rmw/types.h
```diff
@@ -94,6 +93,12 @@ typedef struct RMW_PUBLIC rmw_request_id_t
   int64_t sequence_number;
 } rmw_request_id_t;
 
+typedef struct RMW_PUBLIC rmw_time_t
+{
+  uint64_t sec;
```
jacquelinekay: Since negative values will be used to represent blocking indefinitely (or nonblocking), I could change this struct to signed integers. Alternatively, I could check for a negative `duration` in rclcpp and pass a `bool` that stores "block indefinitely" or "nonblocking" to `rmw_wait`, along with the timeout.

wjwwood: So, I looked at how `select()` works and it takes the struct as a pointer and a NULL pointer is block forever. So maybe we should do it that way. Changing this to a signed number is not a great option in my opinion. The bool is probably my next favorite option.

wjwwood: Also we could instead call this the `rmw_timeout_t` and include a bool as part of the struct which indicates forever.

dirk-thomas: I like the idea to be similar to `select()`.

rmw/include/rmw/rmw.h
```diff
@@ -18,6 +18,7 @@
 #if __cplusplus
 extern "C"
 {
+
```
dirk-thomas: ?

rmw/include/rmw/types.h
```diff
@@ -95,6 +96,12 @@ typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
   int64_t sequence_number;
 } rmw_request_id_t;
 
+typedef struct RMW_PUBLIC rmw_time_t
```
dirk-thomas: Wrong macro, must be `RMW_PUBLIC_TYPE`.

jacquelinekay: Is this the cause of the warning in: http://ci.ros2.org/job/ros2_batch_ci_linux/109/warnings17Result/new/

dirk-thomas: Yes, a type needs to be exported different then a function on Windows.

dirk-thomas:
+1

esteve:
+1

tfoote:
+1 squashed

Issue #6, opened 2015-03-11
fixup include guard names


wjwwood:
Some of the include guards are still using `ROS_MIDDLEWARE_INTERFACE` rather than `RMW`.

PR #30, opened 2015-08-17
Fixed up include guard names to RMW

spaghetti-:
connects to #6

### DIFF
--- a/rmw/include/rmw/allocators.h
+++ b/rmw/include/rmw/allocators.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef ROS_MIDDLEWARE_INTERFACE_ROS_MIDDLEWARE_INTERFACE_ALLOCATORS_H_
-#define ROS_MIDDLEWARE_INTERFACE_ROS_MIDDLEWARE_INTERFACE_ALLOCATORS_H_
+#ifndef RMW_RMW_ALLOCATORS_H_
+#define RMW_RMW_ALLOCATORS_H_
 
 #if __cplusplus
 extern "C"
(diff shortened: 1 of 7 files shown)


dirk-thomas:
Thank you for creating the PR for this. We have not yet decided on a naming scheme for the include guards therefore we will wait merging this until then. Also the current name contains `RMW` twice which does not map to the actual namespace.

spaghetti-:
No problem, I followed the existing RMW header guard format that was found in certain headers.

wjwwood:
For context, the reason `RMW` is there twice is because it is following the Google Style:

https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#The__define_Guard

Where they define the guard to be `<PROJECT>_<PATH>_<FILE>_H_`. Since the the project name and the path contain `RMW` that's why it is there twice. This sort of makes since, because it is conceivable that you can have different headers, with the same path, in different packages. It seems unlikely that this would be needed, but it is more conceptually correct.

tfoote:
Reviewing this it looks to follow our naming convention correctly and resolves #6 and makes all the include guards consistent. Moving to in review. 

+1 for merge

wjwwood:
+1 from me as well.

dirk-thomas:
The proposed pattern does not prevent collisions. Two packages can have a header in their own namespace which ends up using the same include guard. That is imo a no-go. Therefore -1.

wjwwood:
@dirk-thomas I +1 this because it fixes the names, not because it fixes the include guards in general. That's a different mass bug we have that needs to be fixed across all of the packages and noted in the development style guide.

tfoote:
@dirk-thomas  That's a bigger refactor design decision we should consider making. But for now I think we should take this small cleanup commit to homogenize a few remaining files for our current standard as ticketed in #6.

dirk-thomas:
I don't see much use in changing the include guards to something which we don't think is the way we want them to be. Especially since we currently use various different styles throughout the code base. But if you want to merge this please feel free to do so.

wjwwood:
This change is orthogonal to the format of the include guards. It's fixing something I missed when renaming the package to `rmw` from `ros_middleware_interface`.

PR #31, opened 2015-08-17
Accept QoS profiles via an XML file

esteve:
Connect to ros2/ros2#86


mikaelarguedas:
Decided not to provide this feature because of conflicting vendor definition. We may offer a design spec to do that in vendor agnostic way in the future.

PR #32, opened 2015-08-19
Added system default QoS profile

esteve:
This PR adds support for leaving up to the underlying implementation to use the system default QoS settings.

Connects to ros2/ros2#93


tfoote:
+1

PR #34, opened 2015-08-24
remove shared flag from rmw library declaration

codebot:
by removing the SHARED keyword from the add_library call, we can set it externally and thus allow the cmake machinery to select between shared or static builds.

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -7,7 +7,7 @@ find_package(ament_cmake REQUIRED)
 include(cmake/configure_rmw_library.cmake)
 
 include_directories(include)
-add_library(${PROJECT_NAME} SHARED "src/error_handling.c")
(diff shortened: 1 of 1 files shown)


wjwwood:
Will we need a corresponding change to ament in order to keep the default behavior `SHARED`?

dirk-thomas:
It does currently not set anything when invoking CMake.

esteve:
I implemented the behavior that @codebot described in https://github.com/ament/ament_tools/pull/128 and removed the `SHARED` keyword from `add_library(...)` in https://github.com/ros2/rcl/pull/93 https://github.com/ros2/rmw/pull/81 https://github.com/ros2/rosidl/pull/183 and https://github.com/eProsima/Fast-CDR/pull/8

dirk-thomas:
Superseded by #98.

PR #35, opened 2015-08-24
add support for publisher gid's in rmw

wjwwood:
I am opening this pull request, but it is only part of what is needed for ros2/ros2#91.

Basically it introduces the notion of a GID for publishers in rmw, encapsulated in the `rmw_gid_t`. The `rmw_gid_t` contains only an opaque void pointer, which the implementation can use to compare the GID's in an implementation specific way. You can get the GID of a `rmw_publisher_t` with `rmw_create_gid_from_publisher` and you can get a GID for the sending publisher via `rmw_take_with_gid`. And then you can compare GID's with `rmw_compare_gids_equal`.

Later I can use this to compare the GID of the sender for incoming inter process messages with publishers which are connected via intra process as well and filter those out.

Other ideas for usage and for testing:

- One subscriber is receiving messages from multiple publishers, figure out how many messages have come from each publisher in the subscription callback.
- One publisher and subscription in the same node, you could use this to self filter, ignoring messages from the local publisher.

Please review these changes to the API. I've got an implementation for OpenSplice and I'm starting on Connext/Connext Dynamic. Maybe we can parallelize on those implementations.

Connects to ros2/ros2#91

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -98,6 +98,15 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * tak
 
 RMW_PUBLIC
 RMW_WARN_UNUSED
+rmw_ret_t
+rmw_take_with_info(
+  const rmw_subscription_t * subscription,
+  void * ros_message,
+  bool * taken,
+  rmw_message_info_t * message_info);
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
 rmw_client_t *
 rmw_create_client(
   const rmw_node_t * node,
@@ -209,6 +218,16 @@ rmw_count_subscribers(
   const rmw_node_t * node,
   const char * topic_name,
   size_t * count);
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_get_gid_for_publisher(const rmw_publisher_t * publisher, rmw_gid_t * gid);
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_compare_gids_equal(const rmw_gid_t * gid1, const rmw_gid_t * gid2, bool * result);
 
 #if __cplusplus
 }
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -30,6 +30,10 @@ typedef int rmw_ret_t;
 #define RMW_RET_ERROR 1
 #define RMW_RET_TIMEOUT 2
 
+// 24 bytes is the most memory needed to represent the GID by any current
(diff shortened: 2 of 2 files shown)


dirk-thomas:
Instead of having separate create / destroy functions we could extend `rmw_publisher_t` and add a field of type `rmw_gid_t`.

And the existing `rmw_take` function could have an optional argument after `taken` to pass in a `rmw_gid_t` pointer to be filled.

wjwwood:
I tried that originally, but you still need a clean up function because what the `void *` inside of `rmw_gid_t` points to may need to be free'ed. Since you need to clean them up in an implementation specific way, I found it easier to just have them completely managed with create and destroy.

dirk-thomas:
The `destroy_publisher` function could do that cleanup. The life time of a gid is anyway closely tied to the publisher it identifies.

wjwwood:
That's true, but it's not true for the GID returned by take. That one is not tied to any publisher.

rmw/include/rmw/rmw.h
```diff
@@ -210,6 +220,17 @@ rmw_count_subscribers(
   const char * topic_name,
   size_t * count);
 
+// Note: the gid's lifetime is tied to the publisher.
```
jacquelinekay: Maybe also note that the return value could be null if publisher pointer is null or invalid, or if the the publisher gid wasn't set

wjwwood: Yeah, all of these functions need comments like this. I chose to add just this one because it was particularly unclear. For example, all of the `rmw_create_*` functions will behave the same way. I'd like to wait until we do all of them. It's something I want to do during the next week of documentation.

jacquelinekay: OK, sounds good. Are you going to hold off on implementing Connext until after the release?

wjwwood: No, we need to implement Connext first.

rmw/include/rmw/types.h
```diff
@@ -30,6 +30,8 @@ typedef int rmw_ret_t;
 #define RMW_RET_ERROR 1
 #define RMW_RET_TIMEOUT 2
 
+#define RMW_GID_STORAGE_SIZE 24
```
jacquelinekay: Can you add a comment about where you got the bound on the GID size, before we forget?

dirk-thomas:
+1

jacquelinekay:
Besides a small comment, +1

tfoote:
+1

rmw/include/rmw/types.h
wjwwood: 8b3b89e

wjwwood:
Comments addressed and commits squashed.

PR #36, opened 2015-08-27
add fram_intra_process bool to message_info

wjwwood:
Connects to ros2/demos#12

### DIFF
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -20,6 +20,7 @@ extern "C"
 {
 #endif
 
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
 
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/types.h
```diff
@@ -144,6 +144,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
   // const rmw_time_t received_timestamp;
   rmw_gid_t publisher_gid;
+  bool from_intra_process;
```
jacquelinekay: rmw is supposed to have C headers, so this should be an int, not a bool, right?

wjwwood: C has bool: http://en.cppreference.com/w/c/types/boolean

I should include `#include <stdbool.h>` though. I'll add that.

jacquelinekay:
+1

tfoote:
+1

esteve:
+1

PR #37, opened 2015-09-04
fix/use snprintf on all OS

wjwwood:
Currently the Window specific branch causes an error, because of how I was originally using their special version of `snprintf`, but while doing research I found that VS2015 now provides a proper version of `snprintf`, so we can drop the `#ifdef _WIN32` in this case.

CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/323/
- http://ci.ros2.org/job/ros2_batch_ci_osx/253/
- http://ci.ros2.org/job/ros2_batch_ci_windows/379/

### DIFF
--- a/rmw/src/error_handling.c
+++ b/rmw/src/error_handling.c
@@ -23,12 +23,6 @@
 
 #define SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), sizeof(msg), stderr)
 
-#ifndef _WIN32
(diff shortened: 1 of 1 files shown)


esteve:
+1

wjwwood:
For reference: http://blogs.msdn.com/b/vcblog/archive/2014/06/03/visual-studio-14-ctp.aspx

wjwwood:
CI all build, and tests seem to be within normal behavior, and no new warnings. Merging.

dirk-thomas:
+1 (even if the PR never made it to "in review").

PR #39, opened 2015-09-10
Switch services QoS policy from KEEP_ALL to KEEP_LAST

jacquelinekay:
Based on a discussion in https://github.com/ros2/rmw_opensplice/pull/82. May be subject to change after more discussion/experimentation.

sort of connected to ros2/rmw_connext#96


esteve:
+1

dirk-thomas:
LGTM.

We really need to cover these profiles with detailed rational in the article before we forget about the whys.

PR #41, opened 2015-09-23
Added durability settings

esteve:
This PR adds a new QoS policy (durability) with three settings:

- leave it to the system default
- use transient local (needed for services)
- use volatile


esteve:
CI job (including the tests for the services from the `test-services` branch):

http://ci.ros2.org/job/ros2_batch_ci_linux/383/

I haven't run the CI job for Windows or Mac OSX because I know they will fail until this is merged:

https://github.com/ros2/rmw_opensplice/pull/83

jacquelinekay:
really, even the OSX jobs will fail without ros2/rmw_opensplice#83? The PR description makes it sounds like it's only an issue on Windows.

esteve:
I triggered a CI job for OSX just in case:

http://ci.ros2.org/job/ros2_batch_ci_osx/282/

rmw/include/rmw/types.h
```diff
@@ -121,11 +121,18 @@ enum RMW_PUBLIC_TYPE rmw_qos_history_policy_t
   RMW_QOS_POLICY_KEEP_ALL_HISTORY
 };
 
+enum RMW_PUBLIC_TYPE rmw_qos_durability_policy_t
+{
+  RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT,
+  RMW_QOS_POLICY_TRANSIENT_LOCAL_DURABILITY
```
jacquelinekay: Why not also expose VOLATILE durability (just in case)?

esteve: Because I just wanted to get this merged as soon as I can so I can finalize the services stuff on Windows :smile:  Anyway, I guess I can add VOLATILE too.

esteve:
CI jobs with volatile added:

http://ci.ros2.org/job/ros2_batch_ci_linux/384/
http://ci.ros2.org/job/ros2_batch_ci_osx/283/
http://ci.ros2.org/job/ros2_batch_ci_windows/427/

esteve:
The requester/replier test now fails for RTI Connext (same request received multiple times). It passed when this branch didn't have the logic for volatile qos settings, which is actually not used at all, so I don't know where's the issue, why it passed before and now it doesn't.

esteve:
These three CI jobs include the code for handling the read condition separately:

http://ci.ros2.org/job/ros2_batch_ci_linux/385/
http://ci.ros2.org/job/ros2_batch_ci_osx/284/
http://ci.ros2.org/job/ros2_batch_ci_windows/428/

the OSX slave passes, but I don't think it's because of the read condition code, the Linux slave doesn't. So I don't know if the test is wrong, the code here is wrong or if there's something else. When I only changed the Requester's DataWriter and the Replier's DataReader to use transient QoS, tests seemed to pass (though it shouldn't affect), I wonder if using the same QoS settings might be the cause.

dirk-thomas:
For the failing service test have you tried to add a simple delay between starting the server and the client? Without any need to change DDS QoS options.

esteve:
I triggered three more jobs with the QoS settings tweaked and they are still the same:

http://ci.ros2.org/job/ros2_batch_ci_linux/386/
http://ci.ros2.org/job/ros2_batch_ci_osx/285/
http://ci.ros2.org/job/ros2_batch_ci_windows/431/

So I triggered new ones that add a delay between the service and the client:

http://ci.ros2.org/job/ros2_batch_ci_linux/387/
http://ci.ros2.org/job/ros2_batch_ci_osx/286/
http://ci.ros2.org/job/ros2_batch_ci_windows/432/

I can't reproduce the failing tests locally, so I'm relying on the buildfarm.

esteve:
Forget about the latest set of CI jobs, it contains a compiling error.

esteve:
New CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/390/
http://ci.ros2.org/job/ros2_batch_ci_osx/288/
http://ci.ros2.org/job/ros2_batch_ci_windows/434/

esteve:
The test with a delay of 2 seconds now pass. Linux and OSX jobs are green.

wjwwood:
@esteve I can't find where you introduced the delay, can you link to that? (I'm just curious where it was needed)

esteve:
@wjwwood https://github.com/ros2/system_tests/blob/22b45c0ed41f78a2d77e2e7991498e0bf3fd70eb/test_communication/test/test_requester.cpp#L106

This is actually a copy of `test-services` branch pushed as `transient-qos-services` so the farm groups the branches together as part of the same set.

esteve:
Can this be merged? The changes in this PR passes the tests in `test-services` the current testsuite does not fail.

dirk-thomas:
I assume you refer to:
* this PR
* ros2/rmw_connext#101
* ros2/rmw_opensplice#84

The connext PR still has some comments pending, no?

esteve:
The connext branch had a commit that was added by mistake. New CI jobs just to make sure that the latest changes are ok:

http://ci.ros2.org/job/ros2_batch_ci_linux/393/
http://ci.ros2.org/job/ros2_batch_ci_osx/293/
http://ci.ros2.org/job/ros2_batch_ci_windows/440/

dirk-thomas:
+1

esteve:
After the latest CI run, Windows fails (as usual), OSX fails too and Linux passes.

The tests include the 2 second delay in the requester code to give some time for the replier to start up.

I removed the forced volatile QoS, and although the tests passed with it, in any case shouldn't affect because it was only applied to the requester's reader and the replier's writer.

At this point I just give up, I don't know if it's an issue in the test or in the code. Most likely the latter.

esteve:
This is the test that failed on OSX:

http://ci.ros2.org/job/ros2_batch_ci_osx/293/testReport/%28root%29/test_requester_replier_cpp__empty__rmw_opensplice_cpp/test_requester_replier/

esteve:
Since the failing test also happens in master, I think this and the related branches are okay to merge.

esteve:
CI jobs after https://github.com/ros2/system_tests/pull/51 merged:

http://ci.ros2.org/job/ros2_batch_ci_linux/410/
http://ci.ros2.org/job/ros2_batch_ci_osx/307/
http://ci.ros2.org/job/ros2_batch_ci_windows/456

wjwwood:
@esteve If you're talking about http://ci.ros2.org/job/ros2_batch_ci_osx/293/testReport/%28root%29/test_requester_replier_cpp__empty__rmw_opensplice_cpp/test_requester_replier/, that flaky test should be fixed since yesterday. I'd recommend rebasing (only if you changed `system_tests`) and running the CI again.

wjwwood:
Beat me to it.

esteve:
@wjwwood :smile:

esteve:
OSX passes. Windows fails. And Linux too, but only for this test:

http://ci.ros2.org/job/ros2_batch_ci_linux/410/testReport/junit/%28root%29/test_rclcpp/gtest_intra_process__rmw_connext_dynamic_cpp_gtest_missing_result/

AFAIK intraprocess does not use QoS, but I could be wrong. @wjwwood is that correct?

wjwwood:
> AFAIK intraprocess does not use QoS, but I could be wrong.

Which QoS? It certainly is affected by them, but I don't know what QoS setting you're talking about.

esteve:
@wjwwood durability. The default QoS for publishers and subscribers leaves durability up to the system and doesn't enforce anything and shouldn't affect the current tests:

https://github.com/ros2/rmw/blob/transient-qos-services/rmw/include/rmw/qos_profiles.h#L46

wjwwood:
If the setting didn't change then it shouldn't have an effect, but the intra process will be affected by durability since it inherits the QoS settings from the matching inter process topic.

esteve:
Latest CI run:

http://ci.ros2.org/job/ros2_batch_ci_linux/415/ (green)
http://ci.ros2.org/job/ros2_batch_ci_osx/323/ (green)
http://ci.ros2.org/job/ros2_batch_ci_windows/463/ (pending)

jacquelinekay:
+1

PR #42, opened 2015-09-30
use memset in rmw_allocate

jacquelinekay:
In `rmw_allocate`, memset the allocated pointer with zeros.

The rationale for this pull request is:

1) In the general case, initializing malloc'd memory is considered better practice.
2) In real-time computing, accessing the malloc'd memory will map the memory into RAM and the memory management unit, so that accessing the memory will not cause a pagefault later.

### DIFF
--- a/rmw/include/rmw/allocators.h
+++ b/rmw/include/rmw/allocators.h
@@ -22,6 +22,7 @@ extern "C"
 
 #include <stdlib.h>
 
+#include "string.h"
 #include "types.h"
 #include "visibility_control.h"
 
@@ -30,7 +31,11 @@ void *
 rmw_allocate(size_t size)
 {
   // Could be overridden with a general purpose allocator
-  return malloc(size);
+  void * ptr = malloc(size);
+  if (ptr) {
+    memset(ptr, 0, size);
+  }
+  return ptr;
 }
 
 RMW_LOCAL


rmw/include/rmw/allocators.h
```diff
@@ -30,7 +31,9 @@ void *
 rmw_allocate(size_t size)
 {
   // Could be overridden with a general purpose allocator
-  return malloc(size);
+  void * ptr = malloc(size);
+  memset(ptr, 0, size);
```
wjwwood: `ptr` could be null if `malloc` failed.

dirk-thomas:
Is this necessary for the demo at ROSCon? If not I would suggest postponing it until Monday.

jacquelinekay:
CI:
http://ci.ros2.org/job/ros2_batch_ci_linux/402
http://ci.ros2.org/job/ros2_batch_ci_osx/299
http://ci.ros2.org/job/ros2_batch_ci_windows/447

wjwwood:
lgtm

jacquelinekay:
Something is wrong with the Windows build, but I can't tell if it's related to these changes.

jacquelinekay:
I think the original Windows build failure was a fluke, I reran it and it looks fine.

http://ci.ros2.org/job/ros2_batch_ci_windows/449/

I'll squash and merge.

PR #44, opened 2015-10-22
replace strcpy with snprintf

dirk-thomas:
Following `cpplint`.

http://ci.ros2.org/job/ros2_batch_ci_linux/455/
http://ci.ros2.org/job/ros2_batch_ci_osx/367/


esteve:
+1

PR #45, opened 2015-10-29
implement workaround for unresolved symbols problem on Windows

wjwwood:
Connects to ros2/rclcpp#48

Implements a workaround for the fact that the Windows linker cannot handle unresolved symbols when linking a shared library, see: https://github.com/ros2/rclcpp/issues/48#issuecomment-150714947

### DIFF
--- a/rmw/package.xml
+++ b/rmw/package.xml
@@ -12,6 +12,8 @@
   <!-- This is required for the definition of the rosidl typesupport types -->
   <build_export_depend>rosidl_generator_c</build_export_depend>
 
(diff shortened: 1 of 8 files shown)


dirk-thomas:
Can I assume that the copied CMake code has no changes and therefore doesn't need to be reviewed?

wjwwood:
I didn't change the cmake code in the three cmake files that moved.

dirk-thomas:
lgtm

esteve:
+1

PR #46, opened 2015-11-04
move implementation of allocators into c file

dirk-thomas:
This currently prevent rmw implementations to consist of more then a single source file.

The code has only been moved but not modified.

http://ci.ros2.org/job/ros2_batch_ci_osx/415/


esteve:
+1

PR #47, opened 2015-11-13
only persist the rmw impl if it was explicitly set

dirk-thomas:
If the user didn't explicitly set the rmw implementation it should be redetermined on every CMake run.

Otherwise when it was selected dynamically (the first available one) a subsequent build fails when the available rmw implementations change.


tfoote:
+1

PR #48, opened 2015-12-01
changes to support rcl refactor

wjwwood:
While testing, I realized that we need to change some things in the `rmw` API to support the rcl library changes. The two biggest changes are to make the API actually C and not C++, and to expose things like the node name and topic names through the rmw  API. Currently I plan to do the latter by putting these fields directly in the rmw struct rather than adding getter and setter functions to the rmw API and making it larger.

This pr is still a work in progress as I discover things in testing. I'll also be opening pr's against all of the current rmw implementations to bring them up to speed, but I am testing locally with opensplice for the time being.

Later, but probably not on the first pass, I think we should make the rmw interface more like the rcl interface in a few ways, for example exposing an allocator option in some places, using a wait set struct, and thorough documentation which considers thread-safety, memory allocation, complex usage, etc...

Connects to ros2/rcl#5

### DIFF
--- a/rmw/include/rmw/error_handling.h
+++ b/rmw/include/rmw/error_handling.h
@@ -20,6 +20,7 @@ extern "C"
 {
 #endif
 
+#include <stdbool.h>
 #include <stddef.h>
 
 #include "rmw/macros.h"
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -190,7 +190,7 @@ rmw_wait(
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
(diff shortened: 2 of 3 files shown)


wjwwood:
Note to myself, need to check all uses of functions which had their default arguments dropped.

jacquelinekay:
Should services and clients also have the service name in the struct (similar to adding the topic name to subscribers and publishers)?

wjwwood:
Yes, but my `rcl` work so far doesn't support services so I've been sort of ignoring changes related to that.

jacquelinekay:
Ok, sounds fine with me.

jacquelinekay:
+1

dirk-thomas:
+1

PR #49, opened 2015-12-02
Add rmw_waitset_t and create/destroy functions

jacquelinekay:
Improvements discussed in ros2/rmw_opensplice#96 and offline.

Seeking feedback on this approach before implementing it for Connext.

This API change will allow for several improvements to be made.

- Re-use pre-allocated WaitSet and ConditionSeq in rmw_wait, reducing allocations made at the rmw layer.
- Each Executor can keep a handle on its own WaitSet, allowing for multiple Executors to execute concurrently in their own threads. If a global static WaitSet were to be used in this case, a race condition would occur and the accesses would interfere with each other.
- We could add locking to the WaitSet destructor to prevent the race condition with `trigger_guard_condition`.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -182,6 +182,29 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_trigger_guard_condition(const rmw_guard_condition_t * guard_condition);
 
+/// Create a waitset to store conditions that the middleware will block on.
+/**
+ * If max_conditions is 0, the waitset can store an unbounded number of conditions to wait on.
+ * If max_conditions is greater than 0, the number of conditions that can be attached to the
+ * waitset is bounded at max_conditions.
+ * \param[in] fixed_guard_conditions Guard conditions that are attached to the waitset immediately
+ * after initialization and are not removed from the waitset until rmw_destroy_waitset is called.
+ * For example, the guard condition for the CTRL-C signal handler is a "fixed condition".
+ * Must stay valid until rmw_destroy_waitset is called. The caller must keep ownership of the
+ * allocated guard conditions and deallocate them after rmw_destroy_waitset is called.
+ * \param[in] max_conditions The maximum number of conditions that can be attached to the waitset.
+ * \return A pointer to the created waitset, nullptr if an error occurred.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_waitset_t *
+rmw_create_waitset(rmw_guard_conditions_t * fixed_guard_conditions, size_t max_conditions);
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_destroy_waitset(rmw_waitset_t * waitset);
+
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_ret_t
@@ -190,6 +213,7 @@ rmw_wait(
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
+  rmw_waitset_t * waitset,
   const rmw_time_t * wait_timeout);
 
 RMW_PUBLIC
--- a/rmw/include/rmw/allocators.h
+++ b/rmw/include/rmw/allocators.h
@@ -79,6 +79,14 @@ RMW_PUBLIC
 void
 rmw_service_free(rmw_service_t * service);
 
+RMW_PUBLIC
+rmw_waitset_t *
+rmw_waitset_allocate();
+
+RMW_PUBLIC
+void
+rmw_waitset_free(rmw_waitset_t * waitset);
+
 #if __cplusplus
 }
 #endif
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -99,6 +99,13 @@ typedef struct RMW_PUBLIC_TYPE rmw_guard_conditions_t
   void ** guard_conditions;
 } rmw_guard_conditions_t;
 
+typedef struct RMW_PUBLIC_TYPE rmw_waitset_t
+{
+  const char * implementation_identifier;
+  rmw_guard_conditions_t * fixed_guard_conditions;
+  void * data;
+} rmw_waitset_t;
+
 typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
 {
   int8_t writer_guid[16];
--- a/rmw/src/allocators.c
+++ b/rmw/src/allocators.c
@@ -121,3 +121,17 @@ rmw_service_free(rmw_service_t * service)
   // Should have matching overide with rmw_service_allocate
   rmw_free(service);
 }
+
+rmw_waitset_t *
+rmw_waitset_allocate()
+{
+  // Could be overridden with custom (maybe static) client struct allocator
+  return (rmw_waitset_t *)rmw_allocate(sizeof(rmw_waitset_t));
+}
+
+void
+rmw_waitset_free(rmw_waitset_t * waitset)
+{
+  // Should have matching overide with rmw_waitset_allocate
+  rmw_free(waitset);
+}


rmw/include/rmw/rmw.h
```diff
@@ -184,12 +184,28 @@ rmw_trigger_guard_condition(const rmw_guard_condition_t * guard_condition);
...
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_destroy_waitset(rmw_waitset_t * waitset);
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_fini_waitset(rmw_waitset_t * waitset);
```
wjwwood: What does fini do that destroy doesn't? Maybe put another way, what is the expected usage of fini without a matching init function?

rmw/include/rmw/rmw.h
```diff
@@ -184,12 +184,28 @@ rmw_trigger_guard_condition(const rmw_guard_condition_t * guard_condition);
 
 RMW_PUBLIC
 RMW_WARN_UNUSED
+rmw_waitset_t *
+rmw_create_waitset(size_t max_conditions);
```
dirk-thomas: What is the approach to continue using a waitset in the current unbounded way?

jacquelinekay: If max_conditions is 0, then this function will default-construct the `ConditionSeq` objects used by the WaitSet (effectively the curent behavior). If max_conditions is greater than 0, this function will pass max_conditions to the ConditionSeq constructor, which specifies the upper bound on the Sequence objects.

dirk-thomas: Can you please add this explanation in a docblock.

rmw/include/rmw/rmw.h
jacquelinekay: I guess this doesn't need to be its own function right now, since all it does is detach all currently attached conditions, which we expect to happen in the WaitSet destructor.

wjwwood: After reading what it's implementation does in your rmw_opensplice pr, I'd say it is the counter part to `rmw_wait`, such that it would be used like this:

- `rmw_create_waitset`
- while running:
 - `rmw_wait`
 - `rmw_fini_waitset`
- `rmw_destroy_waitset`

But then the question becomes, why not detach at the end of every call to `rmw_wait`?

If we leave the separate function in, maybe we should consider renaming it to something like `rmw_waitset_reset` or `rmw_waitset_clear` or `rmw_waitset_detachall`, etc...

dirk-thomas: I think avoiding the extra function and detaching all conditions within `rmw_wait` makes sense. `rmw_wait` could also ensure the precondition at the beginning that no conditions are attached.

jacquelinekay: I've removed rmw_fini_waitset and moved the code into the scope exit destructor.

It requires more effort, but I still think I should refactor this function to use a goto fail to propagate any errors that occur while detaching conditions.

rmw/include/rmw/rmw.h
```diff
@@ -182,6 +182,24 @@ RMW_WARN_UNUSED
...
+ * If max_conditions is 0, the waitset can store an unbounded number of conditions to wait on.
+ * If max_conditions is greater than 0, the number of conditions that can be attached to the
+ * waitset is bounded at max_conditions.
+ * \param[in] max_conditions The maximum number of conditions that can be attached to the waitset.
+ * \return A pointer to the created waitset, nullptr if an error occurred.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_waitset_t *
+rmw_create_waitset(rmw_guard_conditions_t * fixed_guard_conditions, size_t max_conditions);
```
dirk-thomas: `fixed_guard_conditions` is not documented and I am not sure how it is supposed to be used.

jacquelinekay:
Most recent Jenkins jobs for this set of branches:

http://ci.ros2.org/job/ci_linux/672/
http://ci.ros2.org/job/ci_osx/551/
http://ci.ros2.org/job/ci_windows_connext_dynamic/29/
http://ci.ros2.org/job/ci_windows_connext_static/28/
http://ci.ros2.org/job/ci_windows_opensplice/744/testReport/

gerkey:
@jacquelinekay Do you want to get this set of changes into alpha3?

jacquelinekay:
That would be really fantastic.

It reduces allocations in both rmw implementations. Actually, it makes
rmw_connext allocation-free during the execution path.

There is a pending set of changes that I haven't pull requested yet because
it depends heavily on this one which would finish the job for Opensplice.
But given how much we need to review today it might not get in.

I had a clean set of Jenkins job for this set of branches but I neglected
to post it in the comments. I will rebase these branches and restart CI.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/785/
http://ci.ros2.org/job/ci_osx/652/
http://ci.ros2.org/job/ci_windows/833

jacquelinekay:
I had to rebuild the CI jobs because the examples in realtime_support that pass arguments to the Executor needed to be refactored.

esteve:
+1

dirk-thomas:
LGTM

jacquelinekay:
can you guys clarify if those +1s are for just the rmw PR or the whole thing?

dirk-thomas:
I only looked at this one - reading the other ones now.

jacquelinekay:
I realized as I was rebasing my branches today to get this the build with the rcl refactor that @wjwwood may have miscommunicated about what I call a "waitset" in rmw and what he implemented in rcl as a `wait_set`.

I'm currently thinking about how to incorporate the features that I implemented in my set of branches with rcl so that it's full featured, but it would require a few API changes to rcl and I'm not sure if we have tiem for that right now.

So, I'll probably throw in a dummy_waitset of type rmw_waitset_t into rcl so that the API matches my change to rmw_wait and incorporate the full functionality next week.

jacquelinekay:
I also realized that I have to make a pull request to rmw_fastrtps_cpp now that it's merged, which may take some work because they use a std::condition_variable instead of something like a dds WaitSet.

gerkey:
OK, then let's leave this PR out of alpha3.

jacquelinekay:
alright. This was poor timing/planning by the team. I should have made more of an effort to get this PR reviewed and merged well before the fastrtps repository was merged. I considered it ready for review nearly a week before it received review. Or, rmw_fastrtps should have had an earlier deadline so that we could have had more time to adapt it to the changes we wanted to make the rmw API.

wjwwood:
I agree I think we should have serialized the merge of changes to rmw and inclusion of fastrtps differently. I should have noticed other pull requests, like this one, which had been in review before mine and pointed out the oncoming conflict.

gerkey:
Agreed, we didn't consider the full consequences of adding FastRTPS.

jacquelinekay:
The rmw_fastrtps changes are in this pull request (I don't think we have FastRTPS added to Waffle):

https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/9

The new (and hopefully final?) set of CI jobs are here:

http://ci.ros2.org/job/ci_linux/821/
http://ci.ros2.org/job/ci_osx/688/
http://ci.ros2.org/job/ci_windows/874

jacquelinekay:
Needed some synchronization fixes:

http://ci.ros2.org/job/ci_linux/824/
http://ci.ros2.org/job/ci_osx/689/
http://ci.ros2.org/job/ci_windows/875/

jacquelinekay:
not sure why Jenkins fails; the same tests pass on my machine (except for tests that are also failing on the nightly build, such as test_services_cpp__rmw_fastrtps_cpp)

jacquelinekay:
Current CI:

http://ci.ros2.org/job/ci_linux/836/
http://ci.ros2.org/job/ci_osx/699/
http://ci.ros2.org/job/ci_windows/890

Test failures are all known failures on the nightly builds.

rmw/include/rmw/rmw.h
```diff
@@ -182,6 +182,27 @@ RMW_WARN_UNUSED
...
 rmw_trigger_guard_condition(const rmw_guard_condition_t * guard_condition);
 
+/// Create a waitset to store conditions that the middleware will block on.
+/**
+ * If max_conditions is 0, the waitset can store an unbounded number of conditions to wait on.
+ * If max_conditions is greater than 0, the number of conditions that can be attached to the
+ * waitset is bounded at max_conditions.
+ * \param[in] fixed_guard_conditions Guard conditions that are attached to the waitset immediately
+ * after initialization and are not removed from the waitset until rmw_destroy_waitset is called.
+ * For example, the guard condition for the CTRL-C signal handler is a "fixed condition".
```
dirk-thomas: Is it correct that the `fixed_guard_conditions` are only detached but not destroyed when the waitset is destructed? Should that be mentioned explicitly in the docstring?

jacquelinekay: `fixed_guard_conditions` is expected to be allocated and fully initialized when provided to this function. I think it's better for the caller of this function to be responsible for destroying the object, instead of the function itself, since this function does not allocate the memory.

dirk-thomas: I fully agree to that. I only think something like the following would clarify the documentation:

> fixed_guard_conditions must stay valid until rmw_destroy_waitset hsa been called and the caller keep the ownership of the allocated guard conditions.

jacquelinekay:
I believe I've addressed all of @dirk-thomas's comments, so I've relaunched CI.

http://ci.ros2.org/job/ci_linux/840/
http://ci.ros2.org/job/ci_osx/703/
http://ci.ros2.org/job/ci_windows/901/

rmw/include/rmw/rmw.h
jacquelinekay: this was added in 65768e4

dirk-thomas:
Not sure if those are related to these changes:
* http://ci.ros2.org/job/ci_linux/841/
* http://ci.ros2.org/job/ci_osx/704/
* (http://ci.ros2.org/job/ci_windows/903/)

PR #53, opened 2016-02-04
Fix "function declaration isn't a prototype" warnings

esteve:
Fix "function declaration isn't a prototype" warnings

### DIFF
--- a/rmw/include/rmw/allocators.h
+++ b/rmw/include/rmw/allocators.h
@@ -33,7 +33,7 @@ rmw_free(void * pointer);
 
 RMW_PUBLIC
 rmw_node_t *
-rmw_node_allocate();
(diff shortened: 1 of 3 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/895/
http://ci.ros2.org/job/ci_osx/750/
http://ci.ros2.org/job/ci_windows/976/

jacquelinekay:
Looks fine to me. Where was the warning coming from?

esteve:
From `rclpy`. GCC reports it as a warning, but it actually causes a linking conflict, so when I called `rmw_get_implementation_identifier()`, I got a bogus string instead of the real thing. Properly declaring that function as a prototype fixed the problem.

Issue #38, opened 2015-09-08
RMW_CHECK_TYPE_IDENTIFIERS_MATCH macro for type support


jacquelinekay:
Currently, the macro `RMW_CHECK_TYPE_IDENTIFIERS_MATCH` has an error message of the format:

`" implementation '%s'(%p) does not match rmw implementation '%s'(%p)"`

However, in the rmw implementations this macro is used for checking the types of typesupport identifiers, not only rmw implementation identifiers, leading to confusing and incorrect error messages such as:

`could not create publisher: type support handle implementation 'connext_static'(0x2b54f95e9671) does not match rmw implementation 'introspection'(0x2b54f97eb691)`

This macro should be reworded to be more general, or a different macro should be made for matching typesupport identifiers.

tfoote:
@jacquelinekay to update status on this issue/pr.

dirk-thomas:
I don't see any cases anymore where this macro is being invoked with type support arguments. Therefore I will close this.

PR #55, opened 2016-02-18
#38 adds condition to check for type support in RMW_CHECK_TYPE_IDENTI…

madratman:
…FIERS_MATCH

Connects to #38

### DIFF
--- a/rmw/include/rmw/impl/cpp/macros.hpp
+++ b/rmw/include/rmw/impl/cpp/macros.hpp
@@ -66,13 +66,26 @@
 }
 
 #if RMW_AVOID_MEMORY_ALLOCATION
+#define PRINT_TYPE_SUPPORT_MISMATCH() #ElementName \
+  " implementation '%s'(%p) does not match type support implementation '%s'(%p)", \
+  ElementTypeID, ElementTypeID, ExpectedTypeID, ExpectedTypeID
+
+#define PRINT_RMW_IMPL_MISMATCH() #ElementName \
+  " implementation '%s'(%p) does not match rmw implementation '%s'(%p)", \
+  ElementTypeID, ElementTypeID, ExpectedTypeID, ExpectedTypeID
+
 #define RMW_CHECK_TYPE_IDENTIFIERS_MATCH(ElementName, ElementTypeID, ExpectedTypeID, OnFailure) { \
   if (ElementTypeID != ExpectedTypeID) { \
     char __msg[1024]; \
-    snprintf( \
-      __msg, 1024, \
-      #ElementName " implementation '%s'(%p) does not match rmw implementation '%s'(%p)", \
-      ElementTypeID, ElementTypeID, ExpectedTypeID, ExpectedTypeID); \
+    if(!strcmp(#ElementName, "type support")) { \
+      snprintf( \
+        __msg, 1024, \
+        PRINT_TYPE_SUPPORT_MISMATCH()); \
+    } else { \
+      snprintf( \
+        __msg, 1024, \
+        PRINT_RMW_IMPL_MISMATCH()); \
+    } \
     RMW_SET_ERROR_MSG(__msg); \
     OnFailure; \
   } \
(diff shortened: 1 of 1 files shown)


jacquelinekay:
Can you run the style checker on this code by running this command in the root of your ament workspace:

ament.py test --skip-build --skip-install --only rmw --ctest-args -L linter

rmw/include/rmw/impl/cpp/macros.hpp
```diff
@@ -69,10 +69,17 @@
 #define RMW_CHECK_TYPE_IDENTIFIERS_MATCH(ElementName, ElementTypeID, ExpectedTypeID, OnFailure) { \
   if (ElementTypeID != ExpectedTypeID) { \
     char __msg[1024]; \
-    snprintf( \
-      __msg, 1024, \
-      #ElementName " implementation '%s'(%p) does not match rmw implementation '%s'(%p)", \
-      ElementTypeID, ElementTypeID, ExpectedTypeID, ExpectedTypeID); \
+    if(#ElementName == "type support") { \
```
dirk-thomas: This condition is never true since plain string can't be compared with `==` in C++.

A string comparison on the other hand is pretty expensive. I think having two separate macros for that makes more sense. (Both could still share the majority of the code using a third "internal" macro.)

madratman:
@jacquelinekay fixed the style. Tests pass now. (but not sure if [L75-76](https://github.com/madratman/rmw/blob/a0d1bfc6238750ed22df1b2d44222b84797e3bfc/rmw/include/rmw/impl/cpp/macros.hpp#L75-76), L95-96, L109-110 are readable enough.)

@dirk-thomas This [gist](https://gist.github.com/madratman/122193f12f7863633c59) prints 1. Or I missed something? `#ElementName` is just an `str()` for the node/publisher/whatever handle is passed right?
I agree about code duplication. Will fix that in next commit. (and rebase later when it's final) 
Also, exactly how much of a time difference will the string comparison make? For an example failure case, how many times `RMW_CHECK_TYPE_IDENTIFIERS_MATCH` would be called for instance?

dirk-thomas:
Since the comparison only happens in the case of an error it is probably ok to perform a string comparison.

dirk-thomas:
Regarding the gist if you change the type of `name` to `char*` it won't work and the macro has no way to ensure it gets passed a `std::string`.

madratman:
[Gist_2](https://gist.github.com/madratman/ad83822c6fc40078b6eb) (with `char*`) still works (albeit there are compiler warnings). 
```
 In function 'int main()':
9:18: warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]
10:23: warning: comparison with string literal results in unspecified behaviour [-Waddress]
18:21: warning: comparison with string literal results in unspecified behaviour [-Waddress]
22:22: warning: comparison with string literal results in unspecified behaviour [-Waddress]
``` 
(I am being lazy and using http://cpp.sh/ for this)

madratman:
Anyway, changed to strcmp, and removed duplication

rmw/include/rmw/impl/cpp/macros.hpp
```diff
@@ -66,13 +66,26 @@
...
+  ElementTypeID, ElementTypeID, ExpectedTypeID, ExpectedTypeID
+
 #define RMW_CHECK_TYPE_IDENTIFIERS_MATCH(ElementName, ElementTypeID, ExpectedTypeID, OnFailure) { \
   if (ElementTypeID != ExpectedTypeID) { \
     char __msg[1024]; \
-    snprintf( \
-      __msg, 1024, \
-      #ElementName " implementation '%s'(%p) does not match rmw implementation '%s'(%p)", \
-      ElementTypeID, ElementTypeID, ExpectedTypeID, ExpectedTypeID); \
+    if(strcmp(#ElementName, "type support")) { \
```
dirk-thomas: `strcmp` returns `0` when the strings match. So the logic is inverted here since it is interpreted as bool.

madratman: oh man! >.<

jacquelinekay:
sorry about the delay on review for this. In the current state the code doesn't compile. I'm going to push this back into backlog until there is confirmation from @madratman that the PR works with the current state of ROS 2.

dirk-thomas:
@madratman What is the status of this patch? Are you planning to iterate on it or should it be closed?

dirk-thomas:
I will close this ticket since I don't see why it is needed anymore, see https://github.com/ros2/rmw/issues/38#issuecomment-293041539 Please feel free to comment and it can be reopened.

PR #56, opened 2016-02-27
Add service name to client and service

jacquelinekay:


tfoote:
Before we do this do we want to revisit the named resource discussion?

jacquelinekay:
Sure. Can you link to any existing issues about it? I'm not familiar with that discussion off the top of my head.

I figured this was justified because topic_name is a field for both subscription and publisher right now:
https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h#L46-L58

thought it's not fully implemented on master yet (I have a series of PRs to finish the implementation in review right now: ros2/rcl#29)

wjwwood:
I'm not exactly clear on what @tfoote is talking about, it could be ROS 1 Names (http://wiki.ros.org/Names) or "named" services (ones that the user gives a name for and doesn't have to hold on to the handle to keep it alive). I'd guess the former.

Either way I agree that this should be done to bring it in line with the other `rmw` structures. I wouldn't block it on the other design considerations (we'll have to change lots of things anyways when we decide that stuff).

jacquelinekay:
I've accepted the possibility that at least 80% of the code I'm writing will be rewritten at some point once we've made more design decisions.

tfoote:
@wjwwood more generally I was referring to supporting fully qualified resource addressing so as to remove ambiguity between topics, services, parameters and nodes. I don't necessarily want to hold this up but thought that it might be something we should bump up in priority.

jacquelinekay:
even though this branch is called c_services I've decided to start an identical branch called service_name and split the connext and opensplice changes into their own branches, to make this a separate change from C services (since it's actually orthogonal). I could close this PR and reopen one for the new branch if needed.

jacquelinekay:
CI for the service_name changes:
http://ci.ros2.org/job/ci_linux/1045/
http://ci.ros2.org/job/ci_osx/862/
http://ci.ros2.org/job/ci_windows/1108/

jacquelinekay:
a test for this functionality will come in ros2/rcl#30

jacquelinekay:
need to address warnings on Windows

jacquelinekay:
I've changed strncpy back to memcpy. This pending CI job will hopefully be clean:

http://ci.ros2.org/job/ci_windows/1113

wjwwood:
This and related pr's lgtm, if you resolved the warnings on Windows (it's hard to tell from the CI, but I think you did).

jacquelinekay:
I remember looking at that set of warnings and thinking that they were all present on master at the time.

PR #61, opened 2016-04-01
Document rmw_wait

jacquelinekay:
In accordance with a recent discussion with @dirk-thomas and @wjwwood, this adds a comment describing the behavior of rmw_wait and how it treats its input arrays of void * pointers.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -187,11 +187,6 @@ rmw_trigger_guard_condition(const rmw_guard_condition_t * guard_condition);
  * If max_conditions is 0, the waitset can store an unbounded number of conditions to wait on.
  * If max_conditions is greater than 0, the number of conditions that can be attached to the
  * waitset is bounded at max_conditions.
- * \param[in] fixed_guard_conditions Guard conditions that are attached to the waitset immediately
- * after initialization and are not removed from the waitset until rmw_destroy_waitset is called.
- * For example, the guard condition for the CTRL-C signal handler is a "fixed condition".
- * Must stay valid until rmw_destroy_waitset is called. The caller must keep ownership of the
- * allocated guard conditions and deallocate them after rmw_destroy_waitset is called.
  * \param[in] max_conditions The maximum number of conditions that can be attached to the waitset.
  * \return A pointer to the created waitset, nullptr if an error occurred.
  */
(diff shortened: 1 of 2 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -205,6 +205,25 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_destroy_waitset(rmw_waitset_t * waitset);
 
+/**
+ * Add conditions to the wait set and wait until a response comes in, or until the timeout is
+ * reached.
+ * The entries of the input arrays are void * pointers that are cast to middleware-specific
+ * conditions and added to the waitset.
```
dirk-thomas: The information of this sentence is encoded in the types which are passed in, e.g. `rmw_subscriptions_t`. How they work should not be part of this docblock imo.

rmw/include/rmw/rmw.h
```diff
@@ -205,6 +205,25 @@ RMW_WARN_UNUSED
...
+ * Null pointers are considered invalid and an error is returned if they are encountered.
```
dirk-thomas: Null pointers **within the input arrays** are considered... ?

rmw/include/rmw/rmw.h
```diff
@@ -205,6 +205,25 @@ RMW_WARN_UNUSED
...
+ * The count variables in the arrays represents the number of valid pointers in the array.
```
dirk-thomas: I would move this sentence one line up (before the null pointers).

rmw/include/rmw/rmw.h
```diff
@@ -205,6 +205,25 @@ RMW_WARN_UNUSED
...
+ *
+ * After the wait wakes up, the entries in each array that correspond to conditions that were
+ * not triggered are set to null.
+ * 
+ * \param subscriptions Array of subscriptions to wait on
+ * \param guard_conditions Array of guard conditions to wait on
+ * \param services Array of services to wait on
+ * \param clients Array of clients to wait on
+ * \return RCL_RET_OK if success, RCL_RET_ERROR if error, RCL_RET_TIMEOUT if wait timed out.
+ */
```
dirk-thomas: The parameters `waitset` and `waitset_timeout `are not mentioned with `\param`. Especially for the timeout valid values and their semantic should be mentioned.

rmw/include/rmw/rmw.h
jacquelinekay: True, although if I move this information to the definition to `rmw_subscriptions_t`, then I'll be referencing the information in this function (e.g. the `void *` pointers are cast to middleware-specific conditions in `rmw_wait`).

dirk-thomas: Sounds good. But you might not need an explicit reference since all parameter types will be clickable links anyway.

wjwwood:
+1, with @dirk-thomas's comments addressed.

jacquelinekay:
Comments addressed in 88fb6c1efbfd021dbb769368899a243c3a512a2a

dirk-thomas:
I wrapped each sentence. LGTM.

wjwwood:
@jacquelinekay you can try the new "squash + merge" button!

wjwwood:
nvm :cry:

jacquelinekay:
Oops too late

PR #63, opened 2016-04-06
Register language of typesupport for rmw implementation

jacquelinekay:
I believe that when typesupport implementations are registered in the ament index, there should be a way to distinguish between C and C++ typesupport packages. This is one way to implement it, I'm open to feedback about how to change the design. A companion PR to rosidl_generator_py and various other packages will accompany this to show usage.

Right now a typesupport implementation is registered twice: as a "rmw_typesupport" and as a "rmw_typesupport_c/cpp". That way all typesupports can be selected using the "rmw_typesupport" identifier, but in contexts where either C or C++ typesupport is needed, there is a way to distinguish between them.


jacquelinekay:
CI:

http://ci.ros2.org/job/ci_linux/1171/
http://ci.ros2.org/job/ci_osx/947/
http://ci.ros2.org/job/ci_windows/1227/

rmw/cmake/get_rmw_typesupport.cmake
```diff
@@ -26,5 +26,39 @@ macro(get_rmw_typesupport var rmw_implementation)
     message(FATAL_ERROR "get_rmw_typesupport() called with unused arguments: ${ARGN}")
   endif()
 
-  ament_index_get_resource(${var} "rmw_implementation" "${rmw_implementation}")
+  ament_index_get_resource(${var} "rmw_typesupport" "${rmw_implementation}")
+endmacro()
+
+#
+# Get the C type support package name for a specific ROS middleware implementation.
```
dirk-thomas: Couldn't this return more then one?

Same below.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,41 @@
...
-  if(NOT "${ARGN} " STREQUAL " ")
-    message(FATAL_ERROR "register_rmw_implementation() called with "
-      "unused arguments: ${ARGN}")
+macro(register_rmw_implementation)
+  if("${ARGN} " STREQUAL " ")
+    message(FATAL_ERROR "register_rmw_implementation() called with no arguments!")
+  endif()
+
+  cmake_parse_arguments(_ARG
+    "TRACE"
```
dirk-thomas: This seems to be not used anywhere?

jacquelinekay:
I realized after starting the build that PR was subtly broken. I renamed the key for the ament index entry to `rmw_typesupport` (from `rmw_implementation`, which I thought was misleading). Fixed in 929018e.

jacquelinekay:
I launched a new build where everything looks good except for failing tests in rclpy. I cannot reproduce this locally. Hmmm.

http://ci.ros2.org/job/ci_linux/1173/

wjwwood:
@jacquelinekay it's failing because you don't have any rmw implementations enabled which are supported by `rclpy`. The error message for this case is improved in https://github.com/ros2/rclpy/pull/12/files#diff-aa7e84a9e617f0578e51ed0daace94aaR93.

jacquelinekay:
But use_connext and use_opensplice are both set to true in the job with the failing test (and disable_connext_static is false), so you would expect those rmw implementations to be found and for the test to pass. This leads me to believe that my changes have broken the logic for finding an rmw implementation in Python.

wjwwood:
@jacquelinekay maybe. It's also possible there is a problem with how my code changes and yours have merged? The supported Python rmw extensions are now registered separately.

jacquelinekay:
That's what I was thinking. I thought I rebuilt my workspace locally with an up-to-date version of rclpy; I'm going to try again and take a closer look at your changes in rclpy.

wjwwood:
Ok, if there is an issue, then this commit will be the culprit: https://github.com/ros2/rclpy/commit/0bf96409db80ff12518a2ca51e3e3294bb775811

jacquelinekay:
The rclpy tests passed on Jenkins in this new build:

http://ci.ros2.org/job/ci_linux/1173/

wjwwood:
So, what made the difference in your opinion?

jacquelinekay:
It looks like the Jenkins job with the failed rclpy test was before your rclpy test was merged, and the successful one was launched afterwards.

All I know is that if I do a `git reset --hard HEAD~1` locally in rclpy to revert your most recent PR and rebuild my workspace, the rclpy tests fail (for the same reason as seen on Jenkins, no valid rmw implementation for rclpy found.) But both on Jenkins and locally, if I go back to the current state of rclpy on master, then the test passes.

wjwwood:
Well there was a matching pr merged, so to "undo" what I merged, you'd need to also undo this one: https://github.com/ament/ament_cmake/pull/65

jacquelinekay:
@dirk-thomas I tried to address the multiple typesupports issue but I didn't understand how to use `ament_index_get_resources` and `ament_index_register_resource` for this purpose.

For example, I want to do something like this:

```
ament_index['rmw_typesupport'] = {'rmw_connext_cpp' : ['rosidl_typesupport_connext_c', 'rosidl_typesupport_connext_cpp'] }

ament_index['rmw_typesupport_c'] = {'rmw_connext_cpp' : ['rosidl_typesupport_connext_c'] }
ament_index['rmw_typesupport_cpp'] = {'rmw_connext_cpp' : ['rosidl_typesupport_connext_cpp'] }
```

but the first resource map doesn't work because I cannot store a list of string at `rmw_connext_cpp`.

Should I expect to be able to do this?

I have a workaround in 12b9777.

dirk-thomas:
Within one resource file you can story any kind of content. If you choose to store a CMake list in it that will be exactly what you will get back out.

jacquelinekay:
I see, so if I expect an entry to store multiple strings, I should initialize it and append to it as a list.

jacquelinekay:
I fixed the index entries in this pull request so that they all can store a list of strings in b7d7897

dirk-thomas:
Instead of creating new macros for each language why not have a single function with a *language* parameter?

jacquelinekay:
Probably a better solution. Do you suggest a positional argument or a flag (e.g. `get_rmw_typesupport(var rmw_connext_cpp LANGUAGE C)`)? I prefer the latter.

dirk-thomas:
If it is optional a keyword argument would be ok but I think it is required and therefore I think a positional argument makes more sense.

jacquelinekay:
I was thinking for `get_rmw_typesupport` that the "language" parameter could be optional, and if it's left out then the typesupports for all languages is returned.

dirk-thomas:
That sounds good to me.

jacquelinekay:
Another round of CI:
http://ci.ros2.org/job/ci_linux/1181/
http://ci.ros2.org/job/ci_osx/958/
http://ci.ros2.org/job/ci_windows/1236/

rmw/cmake/get_rmw_typesupport.cmake
```diff
@@ -13,18 +13,30 @@
...
-  if(NOT "${ARGN} " STREQUAL " ")
-    message(FATAL_ERROR "get_rmw_typesupport() called with unused arguments: ${ARGN}")
+  cmake_parse_arguments(_ARG
+    ""
+    "LANGUAGE"
+    ""
+    ${ARGN}
+  )
+  if("${_ARG_LANGUAGE}" STREQUAL "C")
+    ament_index_get_resource(${var} "rmw_typesupport_c" ${rmw_implementation})
```
dirk-thomas: Please avoid hard coding specific languages in the CMake code. This should be able to handle arbitrary languages in the future.

jacquelinekay: ok. I'll rewrite it in a more generic way.

jacquelinekay:
In 9b823e5 I changed it so that `register_rmw_implementation` parses its arguments such that the arguments must come in pairs of (language, typesupport package names). There isn't very much error checking however.

One more CI job for good measure:
http://ci.ros2.org/job/ci_linux/1184

rmw/cmake/get_rmw_typesupport.cmake
```diff
@@ -13,18 +13,30 @@
...
+    ""
+    ${ARGN}
+  )
+  if("${_ARG_LANGUAGE}" STREQUAL "")
+    ament_index_get_resource(${var} "rmw_typesupport" ${rmw_implementation})
+  else()
+    # This output will include type supports the specified language.
+    # The language string should not be case sensitive.
+    string(TOLOWER ${_ARG_LANGUAGE} language_key)
+    ament_index_get_resource(${var} "rmw_typesupport_${language_key}" ${rmw_implementation})
```
dirk-thomas: Please avoid code duplications like line 35 and 40 (same in the next CMake file lines 29 and 31). See #64 for changing this function.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -13,19 +13,67 @@
...
+    list(GET argn ${idx} LABEL)
+    # The key should not be case sensitive
+    string(TOLOWER ${LABEL} language_label)
+    math(EXPR next_idx "${idx}+1")
+    list(GET argn ${next_idx} CONTENT)
+    ament_register_resource_append_if_found(
+      "rmw_typesupport_${language_label}"
+      "${CONTENT}")
+    ament_register_resource_append_if_found("rmw_typesupport" "${CONTENT}")
+  endforeach()
```
dirk-thomas: Is this looping over the arguments, write each one into the index and in every cycle check the index again, reads the value and appends the new one? Then this should be restructured to concatenate the content within CMake and only write it once to the index at the very end.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -13,19 +13,67 @@
 # limitations under the License.
 
 #
+# Helper function to append items to an ament resource list, or create a new
+# entry for the resource if no entry exists.
+#
+# :param _ARG_C: Name of the C typesupport package for the rmw implementation.
```
dirk-thomas: The documented param `_ARG_C` seem to not exist.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -13,19 +13,67 @@
...
+#
+function(ament_register_resource_append_if_found resource_type)
+  if("${ARGN} " STREQUAL " ")
+    message(FATAL_ERROR "register_rmw_implementation() called with no arguments!")
```
dirk-thomas: The error message doesn't match the function name.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -13,19 +13,67 @@
...
+  endif()
+  ament_index_has_resource(HAS_RESOURCE "${resource_type}" "${PROJECT_NAME}")
+  if(${HAS_RESOURCE})
+    # Get the entry and append it to a list
+    list_append_unique(rmw_typesupport ${CONTENT}})
+    ament_index_register_resource("${resource_type}" CONTENT "${ARGN}")
+  else()
+    ament_index_register_resource("${resource_type}" CONTENT "${ARGN}")
+  endif()
+endfunction()
```
dirk-thomas: Maybe this whole function can go away with https://github.com/ros2/rmw/pull/63/files#r59761455

jacquelinekay: I think I agree. I will take a look.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -13,19 +13,67 @@
...
+# of this type support package.
+# For example, a valid usage of this function would be
+# register_rmw_implementation(c rosidl_typesupport_introspection_c
+#                             cpp rosidl_typesupport_introspection_cpp)
 #
-# :param typesupport_package_name: the package name providing type support for
-#   for the registered RMW implementation
-# :type typesupport_package_name: string
+# :param _ARG_LANGUAGE: Name of the language represented by this typesupport package
+# :param _ARG_CONTENT: Name of the typesupport package.
```
dirk-thomas: These parameters seem to not exist.

Instead I would propose to pass in a list of pairs where each pair is `<TYPE_SUPPORT_PKG_NAME>:<LANGUAGE>`. That is much easier to construct on the caller side as well as much easier to document and parse.

jacquelinekay: ok, so rather than using `${ARGN}` I should have one parameter which is a list of pairs?

jacquelinekay:
OK, I've implemented the suggestions and corrected the new "language:typesupport" syntax in each rmw implementation where `register_rmw_implementation` is called.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
 
 #
 # Register the current package as a ROS middleware implementation.
+# Takes as input a list of <language:typesupport> pairs where language is the
+# language of the typesupport package and typesupport is the name of the
+# package.
```
dirk-thomas: Please use the docblock-like structure `:param` and `:type` to document the arguments rather then unstructured text. E.g. https://github.com/ament/ament_cmake/blob/c1d3004db7cb33ded7aed38e00c9a9d888986309/ament_cmake_core/cmake/core/list_append_unique.cmake#L20-L21

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
...
+    message(FATAL_ERROR "register_rmw_implementation() called with no arguments!")
   endif()
 
-  ament_index_register_resource("rmw_implementation" CONTENT "${typesupport_package_name}")
-endmacro()
+  # Get the length of ARGN
+  set(argn ${ARGN})
+  set(all_typesupports "")
+
+  foreach(input ${argn})
```
dirk-thomas: You can use `${ARGN}` here. No need for a separate variable.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
...
+      message(FATAL_ERROR
+        "register_rmw_implementation() called with invalid input")
+    endif()
+    list(GET input 0 language_label)
+    list(REMOVE_AT input 0)
+    message(STATUS "Adding resource ${language_label}:${input}")
+    ament_index_register_resource(
+      "rmw_typesupport_${language_label}" CONTENT "${input}"
+    )
+    list_append_unique(${all_typesupports} ${input})
```
dirk-thomas: The first argument is the listname - which shouldn't be evaluated.

rmw/cmake/register_rmw_implementation.cmake
dirk-thomas: I would suggest to use `arg` for a single argument rather then `input` to keep the naming closer.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
...
   endif()
 
-  ament_index_register_resource("rmw_implementation" CONTENT "${typesupport_package_name}")
-endmacro()
+  # Get the length of ARGN
+  set(argn ${ARGN})
+  set(all_typesupports "")
+
+  foreach(input ${argn})
+    string(TOLOWER ${input} input)
```
dirk-thomas: `TOLOWER` shouldn't be applied to the package name. While we use lower case all over the place this would be the only place which converts the case like this.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
...
+    # replace colon with semicolon to turn into a list
+    string(REGEX REPLACE ":" ";" input ${input})
+    list(LENGTH input input_length)
+    if(${input_length} LESS 1)
+      message(FATAL_ERROR
+        "register_rmw_implementation() called with invalid input")
```
dirk-thomas: The error message should state which argument is invalid.

rmw/cmake/register_rmw_implementation.cmake
jacquelinekay: true, I think TOLOWER should still be used on the "language label" though so that "C:rosidl_typesupport_c" and "c:rosidl_typesupport_c" are interpreted the same way.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
...
-  ament_index_register_resource("rmw_implementation" CONTENT "${typesupport_package_name}")
-endmacro()
+  # Get the length of ARGN
+  set(argn ${ARGN})
+  set(all_typesupports "")
+
+  foreach(input ${argn})
+    string(TOLOWER ${input} input)
+    # replace colon with semicolon to turn into a list
+    string(REGEX REPLACE ":" ";" input ${input})
```
dirk-thomas: `${input}` should be put in quotes to not rely in CMake to do the right think if the string items would contain spaces or are empty or something like that.

Same in line 44.

rmw/cmake/register_rmw_implementation.cmake
dirk-thomas: Since that part maps to a folder name it would be good to enforce a case since some file systems might not be able to distinguish them anyway.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -14,18 +14,49 @@
...
-# :param typesupport_package_name: the package name providing type support for
-#   for the registered RMW implementation
-# :type typesupport_package_name: string
+# For example, a valid usage of this function would be
+# register_rmw_implementation("c:rosidl_typesupport_introspection_c"
+#                             "cpp:rosidl_typesupport_introspection_cpp")
+# If there were multiple valid typesupports for one language in a package,
+# they would be separated by colons, e.g.
+# register_rmw_implementation(
+#   "cpp:rosidl_typesupport_a_cpp:rosidl_typesupport_b_cpp")
```
dirk-thomas: Should `"c:foo" "c:bar"` work as well? Since it may be easier to construct programmatically?

jacquelinekay: ok, I left this out because I am lazy and want to write less cmake, but I can do it

dirk-thomas: Either or is fine by me but anyway the code should check if the input matches the expectation.

jacquelinekay: After thinking about it some more I don't think supporting that usage pattern is necessary given a) how rare it is to support multiple typesupport packages for the same language (we don't have any existing cases and I don't see why you would want to do it) and b) how much Cmake boilerplate would be needed to store the lists of labels (e.g. faking a dictionary) and c) the current pattern does address multiple typesupports per language, it just would require accumulating a list of strings and separating them by colons... If we ever have a case of supporting multiple type support libraries per language then I will readdress this function.

jacquelinekay: Sorry I didn't see your second comment until just a little while ago--how does the error checking now?

5175ce2
f9b92f8

jacquelinekay:
sorry @dirk-thomas, I wasn't getting email updates for your comments on this thread for some reason. Does the style look better in 137e44a?

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -15,17 +15,56 @@
...
+  set(all_typesupports "")
+  set(language_labels "")
+
+  foreach(arg ${ARGN})
+    # replace colon with semicolon to turn into a list
+    string(REGEX REPLACE ":" ";" arg "${arg}")
+    list(LENGTH arg arg_length)
+    if(${arg_length} LESS 1)
+      message(FATAL_ERROR
+        "register_rmw_implementation() called with invalid input: '${arg}'")
```
dirk-thomas: This line doesn't need quoting imo since the value is already clearly separated by the colon (and can't be empty).

The error message should print the unmodified arg (without the colon being replaced already). Otherwise the user might wonder where the change is coming from.

jacquelinekay: 7c31349

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -15,17 +15,56 @@
...
+  set(all_typesupports "")
+  set(language_labels "")
+
+  foreach(arg ${ARGN})
+    # replace colon with semicolon to turn into a list
+    string(REGEX REPLACE ":" ";" arg "${arg}")
+    list(LENGTH arg arg_length)
+    if(${arg_length} LESS 1)
+      message(FATAL_ERROR
+        "register_rmw_implementation() called with invalid input: ${ARGN}")
```
dirk-thomas: This change doesn't tell the user which argument was the problem.

jacquelinekay: got it--I'm saving the argument with colons in it now: 18a7e1d

rmw/cmake/get_rmw_typesupport.cmake
```diff
@@ -13,18 +13,29 @@
...
+# Get the type support package names for a specific RMW implementation.
 #
-# :param rmw_implementation: the package name of the ROS middleware
-#   implementation
+# :param var: the output variable name containing the package names
+# :type var: list of strings
+# :param rmw_implementation: the package name of the RMW implementation
 # :type rmw_implementation: string
+# :param LANGUAGE: Optional flag for the language of the type support to get.
+#   If language is omitted, type supports for all packages are returned.
```
dirk-thomas: Should this be: "If language is omitted, type supports for all **languages** are returned."?

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -15,17 +15,56 @@
...
 #
 # @public
 #
-macro(register_rmw_implementation typesupport_package_name)
-  if(NOT "${ARGN} " STREQUAL " ")
-    message(FATAL_ERROR "register_rmw_implementation() called with "
-      "unused arguments: ${ARGN}")
+function(register_rmw_implementation)
+  if("${ARGN}" STREQUAL "")
+    message(FATAL_ERROR "register_rmw_implementation() called with no arguments!")
```
dirk-thomas: It looks like that the function is able to work as-is if no arguments are passed (aka no language-typesupport tuples are specified). Therefore I would suggest to remove this check. Then callers don't need to explicitly check for this and avoid calling the function.

jacquelinekay: But what about line 69, which would set an empty entry in the ament index if there are no arguments--should that be conditioned on the length of the arguments?

rmw/cmake/get_rmw_typesupport.cmake
jacquelinekay: Yes.

rmw/cmake/register_rmw_implementation.cmake
dirk-thomas: I don't see that as a problem. The caller registers the entry with an empty content which is exactly what he is intending when calling this function with no args.

jacquelinekay: ok, fine with me--fixed in 4aeff56

jacquelinekay:
Latest CI:

http://ci.ros2.org/job/ci_linux/1232/
http://ci.ros2.org/job/ci_osx/989/

Fixed the linter warnings and launched a Windows job (pending):
http://ci.ros2.org/job/ci_windows/1273

jacquelinekay:
Had to fix a rebase, the latest Windows job is clean as far as Windows CI goes: http://ci.ros2.org/job/ci_windows/1274

rmw/cmake/get_rmw_typesupport.cmake
```diff
@@ -13,18 +13,29 @@
...
+  cmake_parse_arguments(ARG "" "LANGUAGE" "" ${ARGN})
+  if(ARG_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "get_rmw_typesupport() called with unused "
+      "arguments: ${ARG_UNPARSED_ARGUMENTS}")
   endif()
-
-  ament_index_get_resource(${var} "rmw_implementation" "${rmw_implementation}")
-endmacro()
+  set(resource_type "rmw_typesupport")
+  if(NOT "${ARG_LANGUAGE}" STREQUAL "")
```
dirk-thomas: Needs trailing spaces in both strings.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -15,17 +15,52 @@
...
+  set(all_typesupports "")
+  set(language_labels "")
 
-  ament_index_register_resource("rmw_implementation" CONTENT "${typesupport_package_name}")
-endmacro()
+  foreach(arg_raw ${ARGN})
+    # replace colon with semicolon to turn into a list
+    string(REGEX REPLACE ":" ";" arg "${arg_raw}")
+    list(LENGTH arg arg_length)
+    if(${arg_length} LESS 1)
```
dirk-thomas: Commonly for a variable which contains a number it isn't evaluated: `if(arg_length LESS 1)`

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -15,17 +15,52 @@
...
-  endif()
+function(register_rmw_implementation)
+  set(all_typesupports "")
+  set(language_labels "")
 
-  ament_index_register_resource("rmw_implementation" CONTENT "${typesupport_package_name}")
-endmacro()
+  foreach(arg_raw ${ARGN})
+    # replace colon with semicolon to turn into a list
+    string(REGEX REPLACE ":" ";" arg "${arg_raw}")
```
dirk-thomas: Since this is not a regex a simple string replace is sufficient.

rmw/cmake/register_rmw_implementation.cmake
```diff
@@ -15,17 +15,52 @@
...
+    list(REMOVE_AT arg 0)
+    string(TOLOWER "${language_label}" language_label)
+    # Cache the existing languages labels found (so that we don't have to
+    # check the ament index twice every loop) and error if we've already
+    # encountered the language
+    list(FIND language_labels "${language_label}" label_found)
+    if(NOT ${label_found} EQUAL -1)
+      message(FATAL_ERROR
+        "register_rmw_implementation() got language '${language_label}'"
+        " multiple times")
```
dirk-thomas: Instead of the leading space in this line we commonly use a trailing space in the previous line.

jacquelinekay:
bump... any pending comments on this set of PRs?

dirk-thomas:
Beside mini comments on some of the related PRs this LGTM.

PR #64, opened 2016-04-14
rework get_rmw_typesupport

dirk-thomas:
@jacquelinekay this is an update for #63 which reworks `get_rmw_typesupport.cmake`.


jacquelinekay:
thanks, that looks better to me

PR #65, opened 2016-04-27
new api to support wait_for_service

wjwwood:
This is my proposed API additions needed to support the `wait_for_service` functionality at the client library level. A similar set of functions would be in `rcl` (or else aliased to these functions), and the client libraries can implement `wait_for_service` using these two functions.

Looking for feedback before/while fixing up changes to each rmw implementation to implement these additions.

Connects to ros2/ros2#215

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -57,6 +57,33 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_destroy_node(rmw_node_t * node);
 
+/// Return a guard condition which is triggered when the ROS graph changes.
+/* The handle returned is a pointer to an internally held rmw guard condition.
+ * This function can fail, and therefore return NULL, if:
+ *   - node is NULL
+ *   - node is invalid
+ *
+ * The returned handle is made invalid if the node is destroyed or if
+ * rmw_shutdown() is called.
+ *
+ * The guard condition will be triggered anytime change to the ROS graph occurs.
+ * A ROS graph change includes things like (but not limited to) a new publisher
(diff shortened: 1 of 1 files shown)


wjwwood:
I'm putting this up for review.

tfoote:
+1

rmw/include/rmw/rmw.h
```diff
@@ -277,6 +304,38 @@ RMW_WARN_UNUSED
...
+ * available for the given client.
+ *
+ * The node parameter must not be NULL, and must point to a valid node.
+ *
+ * The client parameter must not be NULL, and must point to a valid client.
+ *
+ * The given client and node must match, i.e. the client must have been created
+ * using the given node.
+ *
+ * The is_available parameter must not be NULL, and must point a bool variable.
```
dhood: point a bool -> point to a bool

rmw/include/rmw/rmw.h
```diff
@@ -57,6 +57,33 @@ RMW_WARN_UNUSED
...
+/// Return a guard condition which is triggered when the ROS graph changes.
+/* The handle returned is a pointer to an internally held rmw guard condition.
+ * This function can fail, and therefore return NULL, if:
+ *   - node is NULL
+ *   - node is invalid
+ *
+ * The returned handle is made invalid if the node is destroyed or if
+ * rmw_shutdown() is called.
+ *
+ * The guard condition will be triggered anytime change to the ROS graph occurs.
```
wjwwood: Thanks: f758ef6

rmw/include/rmw/rmw.h
wjwwood: Thanks: f758ef6

PR #67, opened 2016-04-30
Check if ament index resource exists before accessing

jacquelinekay:
This changes the behavior of `get_rmw_typesupport` slightly so that it returns an empty list rather than erroring if no typesupport is found for an rmw_implementation/language label.

Connects to ros2/rclpy#21


rmw/cmake/get_rmw_typesupport.cmake
```diff
@@ -36,6 +36,11 @@ function(get_rmw_typesupport var rmw_implementation)
...
     set(resource_type "${resource_type}_${ARG_LANGUAGE}")
   endif()
-  ament_index_get_resource(resource "${resource_type}" "${rmw_implementation}")
-  set(${var} "${resource}" PARENT_SCOPE)
+  ament_index_has_resource(has_resource "${resource_type}" "${rmw_implementation}")
+  if(${has_resource})
+    ament_index_get_resource(resource "${resource_type}" "${rmw_implementation}")
+    set(${var} "${resource}" PARENT_SCOPE)
+  else()
+    set(${var} " " PARENT_SCOPE)
```
dirk-thomas: Why a space?

jacquelinekay: because I was thinking of how we do empty string compare in CMake (with a trailing space) and got mixed up

dirk-thomas: Also only set the `resource` variable in the else case and then do the set in parent scope only once outside of the conditional block.

jacquelinekay: Like in 65f7c70?

dirk-thomas:
LGTM.

PR #69, opened 2016-06-07
Removed -std=c++11 flag

esteve:
Removed -std=c++11 flag and added to `rmw_fastrtps_cpp`

### DIFF
--- a/rmw/cmake/configure_rmw_library.cmake
+++ b/rmw/cmake/configure_rmw_library.cmake
@@ -42,7 +42,7 @@ macro(configure_rmw_library library_target)
     # (this is already the default on Windows)
     set_target_properties(${library_target}
       PROPERTIES
-        COMPILE_FLAGS "-fvisibility=hidden -fvisibility-inlines-hidden -std=c++11"
+        COMPILE_FLAGS "-fvisibility=hidden -fvisibility-inlines-hidden"
     )
   endif()
   if(WIN32)


wjwwood:
Why does this need to be removed?

wjwwood:
Not to say it shouldn't, it's just not clear to me why.

esteve:
@wjwwood the `-std=c++11` flag is only valid for C++ in Clang, which fails to compile `rmw` for Android:

```
/opt/android/android-ndk-r11c/toolchains/llvm/prebuilt/linux-x86_64/bin/clang  -DANDROID -Drmw_EXPORTS -fexceptions -gcc-toolchain /opt/android/android-ndk-r11c/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64 -target armv7-none-linux-androideabi -Qunused-arguments --sysroot=/opt/android/android-ndk-r11c/platforms/android-21/arch-arm -funwind-tables -fsigned-char -no-canonical-prefixes -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16 -Xclang -mnoexecstack  -marm -fno-omit-frame-pointer -fno-strict-aliasing -O0 -g -DDEBUG -D_DEBUG -fPIC -isystem /opt/android/android-ndk-r11c/platforms/android-21/arch-arm/usr/include -isystem /opt/android/android-ndk-r11c/sources/cxx-stl/gnu-libstdc++/4.9/include -isystem /opt/android/android-ndk-r11c/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi-v7a/include -isystem /opt/android/android-ndk-r11c/sources/cxx-stl/gnu-libstdc++/4.9/include/backward -I/home/esteve/Projects/android_ros2/src/ros2/rmw/rmw/include    -fvisibility=hidden -fvisibility-inlines-hidden -std=c++11 -o CMakeFiles/rmw.dir/src/allocators.c.o   -c /home/esteve/Projects/android_ros2/src/ros2/rmw/rmw/src/allocators.c
error: error: invalid argument invalid argument '-std=c++11' '-std=c++11'not  allowednot  withallowed  'C/ObjC'with 
'C/ObjC'
```

this PR would need to be merged at the same time as https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/39, though.

wjwwood:
@esteve personally I'd prefer to expand the existing exclusions in that file to avoid `-std=c++11` when building for Android rather than moving where it is defined. We already avoid that flag (and others) on anything that's not gcc or clang (msvc).

Alternatively we could use CMake to check to see if the compiler supports the flag: https://cmake.org/cmake/help/v3.0/module/CheckCCompilerFlag.html

esteve:
@wjwwood the problem lies in that the CMake toolchain defines `CMAKE_CXX_COMPILER_ID` and that the `COMPILE_FLAGS` property applies to both C and C++. If you want to keep the `std=c++11` flag, an option could be to use `target_compile_options` instead of `set_target_properties`, like the following, to ensure that the `std=c++11` flag only applies to C++

```
target_compile_options(${library_target}
  PRIVATE
  $<$<COMPILE_LANGUAGE:C>:-fvisibility=hidden -fvisibility-inlines-hidden>
  $<$<COMPILE_LANGUAGE:CXX>:-fvisibility=hidden -fvisibility-inlines-hidden -std=c++11>
)
```
generator expressions for `target_compile_options` are available in CMake >= 3.3, but it won't be a problem if ROS2 switches to Xenial)

However, I'm afraid I still don't see why the flag is needed:
- the comment does not match the code, since the former only mentions visibility (https://github.com/ros2/rmw/blob/master/rmw/cmake/configure_rmw_library.cmake#L41-L42)
- dictates rmw implementations to use C++11. I think it should be up to the `rmw` implementation to decide what version of C++ to use (i.e. 11, 14, 17)
- both `rmw_opensplice` (https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/CMakeLists.txt#L6) and `rmw_connext` (https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/CMakeLists.txt#L9) (https://github.com/ros2/rmw_connext/blob/master/rmw_connext_dynamic_cpp/CMakeLists.txt#L9) already set the flag themselves. It's only `rmw_fastrtps` that does not set it, which is why I added it there (https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/39/files#diff-1c2560c14bf3483f4f59c419be91083fR6)
- ROS2 builds just fine without the `std=c++11` flag in `configure_rmw`

esteve:
Just to add something more to the discussion, the current `CMakeLists.txt` for this project calls `configure_rmw` (https://github.com/ros2/rmw/blob/master/rmw/CMakeLists.txt#L13). However, since `rmw` is declared to be a C project (https://github.com/ros2/rmw/blob/master/rmw/CMakeLists.txt#L3), the code in https://github.com/ros2/rmw/blob/master/rmw/cmake/configure_rmw_library.cmake#L40-L47 never gets called on Linux, because `CMAKE_CXX_COMPILER_ID` is not defined. I don't know if that's intentional, though.

wjwwood:
You make a lot of good points. I think the comment could easily be updated, and for `rmw_connext` and `rmw_opensplice` setting the `-std` flag themselves, I think that's for the tests which also use c++11. Also ROS2 only builds fine when removed if you also add it to the FastRTPS implementation.

We're now updated to Xenial and require cmake 3.5, so we could use `target_compile_options` to do the language specific application of the flags as you suggested, but I'm not sure we need that now if we're just passing the visibility flags.

However, I think the idea that the rmw implementations should choose their `-std=` value is the most compelling. Based on that, I'm ok with removing this here and having rmw implementations explicitly set it.

wjwwood:
I started CI with a custom gist (https://gist.githubusercontent.com/wjwwood/fa595f1f785990c81704c28f5f6ba3aa/raw/6af9ae91957b4836899e0e0593aa213351d10069/ros2.repos):

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1514/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1179/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1516/)

dirk-thomas:
The forks are out of date and need to be rebased. The builds show compiler warnings which have been fixed since then and likely the same for some tests.

wjwwood:
I don't see any "new" warnings or test failures that I can attribute to these pr's. So, since I cannot do the rebase myself without opening new pr's, I'm going to merge them as-is, run CI on master, and address anything I notice there.

wjwwood:
Follow up CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1522/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1186/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1520/)

wjwwood:
The OS X job was ok (it hung on the FastRTPS test but the build was ok).

esteve:
@wjwwood thanks.

PR #71, opened 2016-06-22
update compiler check to match AppleClang

dirk-thomas:
With using CMake 3.5 and https://cmake.org/cmake/help/v3.0/policy/CMP0025.html we need to match `AppleClang`.

http://ci.ros2.org/job/ci_osx/1176/


wjwwood:
lgtm

PR #73, opened 2016-09-16
fix compiler warning on Windows

dirk-thomas:
See https://github.com/ros2/rmw/commit/1b28064c8ba75809a2e097a4d8951331022803f9#commitcomment-19025901


dirk-thomas:
Please review.

Issue #77, opened 2016-11-11
Potentially risky to use transient_local durability for services


dhood:
@wjwwood and I were talking yesterday and realised that the default QoS profile for services might lead to undesired behaviour. 

The default durability [is `RMW_QOS_POLICY_TRANSIENT_LOCAL_DURABILITY`](https://github.com/ros2/rmw/blob/a7436c86d477665c38e0122ef71598835d6560f1/rmw/include/rmw/qos_profiles.h#L54). @wjwwood pointed out that this was set before `wait_for_service` was implemented, as a way to compensate for race conditions in server/client bringup.

With this setting, in the case of a server that starts, dies, and restarts, it will receive the service request from the client twice. **While the client is protected from receiving multiple service responses, the server is not protected from side effects of receiving multiple service calls.**

If a service does anything like write to file, increment a counter, toggle a value, etc, the transient_local setting is probably not a good idea.

A better default might be volatile, and we should be careful about exposing the transient_local setting to users.

wjwwood:
+1 to all of this.

I'd actually like to either not expose this QoS setting in the user facing Service interface or otherwise somehow enforce that they do not use something other than volatile. Simply because it will still work, but have this subtle and unsafe behavior.

--

As a side anecdote, we toggled our blade state with a Service call, having a duplicate request (even if the response was discarded) would have been a terrible surprise.

wjwwood:
Err, I should have said "on our autonomous lawnmower, we toggled our blade state..."

dhood:
closed with #79

PR #79, opened 2016-12-08
Update default service QoS durability to volatile

dhood:
Connects to ros2/rmw#77

As outlined in https://github.com/ros2/rmw/issues/77 with transient_local, service calls may be received inappropriately multiple times. This PR changes the default for services to be volatile.

This will affect both the datawriter and datareader on both the client and the server.

I will open a corresponding PR against this article http://design.ros2.org/articles/qos.html

what do you think about adding the following warning to `rcl`?

```
Warning: Setting QoS durability to 'transient local' for service servers can cause them to receive requests from clients that have since terminated.
```

I think clients should be protected against receiving requests they didn't make anyway so one for clients shouldn't be necessary.

---

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1992/)
OS X: [Build Status](http://ci.ros2.org/job/ci_osx/1533/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1965/)

There are two new test failures on OS X and Windows:
`test_services__rmw_fastrtps_cpp` from RCL and 
`test_requester_replier__Primitives__rclpy__rmw_fastrtps_cpp_Release.test_requester_replier` from `test_communication`

These fail in nightlies sometimes on Windows e.g. http://ci.ros2.org/view/nightly/job/nightly_win_rep/484/#showFailuresLink and http://ci.ros2.org/view/nightly/job/nightly_win_rep/483/#showFailuresLink respectively but I couldn't find it fail on osx before (it's a 15s timeout and it usually takes just a few seconds). I ran the above CI jobs with `--retest-until-pass 10` instead of the usual 5 and that still didn't fix it but I will keep trying... hopefully we're not accidentally relying on this.

### DIFF
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -51,7 +51,7 @@ static const rmw_qos_profile_t rmw_qos_profile_services_default =
   RMW_QOS_POLICY_HISTORY_KEEP_LAST,
   10,
   RMW_QOS_POLICY_RELIABILITY_RELIABLE,
-  RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL
+  RMW_QOS_POLICY_DURABILITY_VOLATILE
 };
 
 static const rmw_qos_profile_t rmw_qos_profile_parameter_events =


wjwwood:
The proposed warning lgtm, and it's very possible that we were relying on that timing somewhere, but we should be able to update those places with proper usage of `wait_for_service` now, but I don't have time to look into it personally at the moment.

+1 in general to changing the default to `volatile` for services.

dirk-thomas:
The failing tests even after 10 repeated tries don't sound like a flaky test anymore but like a regression. I would suggest to investigate those before considering to merge this.

Besides that: absolute change the default qos for service to being volatile :+1:

dhood:
as mentioned we are going to merge this so that the defaults change, knowing that it will introduce test failures for services in some packages, but that services are being tested without issues in other areas of the codebase.

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2051/)
OS X: [Build Status](http://ci.ros2.org/job/ci_osx/1584/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2070/)

The tests that are failing are 
`projectroot.test_services__rmw_fastrtps_cpp` from `rcl` and 
`projectroot.rosidl_generator_py.test_requester_replier__{msg_type}__rclpy__rmw_fastrtps_cpp` from `test_communication` using rclpy

PR #80, opened 2016-12-08
Add support for pthread-based thread local storage

esteve:
This PR adds support for pthread-based thread local storage on platforms that don't have compiler builtin support for it. I only enabled it for iOS < 10.0, but it can be extended to any other platform via CMake flags.

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -29,6 +29,10 @@ ament_export_dependencies(rosidl_generator_c)
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 
+if(IOS AND IOS_SDK_VERSION LESS 10.0)
+  ament_export_libraries(pthread)
+endif()
+
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -22,7 +22,7 @@ set(rmw_sources
 set_source_files_properties(
   ${rmw_sources}
   PROPERTIES LANGUAGE "C")
-add_library(${PROJECT_NAME} SHARED ${rmw_sources})
+add_library(${PROJECT_NAME} ${rmw_sources})
 configure_rmw_library(${PROJECT_NAME} LANGUAGE "C")
 
 ament_export_dependencies(rosidl_generator_c)
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -22,7 +22,7 @@ set(rmw_sources
 set_source_files_properties(
   ${rmw_sources}
   PROPERTIES LANGUAGE "C")
-add_library(${PROJECT_NAME} ${rmw_sources})
+add_library(${PROJECT_NAME} SHARED ${rmw_sources})
 configure_rmw_library(${PROJECT_NAME} LANGUAGE "C")
 
 ament_export_dependencies(rosidl_generator_c)
--- a/rmw/include/rmw/macros.h
+++ b/rmw/include/rmw/macros.h
@@ -15,22 +15,36 @@
 #ifndef RMW__MACROS_H_
 #define RMW__MACROS_H_
 
+// This block either sets RMW_THREAD_LOCAL or RMW_THREAD_LOCAL_PTHREAD.
 #if defined _WIN32 || defined __CYGWIN__
+// Windows or Cygwin
   #define RMW_THREAD_LOCAL __declspec(thread)
 #elif defined __APPLE__
-  #include <Availability.h>
-  #if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
-    // iOS 10 added support for thread local storage
-    #if __IPHONE_OS_VERSION_MAX_ALLOWED < 100000
-      #define RMW_THREAD_LOCAL_PTHREAD 1
-      #define RMW_THREAD_LOCAL
+// Apple OS's
+  #include <TargetConditionals.h>
+  #if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+// iOS Simulator or iOS device
+    #include <Availability.h>
+    #if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
+      #if __IPHONE_OS_VERSION_MAX_ALLOWED >= 100000
+// iOS >= 10, thread local storage was added in iOS 10
+        #define RMW_THREAD_LOCAL _Thread_local
+      #else
+// iOS < 10, no thread local storage, so use pthread instead
+        #define RMW_THREAD_LOCAL_PTHREAD 1
+        #undef RMW_THREAD_LOCAL
+      #endif
+    #else
+      #error "Unknown iOS version"
     #endif
-  #endif
-
-  #ifndef RMW_THREAD_LOCAL_PTHREAD
+  #elif TARGET_OS_MAC
+// macOS
     #define RMW_THREAD_LOCAL _Thread_local
+  #else
+    #error "Unknown Apple platform"
   #endif
 #else
+// Some other non-Windows, non-cygwin, non-apple OS
   #define RMW_THREAD_LOCAL _Thread_local
 #endif
 
--- a/rmw/include/rmw/macros.h
+++ b/rmw/include/rmw/macros.h
@@ -17,8 +17,21 @@
 
 #if defined _WIN32 || defined __CYGWIN__
   #define RMW_THREAD_LOCAL __declspec(thread)
+#elif defined __APPLE__
+  #include <Availability.h>
+  #if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
+    // iOS 10 added support for thread local storage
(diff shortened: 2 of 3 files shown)


mikaelarguedas:
@ros2/team review this please

rmw/include/rmw/macros.h
```diff
@@ -17,8 +17,21 @@
...
+      #define RMW_THREAD_LOCAL
+    #endif
+  #endif
+
+  #ifndef RMW_THREAD_LOCAL_PTHREAD
+    #define RMW_THREAD_LOCAL _Thread_local
+  #endif
 #else
-  #define RMW_THREAD_LOCAL __thread
+  #define RMW_THREAD_LOCAL _Thread_local
```
wjwwood: Why this change for the non-iOS<10 and non-windows case?

rmw/include/rmw/macros.h
```diff
@@ -17,8 +17,21 @@
 
 #if defined _WIN32 || defined __CYGWIN__
   #define RMW_THREAD_LOCAL __declspec(thread)
+#elif defined __APPLE__
+  #include <Availability.h>
+  #if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
+    // iOS 10 added support for thread local storage
+    #if __IPHONE_OS_VERSION_MAX_ALLOWED < 100000
+      #define RMW_THREAD_LOCAL_PTHREAD 1
+      #define RMW_THREAD_LOCAL
```
wjwwood: So in this case the PTHREAD alternative is used?

wjwwood (approved):
I had a few questions but otherwise it looks solid to me.

I'll start some CI.

wjwwood:
Oh, actually, it looks like it needs to be rebased. I'll do that too.

rmw/CMakeLists.txt
```diff
@@ -21,13 +21,17 @@ set(rmw_sources
 set_source_files_properties(
   ${rmw_sources}
   PROPERTIES LANGUAGE "C")
-add_library(${PROJECT_NAME} SHARED ${rmw_sources})
+add_library(${PROJECT_NAME} ${rmw_sources})
```
wjwwood: @dirk-thomas this is ok now that we have the "default to SHARED" turned on with https://github.com/ros2/ament_cmake_ros/blob/master/ament_cmake_ros/cmake/build_shared_libs.cmake, right?

wjwwood:
Looks like `rmw` is already defaulting to non-shared and causing issues:

```
02:10:57 /usr/bin/ld: /home/rosbuild/ci_scripts/ws/install/rmw/lib/librmw.a(error_handling.c.o): relocation R_X86_64_TPOFF32 against `__rmw_error_state' can not be used when making a shared object; recompile with -fPIC
```

-- http://ci.ros2.org/job/ci_linux/2410/consoleFull

@dirk-thomas what is blocking us from removing `SHARED` from the `add_library` call and still getting dynamic libraries by default?

@esteve since you're probably wanting to use this in the static linking mode at some point, I guess we should wait to merge this until the above issue is investigated?

rmw/CMakeLists.txt
dirk-thomas: No, this package does currently not use `ament_cmake_ros` since #81 was reverted. Therefore the `SHARED` needs to stay.

wjwwood:
All of the CI I just ran, in case it's helpful for debugging:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2410/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/46/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1859/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2529/)

rmw/CMakeLists.txt
wjwwood: Should we just wait for #81 or should I try this change without removing `SHARED`? I guess it's not relevant to the rest of the thread local related changes.

@esteve I suppose that's up to you. I can easily remove this part of the patch and rerun CI for you.

rmw/include/rmw/macros.h
esteve: C11 added support for TLS via `_Thread_local`, `__thread` is a GCC extension that Clang supports for compatibility reasons. I can revert it to `__thread`, but I thought that given that IIRC ROS2 now requires a C11-capable compiler, it might be nice to use `_Thread_local` instead.

rmw/CMakeLists.txt
esteve: I've readded the `SHARED` keyword in https://github.com/ros2/rmw/pull/80/commits/911bb776da7f778c6ecfb1c36c3a0f27fa073949 until I can figure out why the `ament_cmake_ros` changes don't work.

rmw/include/rmw/macros.h
esteve: Exactly. If we happen to be targeting iOS < 10.0, the `pthread`-based implementation of TLS will be used (`RMW_THREAD_LOCAL_PTHREAD` is defined). This can be expanded to other platforms that have an implementation of `pthread`, but for now it's only for iOS < 10.0.

wjwwood:
Ok @esteve, here's new CI with your latest commit:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2418/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/51/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1871/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2543/)

rmw/include/rmw/macros.h
wjwwood: The MSVC compiler is far from C11 compliant, AFAIK. So I think that we do not require a C11 compiler. It has been turned on in a few of the packages, but I had explicitly tried to avoid it in `rmw`. That being said, we already have to special case here for Windows, and the other compilers are likely C11 compliant. I'll either leave this as-is or try to detect when `_Thread_local` is not available and fallback further to `__thread`.

rmw/include/rmw/macros.h
wjwwood: After thinking about it, I think this might be somewhat dangerous because if someone is using `RMW_THREAD_LOCAL`, but doesn't special case it for `RMW_THREAD_LOCAL_PTHREAD` then they might end up with something that assumes it is thread local, but actually isn't because `RMW_THREAD_LOCAL` is just empty. I think I'll do `#undef RMW_THREAD_LOCAL` instead.

wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2451/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/75/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1900/) (known compiler warning)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2570/)

wjwwood:
Thanks @esteve!

esteve:
@wjwwood thanks for fleshing out this PR. Just a minor comment, I remember deciding not using `TargetConditionals.h` (https://github.com/ros2/rmw/pull/80/commits/df1500bb35031bf4abd2504b3274c3f3ed4649c9), but unfortunately I forgot why. I think it was because `TargetConditionals.h` was not available in certain versions of the iOS SDK or because some of its definitions are deprecated, but nevertheless, if any issue shows up, it can be addressed in the future.

Thanks again for testing this!

wjwwood:
> @wjwwood thanks for fleshing out this PR. Just a minor comment, I remember deciding not using TargetConditionals.h (df1500b), but unfortunately I forgot why. I think it was because TargetConditionals.h was not available in certain versions of the iOS SDK or because some of its definitions are deprecated, but nevertheless, if any issue shows up, it can be addressed in the future.

Doh. That's what I get for trying to make it better. Ok we can address it if we need to in the future. Hopefully it works for some cases now.

> Thanks again for testing this!

No worries, I wanted to get it in before https://github.com/ros2/rmw/pull/101 because this whole code block is getting moved to another package for reuse in other places.

esteve:
@wjwwood oh sorry, I hope I didn't sound ungrateful, the code definitely looks much better now. It's actually my fault for not documenting why I didn't use `TargetConditionals.h`, now I forgot why I opted not to use it :smile:

wjwwood:
> @wjwwood oh sorry, I hope I didn't sound ungrateful

Not at all.

wjwwood:
@esteve do you know if pthread is still required for iOS? I thought I read that it's no longer required.

The reason I'm asking is that I'm refactoring the error handling code to no longer allocate memory when setting an error (trying to improve memory management) and the pthread specific code is making things really difficult.

I'm simultaneously looking into whether or not other potential platforms need pthread support still as well.

esteve:
I recall needing `pthread` for TLS with a Linaro toolchain a while ago, but I imagine that's no longer a problem. As for iOS, C11-based TLS is supported since 10. If this code needs to be revived in the future, I can take a look at it, but for now... KILL IT WITH FIRE

ros-discourse:

https://discourse.ros.org/t/announcing-objective-c-on-macos-and-ios-for-ros2/946/1

PR #82, opened 2016-12-08
Update default QoS history to keep last

dhood:
With keep_all (as is), the depth is ignored (Section `2.2.3.18 HISTORY` of http://www.omg.org/spec/DDS/1.4/PDF) and we have what is akin to an unlimited queue.

This can lead to resource issues (especially using Fast RTPS which has a default history cache size of 5000).

This "addresses" ros2/rmw_fastrtps#68 in the sense that it won't happen using the default QoS settings (it does not fix it because people can still use the QoS settings). 
There is no regression test for ros2/rmw_fastrtps#68, I have only manually confirmed.

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2004/)
OS X: [Build Status](http://ci.ros2.org/job/ci_osx/1535/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1969/)

### DIFF
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -40,7 +40,7 @@ static const rmw_qos_profile_t rmw_qos_profile_parameters =
 
 static const rmw_qos_profile_t rmw_qos_profile_default =
 {
-  RMW_QOS_POLICY_KEEP_ALL_HISTORY,
+  RMW_QOS_POLICY_KEEP_LAST_HISTORY,
   10,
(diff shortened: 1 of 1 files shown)


dhood:
also, there's this note in the dds spec (`2.2.3.4 DURABILITY`) that suggests that this might not be a good default setting (for resource reasons) even with volatile:


> Note that
> although related, this does not strictly control what data the Service will maintain internally. That is, the Service may choose to
> maintain some data for its own purposes (e.g., flow control) and yet not make it available to late-joining readers if the
> DURABILITY QoS policy is set to VOLATILE.

PR #87, opened 2016-12-13
move getenv from rcl to here for more reuse

wjwwood:
Connects to ros2/rmw_fastrtps#79


dirk-thomas:
I don't think `rmw` is a good location for those since it has nothing to do with the ROS middleware interface. Maybe a new package for those functions would be a good idea (so that it can be used from rmw implementations as well as rcl*).

Karsten1987:
btw, should we close this in favor of https://github.com/ros2/c_utilities/issues/2 ?

wjwwood:
@Karsten1987 I'd say that we should use this implementation (documentation, signature, and implementation, if they're not already identical to what's in ros2/c_utilities#2) before this is closed. I spent a good deal of time tweaking this implementation to work the same on Windows and Unix without any inconsistencies and avoiding memory allocation if possible.

wjwwood:
Or at least as nearly as possible, given that this implementation uses `RMW_*` defines.

mikaelarguedas:
the implementation in https://github.com/ros2/c_utilities/pull/2 will be identical on behaviour and argument type. The return codes will be different though

wjwwood:
Closing in favor of https://github.com/ros2/c_utilities/pull/2

PR #88, opened 2016-12-15
Doxygen setup

wjwwood:
This pull request sets up Doxygen and fixes the documentation to look correct when rendered and to reference external data, like cppreference.com (where appropriate).

I've pushed the current state up to http://docs.ros2.org/beta1/api/rmw/ and updated the rcl stuff to cross reference the contents. I'll update this one when I've added docs for the packages it relies on.

### DIFF
--- a/rmw/.gitignore
+++ b/rmw/.gitignore
@@ -1,0 +1,1 @@
+doc_output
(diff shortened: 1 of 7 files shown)


wjwwood:
Now that the beta is out, I'd like to get these basic configurations and documentation updates merged. Looking for a review.

PR #90, opened 2017-01-26
expose list of nodes

Karsten1987:
Connect to ros2/cli_tools#1

### DIFF
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -1,4 +1,4 @@
-// Copyright 2014 Open Source Robotics Foundation, Inc.
+// Copyright 2014-2017 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -192,11 +192,11 @@ typedef struct RMW_PUBLIC_TYPE rmw_topic_names_and_types_t
   char ** type_names;
 } rmw_topic_names_and_types_t;
 
-typedef struct RMW_PUBLIC_TYPE rmw_node_names_t
+typedef struct RMW_PUBLIC_TYPE rmw_string_array_t
 {
-  size_t node_count;
-  char ** names;
-} rmw_node_names_t;
+  size_t size;
+  char ** data;
+} rmw_string_array_t;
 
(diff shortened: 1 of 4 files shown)


rmw/include/rmw/types.h
```diff
@@ -192,6 +192,12 @@ typedef struct RMW_PUBLIC_TYPE rmw_topic_names_and_types_t
   char ** type_names;
 } rmw_topic_names_and_types_t;
 
+typedef struct RMW_PUBLIC_TYPE rmw_node_names_t
+{
+  size_t node_count;
+  char ** names;
```
dirk-thomas: I would either prefix both members with `node_` or none of them. In the second case the type could also be called `rmw_names_t` and potentially reused for other names in the future.

rmw/src/sanity_checks.c
```diff
@@ -42,3 +42,22 @@ rmw_check_zero_rmw_topic_names_and_types(
   }
   return RMW_RET_OK;
 }
+
+rmw_ret_t
+rmw_check_zero_rmw_node_names(
+  rmw_node_names_t * node_names)
+{
+  if (!node_names) {
+    RMW_SET_ERROR_MSG("node_names_t is null");
```
dirk-thomas: Should be: `node_names is null`

dirk-thomas:
I would suggest connecting all other PRs to this one to group the PRs in waffle as well as have a "master" ticket to comment on with e.g. CI builds.

This also needs a PR for rmw_opensplice so that it at least continues to build / link.

rmw/include/rmw/types.h
Karsten1987: hmm, we could even spin this further and call it `rmw_string_array_t` ?

dirk-thomas: Sounds good to me. At the end its only a simple `vector<string>` in C.

Karsten1987:
any updates on this?

dirk-thomas:
I think it would be good to have a CI job which was run for OpenSplice to ensure it compiles. Also I didn't see any test where it checks that the returned nodes names are correct. I think that would be good to cover too.

Karsten1987:
I added a new test for ensuring that both supported rmw implementations are returning the same issue.
Can be found in here: https://github.com/ros2/rcl/blob/rosnode_list/rcl/test/rcl/test_get_node_names.cpp

build result: [Build Status](http://ci.ros2.org/job/ci_linux/2248/)

PR #101, opened 2017-04-13
moved error handling to c_utilities

wjwwood:
Connects to ros2/c_utilities#16


rmw/include/rmw/error_handling.h
```diff
@@ -23,95 +23,38 @@ extern "C"
 #include <stdbool.h>
 #include <stddef.h>
 
-#include "rmw/macros.h"
```
mikaelarguedas: Should the macros (STRINGIFY and friends) be typedefed to their c_utility equivalent ?

mikaelarguedas: maybe this is captured in https://github.com/ros2/c_utilities/pull/16#issuecomment-294263596, if so feel free to discard this comment

wjwwood: No, I missed that, I also missed removing the thread local stuff, which isn't used in RMW anymore. Good catch, see: 93d9825

rmw/include/rmw/error_handling.h
```diff
@@ -23,95 +23,38 @@ extern "C"
...
 #include "rmw/visibility_control.h"
+#include "c_utilities/error_handling.h"
```
wjwwood: fc2d4ea

PR #102, opened 2017-04-18
Add rmw_node_security_options

mikaelarguedas:
connects to ros2/ros2#339


dirk-thomas:
Does this have to be a separate function or could this be added as an option to the existing `rmw_create_node` function?

mikaelarguedas:
@dirk-thomas Do you mean adding an argument to the `rmw_create_node` function and try to set the security parameters only if it's not `NULL` ?

dirk-thomas:
> Do you mean adding an argument to the rmw_create_node function and try to set the security parameters only if it's not NULL ?

Yes, it is only an option influencing how to configure the to be created node.

rmw/include/rmw/rmw.h
```diff
@@ -90,7 +90,12 @@ rmw_init(void);
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_node_t *
-rmw_create_node(const char * name, const char * namespace_, size_t domain_id);
+rmw_create_node(
+  const char * name,
+  const char * namespace_,
+  size_t domain_id,
+  bool enforce_security,
+  const char * security_root_path);
```
dirk-thomas: With the number of arguments growing I suggest to introduce an *options* struct.

rmw/include/rmw/types.h
```diff
@@ -162,6 +162,18 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
   uint64_t nsec;
 } rmw_time_t;
 
+enum RMW_PUBLIC_TYPE rmw_security_enforcement_policy_t
+{
+  RMW_SECURITY_ENFORCEMENT_PERMISSIVE,
+  RMW_SECURITY_ENFORCEMENT_ENFORCE,
```
mikaelarguedas: Looking for feedback on type + enum names

rmw/include/rmw/rmw.h
```diff
@@ -90,7 +90,11 @@ rmw_init(void);
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_node_t *
-rmw_create_node(const char * name, const char * namespace_, size_t domain_id);
+rmw_create_node(
+  const char * name,
+  const char * namespace_,
+  size_t domain_id,
+  const rmw_node_security_options_t * options);
```
dirk-thomas: At some point I would expect `rmw_create_node` to take other `options`. So these ones might be better described as `security_options`. Or they should be a part of a more general `options` struct.

rmw/include/rmw/types.h
dirk-thomas: Looks good to me as-is (related https://github.com/ros2/rcl/pull/125#discussion_r121730241).

PR #107, opened 2017-06-03
Add C11 compiler flag

esteve:
This fixes a compiler error/warning with the Raspberry Pi cross compiler caused by declaring a variable in the following `for` loops:

https://github.com/ros2/rmw/blob/master/rmw/src/validate_node_name.c#L43
https://github.com/ros2/rmw/blob/master/rmw/src/validate_full_topic_name.c#L59
https://github.com/ros2/rmw/blob/master/rmw/src/validate_full_topic_name.c#L79

alternatively, this issue could be solved by just declaring the index variable outside the for loops.

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.5)
 project(rmw)
 
 if(NOT WIN32)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
 endif()
 


mikaelarguedas:
I remember we discussed this a while ago. While all our targetted compilers do support the declaration of variables inside loops. @ros2/team What do we want to do in general? allow ourselves to use it and add the flag wherever we do so? or always declare variables outside of the loops for better compatibility for platforms and compiuler we don't test ?

dirk-thomas:
I think our [developer guidelines](https://github.com/ros2/ros2/wiki/Developer-Guide#programming-conventions) already answer that question:

> Declare variables in the narrowest scope possible.

dirk-thomas:
Latest CI builds:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2671/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/221/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2126/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2790/)

clalancette:
So, there is one reason I would suggest not to use -std=c11, and that is MSVC.  My understanding is that MSVC doesn't support most of C11 (with declaration-inside-loops being one of things it does support).  Thus, to keep our platform MSVC compliant, it seems better to me to not add -std=c11 and not depend on C11 features.

dirk-thomas:
Since all our platforms support this feature (declaring the variable inside the loop) I don't see any reason why not to utilize that.

For other features which might be C11 specific: our rule that a change needs to build on all supported platforms still applies. So C11 features not supported on Windows can't be used.

dirk-thomas:
While technically one approval is sufficient I will leave the merge to a second reviewer.

Karsten1987:
I have the feeling that we should somehow come to a common consensus on
that decision and mark it on a central place as we had this discussions on
multiple PRs already.

mikaelarguedas:
> I am in favor of defining the variables out of the for loops and not adding C11 flags, so I won't approve this pr as-is.

I agree with this. We officially target c99 and don't advertise anywhere that we use "some features" of C11.
I think we should stick to c99 (as it was the case until a few months ago), this will make it easier for people trying to get the stack to compile on different compilers. To make an educated decision I'd be interested in knowing which features of C11 supported by all compilers we want to use. (because I don't think that this one is a very compelling one of sufficient to base our decision on).

> I have the feeling that we should somehow come to a common consensus on
that decision and mark it on a central place as we had this discussions on
multiple PRs already.

:+1:

wjwwood:
Here's my alternative proposal: https://github.com/ros2/rmw/pull/108 In case we decide to go that way.

clalancette:
> I have the feeling that we should somehow come to a common consensus on that decision and mark it on a central place as we had this discussions on multiple PRs already.

Yeah, agreed.  Once we figure out what we want to do, we should put it in the developers guide.  Note that if we ultimately decide not to force c11, we should probably revisit the flags in rcl_lifecycle, rosidl_generator_c, rosidl_typesupport_introspection_c, and rosidl_typesupport_c .

Karsten1987:
Does the warning still occurs when `c99` is explicitly stated instead of `c11`? As there is no `-std` currently passed, i'd assume the raspberry pi compiler uses whatever its default is.

esteve:
I'm withdrawing this in favor of #109, the code compiles fine, the Raspberry Pi crosscompiler defaulted to C89/C90, instead of C99 and wrongly assumed declaring a variable in a for loop was a C11 feature. Sorry for any confusion!

PR #108, opened 2017-06-05
move variable declaration out of for loop, C99 style

wjwwood:
Possible alternative to #107.

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -3,7 +3,8 @@ cmake_minimum_required(VERSION 3.5)
 project(rmw)
 
 if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99 -Wall -Wextra -Wpedantic")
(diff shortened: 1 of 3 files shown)


Karsten1987:
this seems stange to me. Do we need that with c99, I thought that c99 would lift that constraint.

```
#include <stdio.h>

int main()
{
  for (size_t i = 0; i < 10; ++i) {
    fprintf(stderr, "Hello %zu\n", i);
  }
  return 0;
}
```

compiled with `-std=c99 -Wpedantic -Wall` outputs no error/warning

dhood:
> compiled with `-std=c99 -Wpedantic -Wall` outputs no error/warning

from https://github.com/ros2/rmw/pull/107 it seems it's an issue for particular compilers

wjwwood:
Most compilers allow it regardless of the standard version. But some compilers will fully enforce C99 if specified. However, I'm not sure any of our target systems will be affected. I was just being explicit (and bringing the extra warning flags in as we do in the other packages).

esteve:
My bad, C99 actually allows declaring indexes in the for loop, the Raspberry Pi cross compiler still defaults to C89/C90, which forbids that. See #109 where I explicitly set C99 and C++14

wjwwood:
Thanks @esteve I was misinformed about that too. Closing this in favor of https://github.com/ros2/rmw/pull/109.

PR #109, opened 2017-06-06
Explicitly set C11 and C++14 via the CMake variables

esteve:
This PR explicitly sets the version of the C standard to C11

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -2,8 +2,15 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rmw)
 
-set(CMAKE_C_STANDARD 99)
-set(CMAKE_CXX_STANDARD 14)
+# Default to C99
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 99)
+endif()
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
 
 find_package(ament_cmake_ros REQUIRED)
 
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@esteve just let me know when you want to run CI and I'll run it on all your PRs at once

esteve:
@mikaelarguedas thanks. I think all three PRs (https://github.com/ros2/rcutils/pull/26 https://github.com/ros2/rmw/pull/109 https://github.com/ros2/rcl/pull/141) are ready. I've built them (up to rcl, though) with the Raspberry Pi crosscompiler and they seem fine.

dirk-thomas:
I would suggest holding off changes like this until we make a "final" decision on which standard we want to use. That will hopefully happen in today's meeting.

rmw/CMakeLists.txt
```diff
@@ -2,9 +2,8 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rmw)
 
-if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
-endif()
+set(CMAKE_C_STANDARD 99)
+set(CMAKE_CXX_STANDARD 14)
```
dirk-thomas: These variables should only be set if they are not yet defined in order to not prevent overriding the value.

Same below.

esteve: Fixed in https://github.com/ros2/rmw/pull/109/commits/6e12c761af1916f3b6c27f33d9bd61bb772b5286

esteve:
I just rebased this to resolve any potential conflicts, let me know if there's any further work that needs to be done.

PR #110, opened 2017-06-15
Refactor get topic names and types

wjwwood:
Connects to ros2/ros2#361

### DIFF
--- a/rmw/include/rmw/get_service_names_and_types.h
+++ b/rmw/include/rmw/get_service_names_and_types.h
@@ -1,0 +1,62 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RMW__GET_SERVICE_NAMES_AND_TYPES_H_
+#define RMW__GET_SERVICE_NAMES_AND_TYPES_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include "rmw/macros.h"
+#include "rmw/names_and_types.h"
+#include "rmw/types.h"
+#include "rmw/visibility_control.h"
+
+/// Return a list of service names and their types.
+/**
+ * This function returns a list of service names in the ROS graph and their types.
+ *
+ * The node parameter must not be `NULL`, and must point to a valid node.
+ *
+ * The service_names_and_types parameter must be allocated and zero initialized.
+ * The service_names_and_types is the output for this function, and contains
+ * allocated memory.
+ * Therefore, it should be passed to rmw_names_and_types_fini() when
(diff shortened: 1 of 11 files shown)


rmw/include/rmw/get_service_names_and_types.h
```diff
@@ -0,0 +1,62 @@
...
+ * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RMW_RET_BAD_ALLOC` if memory allocation fails, or
+ * \return `RMW_RET_ERROR` if an unspecified error occurs.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_get_service_names_and_types(
+  const rmw_node_t * node,
+  rcutils_allocator_t * allocator,
```
Karsten1987: nitpick: could that be a `const` pointer?

rmw/include/rmw/names_and_types.h
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * \param[inout] names_and_types object to be finalized
+ * \returns `RMW_RET_OK` on successfully running the check, or
+ * \returns `RMW_RET_INVALID_ARGUMENT` if names_and_types is NULL, or
+ * \returns `RMW_RET_ERROR` when an unspecified error occurs.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_names_and_types_fini(rmw_names_and_types_t * names_and_types);
```
Karsten1987: shouldn't that take an allocator as well?

rmw/src/names_and_types.c
```diff
@@ -0,0 +1,122 @@
...
+    if (rcutils_ret != RCUTILS_RET_OK) {
+      RMW_SET_ERROR_MSG(rcutils_get_error_string_safe())
+      return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+    }
+  }
+  if (names_and_types->types) {
+    // Use the allocator in the names string array
+    // (prevents this data structure from having to also store it)
+    names_and_types->names.allocator.deallocate(
+      names_and_types->types, names_and_types->names.allocator.state);
```
Karsten1987: can there be a mismatch between this allocator and the allocator used for the init function?

rmw/src/names_and_types.c
```diff
@@ -0,0 +1,122 @@
...
+    names_and_types->types = NULL;
+  }
+  // Cleanup names string array
+  rcutils_ret = rcutils_string_array_fini(&names_and_types->names);
```
Karsten1987: doesn't this call take an allocator as well?
https://github.com/ros2/rcutils/blob/master/include/rcutils/types/string_array.h#L86

rmw/src/names_and_types.c
wjwwood: No, I don't think so.

rmw/include/rmw/get_service_names_and_types.h
wjwwood: I'm not 100% sure. On the one hand the allocator's state is mutable (i.e. when someone calls `.allocate` it might change the internal state of the allocator), on the other the `const` in the strict functional sense (see: https://stackoverflow.com/questions/13181546/const-correctness-for-structs-with-pointers) will only restrict changing the function pointers in the allocator.

Here I was using non-const to indicate the allocator is not just "read" from, but instead it is "used" which in my mind is something like "written to".

In short, it could be but the rest of our code which takes an allocator pointer is already non-const to indicate the allocator might be mutated in the function.

If we can find reason is should be const, then I'd update all of them at once rather than just this function.

PR #118, opened 2017-08-10
Make the TOPIC and NAMESPACE macros unsigned.

clalancette:
This is more correct; they can never be negative, as that
doesn't make any sense.  This also fixes some warnings on
mismatched types that get thrown up when adding -Wall -Wextra
to the flags, which this commit also does.

connects to ros2/ros2#387

CI jobs are in the above issue

### DIFF
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -12,6 +12,10 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra)
+endif()
+
 find_package(ament_cmake_ros REQUIRED)
 
 find_package(rcutils REQUIRED)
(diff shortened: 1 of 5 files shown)


dirk-thomas:
How are these changes related to memory leaks?

rmw/CMakeLists.txt
```diff
@@ -12,6 +12,10 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra)
```
mikaelarguedas: would this pass with `-Wpedantic` ? Or it would be too painful to comply with it here ?

clalancette:
> How are these changes related to memory leaks?

They are not directly related.  I saw them while I was debugging the memory leaks, and figured I would open up this PR as well.  Other than the branch name not being exactly correct (which won't matter for the final squash/rebase), it shouldn't be much of a problem.  Unless, of course, you disagree with the change :).

rmw/CMakeLists.txt
clalancette: Interesting question.  Let me try and see what happens.

clalancette: It was pretty easy, actually.  One more small tweak, and I was able to turn it on, so I did.

clalancette:
Linux builds are green with this (and other fixes).  MacOS and Windows are still yellow, but those are all known flaky tests.  This is also approved, so I'm going to merge this.

PR #119, opened 2017-08-10
Fix clang warnings from pedantic.

clalancette:
I made a mistake with my last PR turning on pedantic, and
it turns out that clang is a little more picky.  Fix
that here.

I only ran CI on the OSX build that was throwing warnings, and only up to rmw which now seems to be fixed:

http://ci.ros2.org/job/ci_osx/2416


clalancette:
Thanks, merging.

PR #121, opened 2017-09-27
remove unused domain_id field from rmw_node_t

dirk-thomas:
Remove field which is not set by any of the rmw implementation.

An alternative would be to remove the `const` and update all rmw implementations to set the value.


dirk-thomas:
Waiting for review.

PR #124, opened 2017-10-13
Enable setting log levels in DDS implementation

sriramster:
…l DDS api's

* This patch enables support for three basic log levels
* There needs an API inside the DDS specific rmw implementation i.e
  rmw_fastrtps/rmw_opensplice etc to convert the enum to DDS specific variable

Signed-off-by: Sriram Raghunathan <sriram.max@gmail.com>


dirk-thomas:
The package `rcutils` defines logging levels and a logging API:https://github.com/ros2/rcutils/blob/6d6d5507362415fd377098c0b5973b91072dfafa/include/rcutils/logging.h#L60-L67 These should be used in `rmw` (or at least be referenced by rmw specific symbols).

sriramster:
I'll change the patch accordingly, I did check rcutils for the same. But missed checking for the actual enum.

sriramster:
Hi @dirk-thomas 

I read the code from rcutils and rmw, find two ways of doing it. 

* To define rmw specific log_level enums and convert rcutil specific enum types to rmw specific ones. 
* Or to use rcutil types inside API's which might not need any changes inside rmw.h by doing that we'll have to include rcutils/logging.h inside rmw_implementation (I don't see this as the right approach) we are going against the design of rmw and rcutils already followed. 

Let me know your thoughts on this.

dirk-thomas:
I would suggest to define an rmw specific enum but use the enum values from rcutils to define their values. That way the rmw API uses rmw types but is still compatible with rcutils.

sriramster:
Hi @dirk-thomas 

Have update the pull request based on the discussions above. Could you review the changes?

rmw/include/rmw/types.h
```diff
@@ -229,6 +231,18 @@ typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 
 enum {RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT = 0};
 
+// Creating a map of rcutil log level types to
+// rmw specific types. Check with dds available types
+// before using these.
```
dirk-thomas: Please remove the DDS reference from the comment since the rmw interface doesn't require it to be implemented by DDS.

rmw/include/rmw/types.h
```diff
@@ -229,6 +231,18 @@ typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
...
+enum RMW_PUBLIC_TYPE rmw_log_level_t
```
dirk-thomas: Please use consistent terms: `severity` instead of `level`.

rmw/include/rmw/types.h
```diff
@@ -229,6 +231,18 @@ typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
...
+{
+  RMW_LOG_SEVERITY_DEBUG = RCUTILS_LOG_SEVERITY_DEBUG,
+  RMW_LOG_SEVERITY_INFO  = RCUTILS_LOG_SEVERITY_INFO,
```
dirk-thomas: Please remove the double space to align the equal sign.

Same next line.

sriramster:
@dirk-thomas fixed the comments. Have a look and let me know.

dirk-thomas:
> Please also add the new function declaration to actually pass the log severity.

This comment is still pending.

sriramster:
@dirk-thomas https://github.com/ros2/rmw_implementation/pull/30#issuecomment-338263380 have implemented the same here.

dirk-thomas:
The new function also needs to be declared in this file too: https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h

And then it needs to be implemented for all currently supported rmw implementations: rmw_fastrtps_cpp, rmw_connext_cpp, rmw_opensplice_cpp.

sriramster:
@dirk-thomas the API is included in rmw/rmw.h. I'll update the rmw_dds specific ones in other commits.

dirk-thomas:
I have added the phrase "Connect to <this-PR>" to the initial description of the related PRs so that they are being grouped on the Waffle.io board.

sriramster:
@dirk-thomas I've updated the patch sets on all the linked commits. Could you review? 

Except for one discussion which is still active.

sriramster:
@dirk-thomas @dhood Could you please review this patch? This is connected to another approved patch by Dirk

dirk-thomas:
I triggered a set of builds with these changes:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3821/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/960/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3161/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3908/)

I think the failures are because your branches are forked from a much older version. Can you please rebase all PRs to the latest state of the default branch. Then I can rerun the CI builds.

sriramster:
@dirk-thomas Have update all the connecting PR's to master. Could you please trigger the CI build? Just asking, may I trigger the build in future? (If I'm allowed to do so, could you share an link or document for the same?)

dirk-thomas:
> Could you please trigger the CI build?

* Linux *skipped*
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/969/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3173/)
* Windows *aborted*

The code doesn't compile. Please make sure to compile your changes locally, run the tests and once that passed successfully request a new round of CI build.

> Just asking, may I trigger the build in future?

Currently we can't provide access to ci.ros2.org. In the future build.ros2.org should provide that service to the community.

sriramster:
@dirk-thomas fixed the error, creeped in due to the change from error msg setting to parameterized logging made. Could you check now? I've compiled, tested. And, it is passing locally.

dirk-thomas:
Some linters still fail (they should have also failed for you locally?):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3843/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/974/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3178/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3924/)

sriramster:
@dirk-thomas Done. Missed them again! I need to practice "ros" style more. Sorry again.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3848/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/979/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3183/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3929/)

Please also address the compiler warnings reported: http://ci.ros2.org/job/ci_linux/3848/warnings22Result/new/

sriramster:
@dirk-thomas checked, it is due to unused parameter "severity" with our new API's, would you like a (void) severity; declaration just to nullify the warning. The better solution would be to implement the logging usage with connext and opensplice. Let me know your thoughts on this.

dirk-thomas:
As mentioned before implementing the currently empty functions is the goal. If you would like to merge the current set of PR before that happens please use `(void)severity` to avoid the warning.

sriramster:
@dirk-thomas Thanks. I've made the necessary changes to prevent the warning on the connected patches.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3857/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/983/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3188/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3935/)

I can't say if the remaining test failure are related to the patch or have a different cause. Will need to be checked before merging...

dirk-thomas:
Thank you for the patches.

Please try to follow up with pull requests to add the missing implementation for Connext and OpenSplice when you have a chance.

sriramster:
Thank you. Sure, I'll look into it.

PR #125, opened 2017-10-23
*_raw function

Karsten1987:
Connects to ros2/demos#185


rmw/include/rmw/raw_message.h
```diff
@@ -0,0 +1,51 @@
...
+#include "rmw/types.h"
+#include "rmw/visibility_control.h"
+
+RMW_PUBLIC
+rmw_message_raw_t
+rmw_get_zero_initialized_raw_message(void);
+
+RMW_PUBLIC
+rmw_ret_t
+rmw_initialize_raw_message(
```
sloretz: `rmw_raw_message_init()` would match the existing function `rmw_names_and_types_init()` as well as the initialization functions in rcl `rcl_subscription_init()`, `rcl_node_init()`, etc.

rmw/test/test_raw_message.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "rmw/raw_message.h"
+
+TEST(test_raw_message, default_initialization) {
+  auto raw_msg = rmw_get_zero_initialized_raw_message();
+
+  auto allocator = rcutils_get_default_allocator();
+  EXPECT_EQ(RMW_RET_OK, rmw_initialize_raw_message(&raw_msg, 0, &allocator));
+  EXPECT_EQ(0u, raw_msg.buffer_capacity);
+  EXPECT_TRUE(raw_msg.buffer);
```
sloretz: It doesn't appear to be safe to assume `buffer` is not `NULL` using the default allocator. [The documentation for `malloc()`](http://en.cppreference.com/w/c/memory/malloc) says

```
If size is zero, the behavior is implementation defined (null pointer may be returned, or some non-null pointer may be returned that may not be used to access storage, but has to be passed to free). 
```

How about making `rmw_initialize_raw_message()` not call `allocator->allocate()` if the size is zero, and instead expect the buffer to be `NULL`?

rmw/test/test_raw_message.cpp
```diff
@@ -0,0 +1,62 @@
...
+  EXPECT_EQ(RMW_RET_OK, rmw_raw_message_fini(&raw_msg));
+  EXPECT_EQ(0u, raw_msg.buffer_capacity);
+  EXPECT_FALSE(raw_msg.buffer);
+}
+
+TEST(test_raw_message, resize) {
+  auto raw_msg = rmw_get_zero_initialized_raw_message();
+  auto allocator = rcutils_get_default_allocator();
+  rmw_initialize_raw_message(&raw_msg, 5, &allocator);
```
sloretz: Recommend `ASSERT_EQ(RMW_RET_OK, ...)` to prevent a crash if this call fails.

rmw/include/rmw/types.h
```diff
@@ -221,6 +222,15 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
...
 } rmw_gid_t;
 
+typedef struct RMW_PUBLIC_TYPE rmw_message_raw_t
+{
+  // serialized message data
+  char * buffer;
+  unsigned int buffer_length;
+  unsigned int buffer_capacity;
+  rcutils_allocator_t allocator;
+} rmw_message_raw_t;
```
wjwwood: Did you consider any other names, like `rmw_raw_message_t` (which is a more natural sounding to me, i.e. adjective before noun), `rmw_serialized_message_t` which is more technical sounding, or just `rmw_message_t`?

It's a bit late to change it now, but on the other hand changing it later is going to potentially be painful.

rmw/src/raw_message.c
```diff
@@ -0,0 +1,119 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rcutils/error_handling.h"
+
+#include "rmw/error_handling.h"
+#include "rmw/raw_message.h"
+
+RMW_PUBLIC
```
wjwwood: Visibility macros should not be in the `.c` files.

rmw/include/rmw/raw_message.h
```diff
@@ -0,0 +1,93 @@
...
+ * content is copied over.
+ *
+ * \param msg pointer to the instance of rmw_message_raw_t which is being resized
+ * \param new_size the new size of the internal buffer
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_ERROR` if an unexpected error occurs
+ */
+RMW_PUBLIC
+rmw_ret_t
+rmw_raw_message_resize(rmw_message_raw_t * msg, unsigned int new_size);
```
wjwwood: Same here, why not `size_t`?

rmw/src/raw_message.c
```diff
@@ -0,0 +1,119 @@
...
+    RMW_SET_ERROR_MSG("raw message has no valid allocator");
+    return RMW_RET_ERROR;
+  }
+
+  if (new_size == msg->buffer_capacity) {
+    // nothing to do here
+    return RMW_RET_OK;
+  }
+
+  msg->buffer = allocator->reallocate(msg->buffer, new_size * sizeof(char), allocator->state);
```
wjwwood: This is dangerous, use the `reallocf` style function instead or address the potential memory leak directly, see: https://github.com/ros2/rcutils/blob/3595e7c0a1b815d7ff6aa7e3d2978361ab843bc0/include/rcutils/allocator.h#L128-L136

rmw/include/rmw/raw_message.h
```diff
@@ -0,0 +1,95 @@
...
+
+#include "rmw/types.h"
+#include "rmw/visibility_control.h"
+
+/// Return a zero initialized raw message struct.
+/**
+ * \return rmw_message_raw_t a zero initialized raw message struct
+ */
+RMW_PUBLIC
+rmw_message_raw_t
```
wjwwood: Use `RMW_WARN_UNUSED` on all public functions you added here, see: https://github.com/ros2/rmw/blob/8ea66dbbe89e78318cd2f2b4e7d8da51211d67bb/rmw/include/rmw/macros.h#L22

rmw/include/rmw/raw_message.h
```diff
@@ -0,0 +1,93 @@
...
+ * \param buffer_capacity the size of the memory to allocate for the byte stream
+ * \param allocator the allocator to use for the memory allocation
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_ERROR` if an unexpected error occurs
+ */
+RMW_PUBLIC
+rmw_ret_t
+rmw_raw_message_init(
+  rmw_message_raw_t * msg,
+  unsigned int buffer_capacity,
```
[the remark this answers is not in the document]
Karsten1987: same discussion here: https://github.com/ros2/rclcpp/pull/388#discussion_r194980922

rmw/include/rmw/types.h
Karsten1987: ok. I will address the style comments first, and then rename them.

the type: `rmw_serialized_message_t`

what do you propose for the function?
`rmw_take_serialized_message`?

wjwwood:
@Karsten1987 I added back my changes from last night in c044e2a, looks like they got force pushed over (not easy to recover since it was all in the GitHub webui), please pull before pushing in the future 🙏.

dirk-thomas:
@ros2/rmw_implementations FYI these PRs added new API to publish / take raw (not deserialized) messagesto the rmw interface. For [FastRTPS](https://github.com/ros2/rmw_fastrtps/pull/186) and [Connext](https://github.com/ros2/rmw_connext/pull/259) the new API has been implemented in the linked PRs. For OpenSplice only an issue requesting further information has been filled upstream (ADLINK-IST/opensplice#50). Other RMW implementations should be updated to implement this new API in order to stay compatible.

---

If you are maintaining a RMW impl but are not part of the mentioned team please feel free to reach out to be added to the team. That way you can stay in the loop for future changes.

PR #128, opened 2017-11-16
optimize not to access uninitialized scalar variable

gaoethan:
when the `validation_result` is not initialized and it's better not access
it to avoid unknown issue, like in [_rclpy.c](https://github.com/ros2/rclpy/blob/master/rclpy/src/rclpy/_rclpy.c#L414-L416), moreover, the error msg without 
that value is also  enough to reflect the fact.

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rmw/src/validate_namespace.c
+++ b/rmw/src/validate_namespace.c
@@ -79,7 +79,8 @@ rmw_validate_namespace(
           // explicitly not taking return value which is number of bytes written
           rcutils_snprintf(
             default_err_msg, sizeof(default_err_msg),
-            "rmw_validate_namespace(): unknown rmw_validate_full_topic_name() result");
+            "rmw_validate_namespace(): unknown rmw_validate_full_topic_name() result '%d'",
+            t_validation_result);
           RMW_SET_ERROR_MSG(default_err_msg);
         }
         return RMW_RET_ERROR;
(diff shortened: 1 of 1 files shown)


rmw/src/validate_namespace.c
```diff
@@ -79,9 +79,7 @@ rmw_validate_namespace(
           // explicitly not taking return value which is number of bytes written
           rcutils_snprintf(
             default_err_msg, sizeof(default_err_msg),
-            "rmw_validate_namespace(): unknown rmw_validate_full_topic_name() result '%d'",
-            *validation_result
-          );
+            "rmw_validate_namespace(): unknown rmw_validate_full_topic_name() result");
```
dirk-thomas: The idea is that the actual result number is being printed as part of the error message. `*validation_result` is certain the wrong variable. This should print `t_validation_result` instead.

gaoethan: yes, it also makes sense. thanks.

dirk-thomas:
All jobs failed likely due to the merged PR ros2/rclcpp#371 from @tfoote.

wjwwood:
@dirk-thomas actually I think I screwed up the gist, I change `rcl` not `rmw`. I don't think he merged anything into `rmw`.

PR #130, opened 2017-11-22
Optimize namespace node and topic validation

gaoethan:
Make the result reflection more accurate, which should be
different from the default `NULL`
Use `rmw_namespace_validation_result_string` instead
for namespace validation

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rmw/test/test_validate_full_topic_name.cpp
+++ b/rmw/test/test_validate_full_topic_name.cpp
@@ -47,7 +47,8 @@ TEST(test_validate_topic_name, valid_topic) {
   ASSERT_EQ(RMW_RET_OK, ret);
   ASSERT_EQ(RMW_TOPIC_VALID, validation_result);
 
-  ASSERT_EQ((char *)NULL, rmw_full_topic_name_validation_result_string(validation_result));
+  ret = strcmp("valid topic name", rmw_full_topic_name_validation_result_string(validation_result));
+  ASSERT_EQ(0, ret);
 }
 
 TEST(test_validate_topic_name, empty_topic_name) {
--- a/rmw/test/test_validate_full_topic_name.cpp
+++ b/rmw/test/test_validate_full_topic_name.cpp
@@ -67,7 +67,10 @@ TEST(test_validate_topic_name, empty_topic_name) {
   ASSERT_EQ(RMW_TOPIC_INVALID_IS_EMPTY_STRING, validation_result);
   ASSERT_EQ(0ul, invalid_index);
 
-  ASSERT_NE((char *)NULL, rmw_full_topic_name_validation_result_string(validation_result));
+  ret =
+    strcmp("topic name must not be empty", rmw_full_topic_name_validation_result_string(
+        validation_result));
+  ASSERT_EQ(0, ret);
 }
 
 TEST(test_validate_topic_name, not_absolute) {
@@ -91,7 +94,9 @@ TEST(test_validate_topic_name, not_absolute) {
   ASSERT_EQ(RMW_TOPIC_INVALID_NOT_ABSOLUTE, validation_result);
   ASSERT_EQ(0ul, invalid_index);
 
-  ASSERT_NE((char *)NULL, rmw_full_topic_name_validation_result_string(validation_result));
+  ret = strcmp("topic name must be absolute, it must lead with a '/'",
+      rmw_full_topic_name_validation_result_string(validation_result));
+  ASSERT_EQ(0, ret);
 }
 
 TEST(test_validate_topic_name, ends_with_forward_slash) {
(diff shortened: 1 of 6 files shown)


gaoethan:
#https://github.com/ros2/rclcpp/pull/405

mikaelarguedas:
It looks like the string returned was changed from "X is valid" to `NULL` in https://github.com/ros2/rmw/pull/93/commits/c8d25e6b0a825ca71167267d6526432012027dc6.
So I'll leave it to @wjwwood to give feedback on this as I don't remember what motivated the change.

My guess is that this "matches" a return code of 0 when it succeeds, allowing you to easily know that it went well without having to do costly string operations.

If this was the motivation I would recommend returning "unknown X validation result" in the default case rather than changing the return value when the topic/node/namespace name is valid.


Good catch for the use of the wrong function in the namspace tests :+1:

clalancette:
Doing a quick, dirty audit of the callers, it seems like almost none of them expect NULL to be returned from the `validation_error_string` APIs (most of them do something like `RCUTILS_SET_ERROR_MSG`, which will crash with a NULL error_string).  Therefore, I think we should probably change it so that the `validation_error_string` APIs *never* return NULL (even in the default case), and audit all of the callers to make sure the correct thing will happen.

gaoethan:
yes, I opt to unify all the specific error string in `validation_error_string` to avoid extra code to deal with NULL out.

gaoethan:
@wjwwood @mikaelarguedas The NULL return is removed and unify the result sting therein, and the relevant adaption required in the components which are affected by it are referenced, please review and give your comments, thanks !

rmw/test/test_validate_full_topic_name.cpp
```diff
@@ -47,7 +47,8 @@ TEST(test_validate_topic_name, valid_topic) {
   ASSERT_EQ(RMW_RET_OK, ret);
   ASSERT_EQ(RMW_TOPIC_VALID, validation_result);
 
-  ASSERT_EQ((char *)NULL, rmw_full_topic_name_validation_result_string(validation_result));
+  ret = strcmp("valid topic name", rmw_full_topic_name_validation_result_string(validation_result));
```
mikaelarguedas: ret is of type `rmw_ret_t` that may or may not be an int in the future. I'd recommend using
```
ASSERT_STREQ("valid topic name", rmw_full_topic_name_validation_result_string(validation_result));
``` 
instead. (same in all tests using strcmp).

wjwwood:
> I still think that having to do string comparison instead of checking for NULL is pretty expensive and would prefer to have NULL returned for the valid cases,

Agreed. -1 to string comparison. You could instead have it return a ret code and pass a cstring pointer in to be filled with the result. But it should be possible without string comparison to know if a valid string was passed.

gaoethan:
actually, I also think it's a little heavy to compare with the specific string, so I agree with you to return NULL for valid case.  however: 

- you know, when there is NULL returned, the **valid case** (e.g `RMW_NODE_NAME_VALID`) should be checked firstly before stepping next to `rmw_*_validation_result_string` to make sure no `NULL `will returned, or it has potential to deference to `NULL `like in `RCUTILS_SET_ERROR_MSG` and [rclcpp PR 405](https://github.com/ros2/rclcpp/pull/405)

- If merely checking not NULL in test to avoid specific string comparing, it's not sure for really at a certain illegal case, even though it's not a big problem here for the tests, you know, non-NULL has several illegal cases.

gaoethan:
@ros2/team, any new comments ? I can tweak it if there is to avoid its pending :smile: thanks !

mikaelarguedas:
sorry for the delay @gaoethan I'll review this and the corresponding PRs in the next 24h

mikaelarguedas:
I just reviewed these. One last comments are addressed we'll need to rerun CI on this set of PRs before merging

mikaelarguedas:
@gaoethan can you please rebase the branches of all these PRs? as we changed some things in the rmw API this cannot be built / tested on CI without being rebased

gaoethan:
@mikaelarguedas All have already been rebased, and you can go ahead now, :) thanks !

mikaelarguedas:
thanks @gaoethan for iterating, here's a round of ci on this set of PRs

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3813/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/951/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3152/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3900/)

gaoethan:
@mikaelarguedas it seems that CI has passed, should it be merged now ? thanks.

mikaelarguedas:
Sorry @gaoethan for the delay, I meant to tag @wjwwood on this set of PRs to see if the change makes sense to him as he is the original author. This looks good to me and CI is still passing

I reran CI just to make sure nothing changes since.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3844/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/975/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3179/) (unrelated)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3925/) (unrelated)

mikaelarguedas:
Thanks @wjwwood I'll merge the connected PRs as well then

wjwwood:
Oh, sorry I missed that there were connected pr's. Thanks.

PR #132, opened 2017-12-07
add function to parse key-value user_data

dirk-thomas:
Move parse key-value function from `rmw_opensplice_cpp` here for reuse.

Connect to ros2/ros2#438.


wjwwood:
Changing label to in progress to match connected issue's status.

dirk-thomas:
I changed it back since this PR is "in review" (and actually already reviewed).

PR #138, opened 2018-05-22
check the return value of rcutils_snprintf

wjwwood:
Connects to ros2/rcutils#85

### DIFF
--- a/rmw/include/rmw/impl/cpp/macros.hpp
+++ b/rmw/include/rmw/impl/cpp/macros.hpp
@@ -95,7 +95,7 @@
 #define RMW_CHECK_TYPE_IDENTIFIERS_MATCH(ElementName, ElementTypeID, ExpectedTypeID, OnFailure) \
   { \
     if (ElementTypeID != ExpectedTypeID) { \
-      size_t __bytes_that_would_have_been_written = rcutils_snprintf( \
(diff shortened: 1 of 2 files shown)


rmw/include/rmw/impl/cpp/macros.hpp
```diff
@@ -90,21 +95,37 @@
...
+        } else { \
+          int ret = rcutils_snprintf( \
+            __msg, __bytes_that_would_have_been_written + 1, \
+            #ElementName " implementation '%s'(%p) does not match rmw implementation '%s'(%p)", \
+            ElementTypeID, reinterpret_cast<const void *>(ElementTypeID), \
+            ExpectedTypeID, reinterpret_cast<const void *>(ExpectedTypeID)); \
+          if (ret < 0) { \
+            RMW_SET_ERROR_MSG( \
+              "RMW_CHECK_TYPE_IDENTIFIERS_MATCH(): rcutils_snprintf() failed"); \
+          } else if (NULL != __msg) { \
```
dhood: Does the null check still apply (it is not used [below](https://github.com/ros2/rmw/pull/138/files#diff-2d08bb6c3f837ed698aab8d8b3e2734cR97))? If so, I'd OR it with the `ret < 0` check above (if `__msg` is null at this point it's an issue that should be raised not skipped)

PR #140, opened 2018-06-15
use rcutils allocator in allocators.c to avoid direct use of malloc/free

wjwwood:
This pull request makes sure that we're using `rcutils_allocator_t` everywhere and not using `malloc`, `realloc`, `calloc`, or `free` anywhere at all.

Connects to ros2/rcutils#102

### DIFF
--- a/rmw/src/allocators.c
+++ b/rmw/src/allocators.c
@@ -17,13 +17,16 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <rcutils/allocator.h>
+
 #include "rmw/types.h"
 
 void *
 rmw_allocate(size_t size)
 {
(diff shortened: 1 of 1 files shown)


wjwwood:
I was a little worried about a performance impact with this change (as the rcutils allocator is across a shared library boundary most of the time), but I was looking at the decompiled code and apparently it jumps directly to `__default_allocate` in `allocators.c` of `rcutils` from `rmw_allocate()`. I think the compiler is smart enough to know that what `rcutils_get_default_allocator()` returns is stored locally as `static` and is unmodified, so it makes it `const`, and because `rmw_allocate` also doesn't modify the copy of it, the compiler just removes it all together and goes where `.allocate` points with what `.state` points to.

PR #143, opened 2018-07-16
get encoding identifier

Karsten1987:
With respect to the upcoming development for rosbags, we need to extract the encoding format for each RMW middleware. 

The encoding identifier differs from the implementation identifier in the sense that each implementation has to be uniquely identified, but multiple RMW can use the same encoding (e.g. cdr for DDS implementations).

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -104,9 +104,10 @@ rmw_get_implementation_identifier(void);
 /**
  * Return the format in which binary data is serialized.
  * One middleware can only have one encoding.
- * In contrast to the implementation identifier, the serialization format can be
- * equal between multiple RMW implementations. This means, that the same binary
- * messages can be deserialized by RMW implementations with the same format.
+ * In contrast to the implementation identifier, the serialization format can be equal between
+ * multiple RMW implementations.
+ * This means, that the same binary messages can be deserialized by RMW implementations with the
+ * same format.
  * \sa rmw_serialize
  * \sa rmw_deserialize
  * \return serialization format
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -100,6 +100,22 @@ RMW_WARN_UNUSED
 const char *
 rmw_get_implementation_identifier(void);
 
+/// Get the unique encoding identifier for this middleware.
+/**
+ * The unique encoding identifier states in which encoding the
+ * serialized data has to be interpreted.
+ * One middleware can only have one encoding.
+ * In contrast to the implementation identifier, the encoding identifier can be
+ * equal between multiple RMW implementations. This means, that the same binary
+ * messages can be deserialized by RMW implementations with the same encoding ID.
+ * See also rmw_serialize, rmw_deserialize
+ * \return encoding identifier
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+const char *
+rmw_get_encoding_identifier(void);
+
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_ret_t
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -100,6 +100,22 @@ RMW_WARN_UNUSED
...
 
+/// Get the unique encoding identifier for this middleware.
+/**
+ * The unique encoding identifier states in which encoding the
+ * serialized data has to be interpreted.
+ * One middleware can only have one encoding.
+ * In contrast to the implementation identifier, the encoding identifier can be
+ * equal between multiple RMW implementations. This means, that the same binary
+ * messages can be deserialized by RMW implementations with the same encoding ID.
+ * See also rmw_serialize, rmw_deserialize
```
wjwwood: Could use the doxygen `\sa` to reference them.

rmw/include/rmw/rmw.h
```diff
@@ -100,6 +100,22 @@ RMW_WARN_UNUSED
...
+ * \return encoding identifier
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+const char *
```
wjwwood: Is it ever going to be a performance problem to compare strings here? Unlike the implementation identifier we cannot just compare pointer values because the same encoding string might be defined in multiple places. Not sure what else we would do, maybe some kind of hash or a "registry" of encodings? Just a thought, not a blocker.

dirk-thomas:
I am not sure if "encoding" is the right term here. Isn't "serialization" more appropriate? E.g. `CDR` is a serialization format.

mikaelarguedas:
Yeah I had the same concern. Especially as we might use the term "encoding" when talking about e.g. strings

wjwwood:
I'm fine with `serialization_format` instead. Some people (at least) seem to draw some difference between the two:

- https://www.quora.com/What-is-the-difference-between-serialization-and-encoding
- https://stackoverflow.com/questions/22574039/serialization-versus-encoding

Karsten1987:
The CDR spec also uses the expression of a "CDR encoding", but I really don't care.
So what's the correct function name?

`rmw_get_serialization_format_identifier` ?

wjwwood:
I'd be fine with `rmw_get_serialization_format`.

Karsten1987:
CI (`--packages-up-to rmw rmw_implementation rmw_connext_cpp rmw_fastrtps_cpp rmw_opensplice_cpp`)

win: [Build Status](https://ci.ros2.org/job/ci_windows/4942/)
osx: [Build Status](https://ci.ros2.org/job/ci_osx/4086/)
aarch: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1748/)
linux: [Build Status](https://ci.ros2.org/job/ci_linux/4921/)

Karsten1987:
I changed encoding to `serialization_format` as requested. Putting this in review.

rmw/include/rmw/rmw.h
```diff
@@ -100,6 +100,22 @@ RMW_WARN_UNUSED
 const char *
 rmw_get_implementation_identifier(void);
 
+/// Get the unique serialization format for this middleware.
+/**
+ * Return the format in which binary data is serialized.
+ * One middleware can only have one encoding.
+ * In contrast to the implementation identifier, the serialization format can be
+ * equal between multiple RMW implementations. This means, that the same binary
```
dirk-thomas: New sentence, new line.

Karsten1987:
CI:

osx: [Build Status](https://ci.ros2.org/job/ci_osx/4125/)
linux: [Build Status](https://ci.ros2.org/job/ci_linux/4965/) (unrelated)
win: [Build Status](https://ci.ros2.org/job/ci_windows/4981/)

dirk-thomas:
@ros2/rmw_implementations FYI these PRs added new API to the rmw interface. For FastRTPS, Connext and OpenSplice referenced PRs implement the new API (returning "Cdr" as the format). Other RMW implementations should be updated to implement this new API in order to stay compatible.

---

If you are maintaining a RMW impl but are not part of the mentioned team please feel free to reach out to be added to the team. That way you can stay in the loop for future changes.

Issue #144, opened 2018-07-31
move rmw_serialized_message_t to rcutils


Karsten1987:
For reusability purposes, the `rmw_serialized_message_t` type should be excluded from this `rmw` package. The data type shall be placed in `rcutils` as it is basically _only_ a vector of characters.

Karsten1987:
CI:

linux: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_linux&build=4984)](https://ci.ros2.org/job/ci_linux/4984/)
aarch: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=1797)](https://ci.ros2.org/job/ci_linux-aarch64/1797/)
osx: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_osx&build=4140)](https://ci.ros2.org/job/ci_osx/4140/)
win: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_windows&build=4996)](https://ci.ros2.org/job/ci_windows/4996/)

Karsten1987:
new round of CI:

linux: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_linux&build=4985)](https://ci.ros2.org/job/ci_linux/4985/)
aarch: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=1798)](https://ci.ros2.org/job/ci_linux-aarch64/1798/)
osx: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_osx&build=4141)](https://ci.ros2.org/job/ci_osx/4141/)
win: [![Build Status](https://ci.ros2.org/buildStatus/icon?job=ci_windows&build=4997)](https://ci.ros2.org/job/ci_windows/4997/)

PR #145, opened 2018-07-31
rmw serialized to rcutils serialized

Karsten1987:
fixes https://github.com/ros2/rmw/issues/144

I let the original test for the serialized message in this package to ensure the typedefs working.
I am happy to remove them if considered as unnecessary.


rmw/test/test_serialized_message.cpp
```diff
@@ -16,7 +16,7 @@
 
 #include "rcutils/allocator.h"
 
-#include "rmw/serialized_message.h"
+#include "rmw/types.h"
```
mikaelarguedas: Nit: Is this change necessary ?

Karsten1987: I need that for the return values.

PR #152, opened 2018-10-17
RMW configuration from XML only

JuanCarlos-eProsima:
We have implemented a new feature to apply the RMW configuration only from XML files, avoiding the current configuration set by code. To do that we have added a new boolean value to the `rmw_qos_profile_t` struct.

Connects to ros2/ros2#589


JuanCarlos-eProsima:
After checking this solution with @dirk-thomas and @mjcarroll we have decided to change the approach and with the new management, we don't need any change in this package.

PR #163, opened 2019-01-10
pass context to wait set, and provide fini function for context

wjwwood:
This is a follow up of https://github.com/ros2/rmw/pull/154.

This is not safe for inclusion in Crystal patch releases because it breaks API and adds to API, both of which would require all rmw implementations to be updated.

### DIFF
--- a/rmw/include/rmw/init.h
+++ b/rmw/include/rmw/init.h
@@ -108,6 +108,33 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_shutdown(rmw_context_t * context);
 
+/// Finalize a context.
+/**
+ * The context to be finalized must have been previously initialized with
+ * `rmw_init()`, and then later invalidated with `rmw_shutdown()`.
+ * If context is `NULL`, then `RMW_RET_INVALID_ARGUMENT` is returned.
+ * If context is zero-initialized, then `RMW_RET_INVALID_ARGUMENT` is returned.
+ * If context is initialized and valid (`rmw_shutdown()` was not called on it),
+ * then `RMW_RET_INVALID_ARGUMENT` is returned.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please create a stub page for Dashing in [ros2_documentation](https://github.com/ros2/ros2_documentation) and mention this RMW interface change there so that we keep track of such changes from the beginning.

Also @ros2/rmw_implementations FYI.

wjwwood:
See https://github.com/ros2/ros2_documentation/pull/94 for documentation.

rmw/include/rmw/init.h
```diff
@@ -108,6 +108,33 @@ RMW_WARN_UNUSED
...
+ * <i>[1] if `atomic_is_lock_free()` returns true for `atomic_uint_least64_t`</i>
+ *
+ * \return `RMW_RET_OK` if the shutdown was completed successfully, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RMW_RET_ERROR` if an unspecified error occur.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_context_fini(rmw_context_t * context);
```
sloretz: Separating shutdown seems reasonable; is there a technical reason that made it mandatory?

wjwwood: It's mostly for symmetry with rcl, but also it allowed the rmw micro implementation to defer cleanup of resources into fini of the context rather than in shutdown of the context which happens sometimes far earlier than fini.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6079/)
* Linux OpenSplice+FastRTPS [Build Status](https://ci.ros2.org/job/ci_linux/6080/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2557/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5006/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5934/)

PR #166, opened 2019-02-08
Backport #165 for Crystal.

nuclearsandwich:
Without backporting this change, Crystal's upstream development branches don't build since ament_cmake is still tracking master. If this change isn't compatible then we'll also need to branch ament_cmake.


dirk-thomas:
I am not sure this is a good rational for doing backports. Nothing requires this change to be backported until the related `ament_cmake` change gets backported.

nuclearsandwich:
> related `ament_cmake` change gets backported.

ament_cmake in Crystal is still using `master` for the development branch. While that change hasn't been released it's on crystal's development branch. The alternative to backporting this change is creating a crystal branch of ament_cmake which drops the fix from https://github.com/ament/ament_cmake/pull/155

dirk-thomas:
Also the target branch being `master` doesn't make much sense here...

nuclearsandwich:
Given that our backport strategy is now explicitly only when necessary. I'm closing this backport as abandoned.

PR #168, opened 2019-03-15
Rmw events

dabonnie:
Initial commit with the RMW QoS structs defined. I have a few questions that require changes in the code:

- how should we propagate the non-pimitive types? void * could work, but this puts the burden of discovery upon the user based on RMW implementation
- are the struct definitions suffcient? should they contain a field with a common struct containing the rmw_event_type_t?
- which rmw_ret_t corresponds to each rmw_event_type_t?


dabonnie:
Closing and deleted as created a PR against the wrong repo. My mistake.

PR #169, opened 2019-03-25
Add function to get publisher actual qos settings

ivanpauno:
Added rmw_publisher_get_actual_qos function.
This function was proposed in the discussion here: https://github.com/ros2/rclcpp/pull/664#discussion_r268274073.

I will complete the implementation of the api, following these steps:

- Open a [PR in rmw_implementation](https://github.com/ros2/rmw_implementation/pull/56) to expose the api
- Open a [PR in rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/pull/267) to implement the api. It will call the appropriate fast_rtps method and translate from the FastRTPS QoS structure to the rmw QoS structure.
- Open a [PR in rcl](https://github.com/ros2/rcl/pull/406) proposing an equivalent api and implement it.
- Open a [PR in rclcpp](https://github.com/ros2/rclcpp/pull/667) proposing an equivalent api and implement it.
- Use the rclcpp api in https://github.com/ros2/rclcpp/pull/664 to solve the TODO.
- Open PRs in [rmw_connext](https://github.com/ros2/rmw_connext/pull/350) and [rmw_opensplice](https://github.com/ros2/rmw_opensplice/pull/265) to implement the api for them.


rmw/include/rmw/rmw.h
```diff
@@ -251,6 +251,24 @@ rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count);
 
+/// Retrieve the actual qos settings of the publisher.
+/**
+ * Query the underlying middleware to determine the qos settings
+ * of the publisher.
+ *
+ * \param[in] publisher the publisher object to inspect
+ * \param[out] qos the actual qos settings
```
mjcarroll: Do you allocate a new `rmw_qos_profile_t` in this case?  If so, we should probably document it here.  If you are not, it may make more sense to have a pointer to `const` so that users cannot mutate the internal state.

ivanpauno: My idea is that it should be previously allocated, and the function modify its state to get the actual qos settings. So, the signature can't be modified like a pointer to const (it's an output).

The other alternative is to allocate memory in the function.

wjwwood: We use this pattern in other places. I think it ought to be ok. You could consider making the name more descriptive, like `actual_qos_out`, but that's just if you guys think the docs are insufficient.

rmw/include/rmw/rmw.h
```diff
@@ -251,6 +251,24 @@ rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count);
 
+/// Retrieve the actual qos settings of the publisher.
+/**
+ * Query the underlying middleware to determine the qos settings
+ * of the publisher.
```
wjwwood: I think it is worth pointing out why this exists, i.e. to resolve `_SYSTEM_DEFAULT` settings.

ivanpauno:
@mjcarroll @wjwwood All comments have already been solved here. Could you check it again?
Also, RMW_QOS_POLICY_*_UNKNOWN option was added to some enums, as requested here https://github.com/ros2/rmw_fastrtps/pull/267#issuecomment-476310352.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6556/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2895/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5373/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6354/)

ivanpauno:
CI again:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6570/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2908/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5386/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6371/)

PR #170, opened 2019-03-26
Update names and types struct

jacobperron:
This change enables storing a fully qualified type name as a tuple, rather than a single string.
E.g. ["pkg_name", "msg", "TypeName"] vs. "pkg_name/msg/TypeName".
This should help avoid duplicate string splitting logic in client libraries.

Each element of the outer array represents a list (`rcutils_array_list_t`) of types for the associated name.
Each element of the array list is a string array (`rcutils_string_array_t`) that is the fully qualified type name tuple.

Connects to ros2/ros2#677


jacobperron:
Decided to return a delimited string rather than returning a nested list to keep the API simpler.

PR #182, opened 2019-08-29
Add specific return type for non existent node

ivanpauno:
Addresses https://github.com/ros2/ros2cli/pull/322#issuecomment-525364023

### DIFF
--- a/rmw/include/rmw/ret_types.h
+++ b/rmw/include/rmw/ret_types.h
@@ -38,7 +38,8 @@ typedef int32_t rmw_ret_t;
 
 // rmw node specific ret codes in 2XX
 /// Failed to find node name
-#define RMW_RET_NON_EXISTENT_NODE_NAME 203
+// Using same return code than in rcl
+#define RMW_RET_NODE_NAME_NON_EXISTENT 203
 
 #ifdef __cplusplus
 }
--- a/rmw/include/rmw/ret_types.h
+++ b/rmw/include/rmw/ret_types.h
@@ -36,6 +36,10 @@ typedef int32_t rmw_ret_t;
 /// Incorrect rmw implementation.
 #define RMW_RET_INCORRECT_RMW_IMPLEMENTATION 12
 
+// rmw node specific ret codes in 2XX
+/// Failed to find node name
+#define RMW_RET_NON_EXISTENT_NODE_NAME 203
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


rmw/include/rmw/ret_types.h
```diff
@@ -36,6 +36,10 @@ typedef int32_t rmw_ret_t;
 /// Incorrect rmw implementation.
 #define RMW_RET_INCORRECT_RMW_IMPLEMENTATION 12
 
+// rmw node specific ret codes in 2XX
+/// Failed to find node name
+#define RMW_RET_NON_EXISTENT_NODE_NAME 203
```
dirk-thomas: Why 203 instead of maybe 20?

dirk-thomas: I found the matching define in `rcl` so this number makes sense. Maybe mention `rcl` in the existing comment.

jacobperron: Does it matter if they're matching? There's a [convert function](https://github.com/ros2/rcl/blob/bad86c204db17f2538a537366edf74b4d79645bd/rcl/src/rcl/common.c#L68-L69) anyway.

Seems arbitrary to me.

jacobperron: Also, I think the name needs to be updated according to https://github.com/ros2/rcl/pull/492#discussion_r319266399

```diff
- RMW_RET_NON_EXISTENT_NODE_NAME
+ RMW_RET_NODE_NAME_NON_EXISTENT
```

ivanpauno: > Does it matter if they're matching? There's a convert function anyway.
>
> Seems arbitrary to me.

It's arbitrary yes, but I think it have some sense.

> Also, I think the name needs to be updated according to ros2/rcl#492 (comment)
> ```
> - RMW_RET_NON_EXISTENT_NODE_NAME
> + RMW_RET_NODE_NAME_NON_EXISTENT
> ```

Thanks, forgot to push that commit.

rmw/include/rmw/ret_types.h
```diff
@@ -36,6 +36,11 @@ typedef int32_t rmw_ret_t;
 /// Incorrect rmw implementation.
 #define RMW_RET_INCORRECT_RMW_IMPLEMENTATION 12
 
+// rmw node specific ret codes in 2XX
+/// Failed to find node name
+// Using same return code than in rcl
```
hidmic: @ivanpauno why?

ivanpauno: I prefer using the same code when possible, I don't think it's important

PR #187, opened 2019-10-04
add pub/sub option structures to support rmw specific payload feature

wjwwood:
Connects to ros2/rclcpp#882


rmw/include/rmw/types.h
```diff
@@ -45,18 +45,81 @@ typedef struct RMW_PUBLIC_TYPE rmw_node_t
...
+   */
+  void * rmw_specific_subscription_payload;
+
+  /// If true then the middleware should not deliver data from local publishers.
+  /**
+   * This setting is most often used when data should only be received from
+   * remote nodes, especially to avoid "double delivery" when both intra- and
+   * inter- process communication is taking place.
+   *
+   * \TODO(wjwwood): nail this down when participant mapping is sorted out.
```
mjcarroll: It may be worth putting a ref to the design document here

wjwwood: This one? https://github.com/ros2/design/pull/250

mjcarroll: Yeah, that's the one I was thinking.

eboasson:
@wjwwood it would be really nice if there were (e.g.) a version macro for the RMW interface so that a single source can be used to cover some of these changes. I suppose ``RMW__PUBLISHER_OPTIONS_H_`` and ``RMW__SUBSCRIPTION_OPTIONS_H_`` could be used as a proxy, but it feels a bit dirty.

wjwwood:
@eboasson that's a good idea, can you open an issue to track that? This isn't the first time we've needed to change API in rmw and it won't be the last, so I think having tools for you to have a "single" source is out of scope for this pr at least.

rmw/include/rmw/types.h
wjwwood: 09f64c7

PR #190, opened 2019-10-11
Add function to enable localhost communication only from env var

BMarchi:
PR to implement [network traffic restriction for localhost](https://github.com/ros2/ros2/issues/798). There are no test yet since this is a mock open for implementation discussion. [FastRTPS implementation](https://github.com/ros2/rmw_fastrtps/pull/331)

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8441/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4343/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8352/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/6866/)


rmw/include/rmw/host.h
```diff
@@ -0,0 +1,52 @@
...
+ * \returns `RMW_ENV_VAR_NOT_DEFINED_OR_EMPTY` if the env var is not defined or is empty.
+ * \returns `RMW_LOCAL_HOST_ENABLED` if the environment variable is localhost
+ * \returns `RMW_INVALID_ALLOWED_HOSTS` when the user provides a non empty string
+ * different than localhost.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_allowed_hosts(
+	char * allowed_hosts);
```
dirk-thomas: Nitpick: indentation is off.

Same below.

rmw/src/host.c
```diff
@@ -0,0 +1,50 @@
...
+
+#include <stdlib.h>
+#include <string.h>
+
+rmw_ret_t
+rmw_allowed_hosts(
+	char * allowed_hosts)
+{
+  rmw_ret_t ret = RMW_ENV_VAR_NOT_DEFINED_OR_EMPTY;
+	const char * ros_host_env = "ROS_ALLOWED_HOSTS";
```
dirk-thomas: This constant should be a global constant to be publically accessible.

rmw/src/host.c
```diff
@@ -0,0 +1,50 @@
...
+#ifndef _WIN32
+  const char * ros_host_env_val = getenv(ros_host_env);
+  if (ros_host_env_val != NULL && strlen(ros_host_env_val) > 0) {
+    allowed_hosts = malloc(strlen(ros_host_env_val)+1);
+    strcpy(allowed_hosts, ros_host_env_val);
+  }
+#else
+  size_t ros_host_env_val_size;
+  _dupenv_s(&allowed_hosts, &ros_host_env_val_size, ros_host_env);
+#endif
```
dirk-thomas: Use the function provided by [rcutils](https://github.com/ros2/rcutils/blob/3daf12617c2bfe7d6eb49c68e2710d7727ef2e6d/include/rcutils/get_env.h#L62) instead.

rmw/include/rmw/host.h
dirk-thomas: This API is fairly limited imo. It leaves the parsing of the actual string up to the caller instead of encapsulating this in the function and e.g. populating a list of hostnames. The special value for local host is changing the return value though. This feel pretty inconsistent.

BMarchi:
I wanted a third return option, that's why I didn't use the bool one, so proper behavior can be done if the variable has a wrong value. I guess I can remove the possibility to return the host list without parsing and just return the code. If we want `list parsing` for the future then we add another function as you said. Does it sound good for you?

dirk-thomas:
> I guess I can remove the possibility to return the host list without parsing and just return the code.

Sounds good to me.

> `#define RMW_ENV_VAR_NOT_DEFINED_OR_EMPTY 0`
> `#define RMW_LOCAL_HOST_ENABLED 1`
> `#define RMW_INVALID_ALLOWED_HOSTS 2`

I am not sure how to name / value the three return codes for something like `use_localhost_only` though.

Instead of `LOCAL_HOST_ENABLED` I would propose `LOCALHOST_ONLY` since it not only states that localhost is enabled but that it is the only enabled target.

I would be worried that callers interpret the numeric return value as a boolean. So if `LOCALHOST_ONLY` is non-zero I would wish the two other return codes would both be zero (which is obviously not an option).

Maybe a signature like `rc use_localhost_only(bool *)` would be more natural to use? The `rc` is `OK / 0` or `non-zero / UNSUPPORTED_VALUE`. And the bool gets set either way - in the case of an unsupported value or the env var not set at all the default value of our choice.

[All this hops are only "necessary" because we try to use a env var with a wider value space than we are comfortable to actually specify atm. I still think just using an env var `ROS_ENABLE_NETWORK` is the much cleaner choice for the time being - with a simple boolean getter function - no extras.]

BMarchi:
After reading the discussion in the issue, it seems that people by majority prefers the boolean rather than a list of hosts. So I'll make the changes accordingly!

rmw/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(rmw_sources
   "src/allocators.c"
   "src/convert_rcutils_ret_to_rmw_ret.c"
   "src/event.c"
+  "src/host.c"
```
dirk-thomas: How about renaming this file to `localhost`?

rmw/include/rmw/host.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rmw/macros.h"
+#include "rmw/types.h"
+
+const char * RMW_LOCAL_HOST_ENV_VAR = "ROS_LOCALHOST_ONLY";
```
dirk-thomas: Since `localhost` is one word no underscore between `local` and `host`.

Same for the function below.

rmw/include/rmw/host.h
```diff
@@ -0,0 +1,43 @@
...
+
+/// Determine if the user wants to communicate using loopback only.
+/**
+ * Checks if localhost should be used for network communication checking ROS_LOCALHOST_ONLY env
+ * variable
+ * \returns true if ROS_LOCALHOST_ONLY is 1 or true, false otherwise.
```
dirk-thomas: None of the other env vars support "true". Therefore I would suggest to only support `1`, `0` and "not set" / "empty".

rmw/src/host.c
```diff
@@ -0,0 +1,29 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+bool
+rmw_local_host_only()
+{
+	const char * ros_local_host_env_val = NULL;
+  return rcutils_get_env(RMW_LOCAL_HOST_ENV_VAR, &ros_local_host_env_val) == NULL &&
+    ros_local_host_env_val != NULL &&
+    (strcmp(ros_local_host_env_val, "true") == 0 || strcmp(ros_local_host_env_val, "1") == 0);
```
dirk-thomas: Regarding the default value lets keep the default to `false` when the env var isn't set or empty. Mostly to not hold up the merge by changing current behavior. We can swap the default by setting the env var in the `ros_environment` package later if decided.

dirk-thomas:
With the RMW API exposing the flag as an argument of the create node function (same as the domain ID) I would argue this function should actually be implemented in `rcl` instead (same as where the domain id env var is being checked).

PR #186, opened 2019-09-27
Adding required structs and methods to get a list  of publishers or subscribers with their respective qos

jaisontj:
Includes creating relevant data structures to enable the feature mentioned [here](https://github.com/ros2/rmw/issues/151). Relevant information about implementation is discussed [here](https://github.com/ros2/rmw_fastrtps/issues/319)

Summary:
- Added two new structs
  - `rmw_participant_qos_profile_t` to store participant name and their qos policy in one struct
  - `rmw_participants_t` to store a list of `rmw_participant_qos_profile_t`
- Added two new functions:
  - `rmw_participant_qos_profile_allocate`
  - `rmw_participant_qos_profile_free`
- Tests for the above two functions

Note: `colcon build --packages-up-to rmw && colcon test --packages-up-to rmw && colcon test-result --verbose --test-result-base build/rcl` passes without any errors or failures.

PR #193, opened 2019-11-15
Support for ON_REQUESTED_INCOMPATIBLE_QOS and ON_OFFERED_INCOMPATIBLE_QOS events

jaisontj:
Related to [this](https://github.com/ros2/ros2/issues/822) feature request. The design and implementation details can also be found there.

- Added new enum type `RMW_EVENT_REQUESTED_INCOMPATIBLE_QOS` and `RMW_EVENT_OFFERED_INCOMPATIBLE_QOS` to rmw_event_type_t
- Defined new structures that encapsulate data for the new events above

Signed-off-by: Jaison Titus <jaisontj92@gmail.com>


dirk-thomas:
Please reference all related PR from at least one ticket for visibility.

Also for an RMW API change please include significantly more context / information.

jaisontj:
Hi @dirk-thomas, 
Apologies for the lack of context, I planned on adding more information and that is why I had open this as a draft PR.

I've now opened up an issue [here](https://github.com/ros2/ros2/issues/822) tracking all of the related PRs as well as detailing out the need for this feature and the implementation details. Intention is also to initiate a design discussion towards this.

rmw/include/rmw/types.h
```diff
@@ -400,6 +400,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_requested_deadline_missed_status_t
...
 } rmw_requested_deadline_missed_status_t;
 
+/// QoS Requested Incompatible QoS information provided by a subscription.
+typedef struct RMW_PUBLIC_TYPE rmw_requested_incompatible_qos_status_t
+{
+  /**
+       * Total cumulative number of times the concerned subscription discovered a
+       * publisher for the same topic with an offered QoS that was incompatible
+       * with that requested by the subscription..
+       */
```
prajakta-gokhale: Fix indentation here and all other comment blocks.

rmw/include/rmw/types.h
```diff
@@ -400,6 +400,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_requested_deadline_missed_status_t
   int32_t total_count_change;
 } rmw_requested_deadline_missed_status_t;
 
+/// QoS Requested Incompatible QoS information provided by a subscription.
```
piraka9011: nit: by a subscriber
(Unless that's the terminology used elsewhere)

jaisontj: So, in ros; the terminology to be used is publisher and subscription. This seems to have been used interchange-ably but according to the conversation [here](https://github.com/ros2/rmw/pull/186#issuecomment-541947877), subscription seems to the right one.

Quoting the relevant para: "Note: the name convention in ROS2 is subscription not subscriber. There is no subscriber in ROS2. Unfortunately it seems that a handful of functions that mention subscriber have made it into rmw. Lets make sure not to continue the trend."

rmw/include/rmw/types.h
jaisontj: Done! 

ps: Strangely that indentation was created by uncrustify :/

rmw/include/rmw/types.h
piraka9011: Good to know :)

jaisontj: Yeah, I didn't either. I was using `subscriber` everywhere, haha.

rmw/include/rmw/types.h
```diff
@@ -400,6 +412,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_requested_deadline_missed_status_t
...
+   * with that requested by the subscription..
+   */
+  int32_t total_count;
+  /**
+   * The change in total_count since the last time the status was read.
+   */
+  int32_t total_count_change;
+  /**
+   * The Qos Policy Id of one of the policies that was found to be
+   * incompatible the last time an incompatibility was detected
```
ivanpauno: Missing period.

rmw/include/rmw/types.h
```diff
@@ -400,6 +412,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_requested_deadline_missed_status_t
...
+   */
+  int32_t last_policy_id;
+} rmw_requested_incompatible_qos_status_t;
```
ivanpauno: Not mandatory, but it would be good to stop populating an already big `types.h`.
Maybe create an `incompatible_qos_events_statuses.h`, with the three types added to this file.

rmw/include/rmw/types.h
```diff
@@ -295,6 +295,18 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
...
+/// RMW equivalent of the DDS QoS Policy IDs
+enum RMW_PUBLIC_TYPE rmw_qos_policy_id_t
+{
+  RMW_QOS_POLICY_ID_INVALID    = 0,
+  RMW_QOS_POLICY_ID_DURABILITY = 2,
+  RMW_QOS_POLICY_ID_DEADLINE   = 4,
+  RMW_QOS_POLICY_ID_LIVELINESS = 8,
+  RMW_QOS_POLICY_ID_RELIABILITY= 11,
+  RMW_QOS_POLICY_ID_HISTORY    = 13,
+  RMW_QOS_POLICY_ID_LIFESPAN   = 21
```
ivanpauno: Why those strange numbers?
If they were taken from an specific implementation, please use contiguous numbers here, and convert them in each implementation.

jaisontj: I took these values from the [DDS implementation spec](https://www.omg.org/spec/DDS/1.4/PDF) assuming that all RMW implementations will follow the same values. However, that may not be a correct assumption.

ivanpauno: We shouldn't assume a DDS based implementation.
The numbers in `rmw` can be just made up by you (if you decide to keep the ones in the dds spec that's ok, but add a comment stating that).
The rmw implementations should have a map between the ros ids and the specific implementations ids, using the macros that both provide.
In that way, we don't depend on future changes of the numbers.

rmw/include/rmw/types.h
```diff
@@ -295,6 +295,18 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
 #define RMW_QOS_LIFESPAN_DEFAULT {0, 0}
 #define RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT {0, 0}
 
+/// RMW equivalent of the DDS QoS Policy IDs
```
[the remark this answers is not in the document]
wjwwood: I would say we should not mention DDS in the rmw API or its doc strings, as the underlying middleware may not be DDS. So we should just define what makes sense for our layer and if that happens to be similar to DDS's then we can mention that, but our interface should stand on its own.

rmw/include/rmw/types.h
wjwwood: 👍

rmw/include/rmw/types.h
```diff
@@ -295,6 +295,18 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
...
+enum RMW_PUBLIC_TYPE rmw_qos_policy_id_t
```
wjwwood: I'd use the word `kind` rather than `id`. For an analogy, if we had an enum for colors we'd not call the enum `rmw_color_id_t`, we'd just use `rmw_color_kind_t` or `rmw_color_t` most likely. So I think this should probably just be `rmw_qos_policy_kind_t` or `rmw_qos_policy_t`.

rmw/include/rmw/types.h
```diff
@@ -295,6 +295,18 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
...
+{
+  RMW_QOS_POLICY_ID_INVALID    = 0,
```
wjwwood: If we take my other recommendation, these should become `RMW_QOS_POLICY_X` or `RMW_QOS_POLICY_KIND_X` where `X` is each kind, like `INVALID` or `RELIABILITY`.

rmw/include/rmw/types.h
wjwwood: Just in case you guys are wondering about the logic there, I put it as follows:

A node is the "subscriber" in ROS, because it "subscribes" to a topic which results in a "subscription" which is treated sort of like a receipt of the act. A "subscription" is not used to "subscribe" in the future, hence it is not a "subscriber". In the future, we might have a "subscriber" object that is used to create new subscriptions, but no such thing exists at the moment. Instead the node serves this purpose.

Conversely, a "publisher" is the object which is the result of a node "advertising a new publication" (borrowing some from ROS 1 terminology) on a topic, and the "publisher" is actually used to publish data in the future, and is called as such because it publishes data (as opposed to advertising more publications). Mirroring subscriptions, in the future we may have an "advertiser" object, but for now that is another role of the node and no such object exists.

Hopefully that makes sense. It's an somewhat awkward mixture of ROS 1 and DDS terminology, with some compromises made to make it make as much sense as possible when explained as above.

rmw/include/rmw/types.h
wjwwood: Is there any logic to which policy id will be here if more than one are incompatible?

rmw/include/rmw/types.h
wjwwood: Agreed.

rmw/include/rmw/event.h
```diff
@@ -35,10 +35,12 @@ typedef enum rmw_event_type_t
   // subscription events
   RMW_EVENT_LIVELINESS_CHANGED,
   RMW_EVENT_REQUESTED_DEADLINE_MISSED,
+  RMW_EVENT_REQUESTED_INCOMPATIBLE_QOS,
```
wjwwood: This is hard for me to read and make sense of and it is worse in some of the struct descriptions below.

Would another wording make more sense? Maybe `RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE` would be better. I would read it as:

> An `RMW_EVENT` occurred due to a `REQUESTED_QOS` being `INCOMPATIBLE` with the subscription's QoS.

Which I think matches, for example, the deadline one better, which I would read as:

> An `RMW_EVENT` occurred due to a `REQUESTED_DEADLINE` being `MISSED`.

---

Extending this to the other new event, I'd call it `RMW_EVENT_OFFERED_QOS_INCOMPATIBLE`.

rmw/include/rmw/types.h
```diff
@@ -400,6 +412,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_requested_deadline_missed_status_t
   int32_t total_count_change;
 } rmw_requested_deadline_missed_status_t;
 
+/// QoS Requested Incompatible QoS information provided by a subscription.
+typedef struct RMW_PUBLIC_TYPE rmw_requested_incompatible_qos_status_t
```
wjwwood: Maybe you're just basing this off the DDS names, which a good place to start, but status doesn't really make sense to me here. I think that comes from this being the result of a "StatusCondition" in DDS. However, this strikes me as state more than status. I'm not sold on this (I know other similar structs may already use `status` rather than `state`), but I think `state` or `event_state` are better descriptors. What do other think?

Also, following from my other comment about naming, I'd swap them here too to be like this:

```c
/// Event state for a subscription's 'RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE' events.
typedef struct RMW_PUBLIC_TYPE rmw_requested_qos_incompatible_event_state_t
{
  ...
```

Or we could leave `status` in place of `state` and get:

```c
/// Event status for a subscription's 'RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE' events.
typedef struct RMW_PUBLIC_TYPE rmw_requested_qos_incompatible_event_status_t
{
  ...
```

Both options are an improvement in "at-a-glance comprehension", at least for me.

rmw/include/rmw/types.h
```diff
@@ -400,6 +412,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_requested_deadline_missed_status_t
...
+{
+  /**
+   * Total cumulative number of times the concerned subscription discovered a
+   * publisher for the same topic with an offered QoS that was incompatible
+   * with that requested by the subscription..
```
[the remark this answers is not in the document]
mm318: Done.

rmw/include/rmw/types.h
```diff
@@ -428,6 +460,26 @@ typedef struct RMW_PUBLIC_TYPE rmw_offered_deadline_missed_status_t
...
+   * with that offered by the publisher.
+   */
+  int32_t total_count;
+  /**
+   * The change in total_count since the last time the status was read.
+   */
+  int32_t total_count_change;
+  /**
+   * The Qos Policy Id of one of the policies that was found to be
+   * incompatible the last time an incompatibility was detected
```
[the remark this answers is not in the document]
mm318: Done.

rmw/include/rmw/types.h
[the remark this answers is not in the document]
mm318: I have changed these values to one-hot bit values, to possibly address the question at https://github.com/ros2/rmw/pull/193#discussion_r370904092 later.

rmw/include/rmw/types.h
mm318: Done.

rmw/include/rmw/types.h
mm318: Done.

rmw/include/rmw/types.h
mm318: I have changed these values to one-hot bit values, to possibly address the question at https://github.com/ros2/rmw/pull/193#discussion_r370904092 later.

rmw/include/rmw/types.h
mm318: Done.

rmw/include/rmw/types.h
mm318: I have changed these enum names to `RMW_QOS_POLICY_<X>`.

rmw/include/rmw/types.h
mm318: I am thinking of `last_policy_kind` (renamed from `last_policy_id`) as a bitmask, where each 1 bit represents a different policy kind that was incompatible.

rmw/include/rmw/event.h
mm318: Done.

rmw/include/rmw/types.h
mm318: Swapped the positions of `incompatible` and `qos`, but let's keep the use of `status` for consistency.

rmw/include/rmw/incompatible_qos_events_statuses.h
```diff
@@ -0,0 +1,79 @@
...
+  int32_t total_count_change;
+  /**
+   * The Qos Policy Kind of one of the policies that was found to be
+   * incompatible the last time an incompatibility was detected.
+   */
+  int32_t last_policy_kind;
+} rmw_requested_qos_incompatible_event_status_t;
+
+/// Event state for a publisher's 'RMW_EVENT_OFFERED_QOS_INCOMPATIBLE' events.
+typedef struct RMW_PUBLIC_TYPE rmw_offered_qos_incompatible_event_status_t
```
ivanpauno: nit: I think it's good to have a different struct for offered/requested, but as currently the definitions are the same, can't we just `typedef` the above `struct`?

mm318: Done.

mm318:
Updated set of pull requests:
- https://github.com/ros2/rmw_connext/pull/398
- https://github.com/ros2/rmw_cyclonedds/pull/125
- ~~https://github.com/ros2/rmw_opensplice/pull/294~~
- https://github.com/ros2/rmw/pull/193
- https://github.com/ros2/rcl/pull/535
- https://github.com/ros2/rclcpp/pull/924
- https://github.com/ros2/rclpy/pull/459
- https://github.com/ros2/ros2cli/pull/410
- https://github.com/ros2/demos/pull/416

TODOs:
- Add [default incompatible QoS callback](https://github.com/ros2/ros2/issues/822#issuecomment-598024481) to `rclcpp`/`rclpy`
- Verify functionality of `ros2topic` changes (similar to the changes in `quality_of_service_demo_cpp`/`quality_of_service_demo_py`)
- Re-run tests for the repos

mm318:
@ros2/aws-oncall - please run this CI job
Gist: `https://gist.githubusercontent.com/mm318/6e086798710e344feb2df41f94baf717/raw/26f6df7bdb054887b226bac7fc22fca942c04384/ros2_qos.repos`
BUILD args: `--packages-up-to rclcpp rclpy`
TEST args: `--packages-select rmw_connext_cpp rmw_connext_shared_cpp rmw_cyclonedds_cpp rmw rcl`
Job: `ci_launcher`

dabonnie:
> https://gist.githubusercontent.com/mm318/6e086798710e344feb2df41f94baf717/raw/26f6df7bdb054887b226bac7fc22fca942c04384/ros2_qos.repos

https://ci.ros2.org/job/ci_launcher/5003/console

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9792/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5414/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7958/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9652/)

rmw/include/rmw/incompatible_qos_events_statuses.h
```diff
@@ -0,0 +1,68 @@
...
+   * publisher for the same topic with an offered QoS that was incompatible
+   * with that requested by the subscription.
+   */
+  int32_t total_count;
+  /**
+   * The change in total_count since the last time the status was read.
+   */
+  int32_t total_count_change;
+  /**
+   * The Qos Policy Kind of one of the policies that was found to be
```
emersonknapp: since you've changed the enum to bitflags, is the intent to only return "one of the policies" or is it now intended to return a bitfield with "all of the policies" that were incompatible?

mm318: Defining it as a bitflags opens up the possibility of returning multiple policies. However, currently the `rmw` implementations are still only capable of returning one policy.

mm318:
Hi @ivanpauno and @wjwwood, do you think the changes up to `rcl` are ready to merge? I mean the following pull requests:
- https://github.com/ros2/rmw_connext/pull/398
- https://github.com/ros2/rmw_cyclonedds/pull/125
- https://github.com/ros2/rmw/pull/193
- https://github.com/ros2/rcl/pull/535

The test results are here: https://github.com/ros2/rmw/pull/193#issuecomment-604621384. I don't think the test failures on Windows and macOS are related to these changes.

ivanpauno:
@mm318 SGTM! CI failures also look unrelated to me.
@wjwwood Do you think that does PRs are ready to be merged?

wjwwood:
Fine by me, I trust you guys to review and merge it since my initial review.

mm318:
> Fine by me, I trust you guys to review and merge it since my initial review.

Let's merge it then please, @ivanpauno. Thanks!

ivanpauno:
@mm318 I merged those four PRs.
Please, open a issue in `rmw_fastrtps`, to track the future addition of these events.

mm318:
Thanks!

I have opened https://github.com/ros2/rmw_fastrtps/issues/356 to track the future addition of support for these types of QoS events.

PR #199, opened 2020-03-06
add option to filter available RMW implementations

dirk-thomas:
This allows to e.g. only test with a subset of RMW implementations in an overlay workspace - even if the underlay workspace contains more RMW implementations.

### DIFF
--- a/rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
+++ b/rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
@@ -54,7 +54,7 @@ function(get_available_rmw_implementations var)
       set(middleware_implementations ${filtered})
     endforeach()
     message(STATUS
-      "Override available RMW implementations: ${middleware_implementations}")
+      "Filtered available RMW implementations: ${middleware_implementations}")
   endif()
 
   set(${var} ${middleware_implementations} PARENT_SCOPE)
--- a/rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
+++ b/rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
@@ -15,6 +15,12 @@
 #
 # Get the package names of the available ROS middleware implemenations.
 #
+# The result can be overridden by setting either a CMake or environment
+# variable named ``RMW_IMPLEMENTATIONS``.
+# The variable can contain RMW implementation names separated by the platform
+# specific path separator.
+# Including an unavailable RMW implementation results in a fatal error message.
+#
 # :param var: the output variable name containing the package names
 # :type var: list of strings
 #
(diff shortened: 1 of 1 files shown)


rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
```diff
@@ -24,5 +30,32 @@ function(get_available_rmw_implementations var)
     list(REMOVE_ITEM middleware_implementations "rmw_implementation")
     list(SORT middleware_implementations)
   endif()
+
+  # use explicitly provided list if provided
+  # option()
```
jacobperron: I think this line can be removed.

rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
```diff
@@ -24,5 +30,32 @@ function(get_available_rmw_implementations var)
...
+    endforeach()
+    set(filtered)
+    foreach(rmw_implementation ${middleware_implementations})
+      if("${rmw_implementation}" IN_LIST RMW_IMPLEMENTATIONS)
+        list(APPEND filtered "${rmw_implementation}")
+      endif()
+      set(middleware_implementations ${filtered})
+    endforeach()
+    message(STATUS
+      "Override available RMW implementations: ${middleware_implementations}")
```
jacobperron:
```suggestion
      "Filtered available RMW implementations: ${middleware_implementations}")
```

rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
dirk-thomas: In several places we mark configuration options in CMake with this comment. Classically for `catkin` there was even a script which extracted those for documentation purposes. Even without such scripts this helps to find CMake configuration options throughout the code base. So I am inclined to keep it as-is.

rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
dirk-thomas: Applied in 6bd6abeebee4a45b4196c93dde66e27336839d88 to pass DCO.

rmw_implementation_cmake/cmake/get_available_rmw_implementations.cmake
jacobperron: Okay. I just thought it was vestigial.

dirk-thomas:
> with CI to confirm nothing broke and linters are happy

[Build Status](https://ci.ros2.org/job/ci_linux/9561/)

jacobperron:
@dirk-thomas Seeing how this is being used, what do you think about renaming the variable so it is less likely to be confused with `RMW_IMPLEMENTATION`? (see https://github.com/ros2/rclpy/pull/522#discussion_r389198482)

dirk-thomas:
> Seeing how this is being used, what do you think about renaming the variable so it is less likely to be confused with `RMW_IMPLEMENTATION`? (see ros2/rclpy#522 (comment))

See my comment there: https://github.com/ros2/rclpy/pull/522#discussion_r389199639

PR #200, opened 2020-03-10
Add timestamps to rmw-internal structures.

iluetkeb:
This goes together with https://github.com/ros2/rcl/pull/591

### DIFF
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -199,7 +199,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_services_t
   /// Pointer to an array of void * pointers of services.
   void ** services;
   /// Pointer to an array rcutil timestamps, to be filled in same order as the above
-  rcutils_time_point_value_t *timestamps;
+  rcutils_time_point_value_t * timestamps;
 } rmw_services_t;
 
 /// Array of client handles.
@@ -216,7 +216,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_clients_t
   /// Pointer to an array of void * pointers of clients.
   void ** clients;
   /// Pointer to an array rcutil timestamps, to be filled in same order as the above
-  rcutils_time_point_value_t *timestamps;
+  rcutils_time_point_value_t * timestamps;
 } rmw_clients_t;
 
 typedef struct RMW_PUBLIC_TYPE rmw_events_t
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -208,7 +208,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_services_t
 
 /// Array of client handles.
 /**
- * An array of void * pointers representing type-erased middleware-specific clients.
+ * An array of void * pointers representing type-erased middleware-specific clients,
  * and arrival timestamp information for the entries (in the same order).
  * The number of non-null entries may be smaller than the allocated size of the array.
  * The number of clients represented may be smaller than the allocated size of the array.
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -338,7 +338,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
 
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
-  const rmw_time_t received_timestamp;
+  rmw_time_t received_timestamp;
   rmw_gid_t publisher_gid;
   bool from_intra_process;
 } rmw_message_info_t;
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -338,7 +338,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
 
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
-  // const rmw_time_t received_timestamp;
+  const rmw_time_t received_timestamp;
   rmw_gid_t publisher_gid;
   bool from_intra_process;
 } rmw_message_info_t;
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -221,7 +221,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_clients_t
   size_t client_count;
   /// Pointer to an array of void * pointers of clients.
   void ** clients;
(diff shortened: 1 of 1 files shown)


ralph-lange:
Connects to https://github.com/ros2/design/issues/259.

rmw/include/rmw/types.h
```diff
@@ -179,6 +179,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscriptions_t
   size_t subscriber_count;
   /// Pointer to an array of void * pointers of subscriptions.
   void ** subscribers;
+  /// Pointer to an array rcutil timestamps, to be filled in same order as the above
```
wjwwood: Here and below.

```suggestion
  /// Pointer to an array of rcutils timestamps, to be filled in same order as the above.
```

rmw/include/rmw/types.h
```diff
@@ -179,6 +179,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscriptions_t
...
+  rcutils_time_point_value_t *timestamps;
```
wjwwood: Our style checking linter will choke on this otherwise. Here and below.

```suggestion
  rcutils_time_point_value_t * timestamps;
```

iluetkeb:
> Also, we should consider any changes to tests and documentation.

In principle, I would test that a) allocation and deallocation works and b) that the fields are filled properly.

However, it seems that a) happens in rcl and b) happens in the rmw implementation. So it doesn't appear as if I can test anything in rmw.

Suggestions to the contrary very welcome, I want this to be solid :-)

rmw/include/rmw/types.h
```diff
@@ -194,6 +196,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_services_t
   size_t service_count;
   /// Pointer to an array of void * pointers of services.
   void ** services;
+  /// Pointer to an array rcutil timestamps, to be filled in same order as the above
```
wjwwood: Sorry I didn't comment on these explicitly because they were the same as above, but this line and at least one other needs fixups too.

rmw/include/rmw/types.h
```diff
@@ -342,7 +348,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
 
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
-  // const rmw_time_t received_timestamp;
+  const rmw_time_t received_timestamp;
```
wjwwood: I forgot to mention, we should probably drop the `const` here. There's been a few discussions around this point and I think we've landed on not making members like this const. Also it would keep it consistent with the other two fields in this struct.

iluetkeb: Done

rmw/include/rmw/types.h
wjwwood: This is still pending I think.

iluetkeb: What exactly is pending here? You commented earlier about the whitespace around the "*". That should have been fixed for all the entries. Is there something wrong with the comment itself? 

btw, I ran the tests for rmw and rcl on my machine, and nothing came up in rcl or rmw. There are a few failures in other packages, but I reckon those are just the current state of Foxy.

wjwwood:
```suggestion
  /// Pointer to an array of rcutils timestamps, to be filled in same order as the above.
```

rmw/include/rmw/types.h
```diff
@@ -168,47 +168,59 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscription_allocation_t
...
  * The creator of this struct is responsible for allocating and deallocating the array.
+ * including the timestamp array.
  */
 typedef struct RMW_PUBLIC_TYPE rmw_clients_t
 {
   /// The number of clients represented by the array.
   size_t client_count;
   /// Pointer to an array of void * pointers of clients.
   void ** clients;
+  /// Pointer to an array of rcutil timestamps, to be filled in same order as the above
```
wjwwood:
```suggestion
  /// Pointer to an array of rcutils timestamps, to be filled in same order as the above.
```

rmw/include/rmw/types.h
```diff
@@ -168,47 +168,59 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscription_allocation_t
...
+  rcutils_time_point_value_t * timestamps;
 } rmw_services_t;
 
 /// Array of client handles.
 /**
  * An array of void * pointers representing type-erased middleware-specific clients.
+ * and arrival timestamp information for the entries (in the same order).
  * The number of non-null entries may be smaller than the allocated size of the array.
  * The number of clients represented may be smaller than the allocated size of the array.
  * The creator of this struct is responsible for allocating and deallocating the array.
```
wjwwood:
```suggestion
 * The creator of this struct is responsible for allocating and deallocating the array,
```

rmw/include/rmw/types.h
```diff
@@ -168,47 +168,59 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscription_allocation_t
...
   size_t service_count;
   /// Pointer to an array of void * pointers of services.
   void ** services;
+  /// Pointer to an array of rcutil timestamps, to be filled in same order as the above
+  rcutils_time_point_value_t * timestamps;
 } rmw_services_t;
 
 /// Array of client handles.
 /**
  * An array of void * pointers representing type-erased middleware-specific clients.
```
wjwwood:
```suggestion
 * An array of void * pointers representing type-erased middleware-specific clients,
```

rmw/include/rmw/types.h
```diff
@@ -168,47 +168,59 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscription_allocation_t
...
+ * The creator of this struct is responsible for allocating and deallocating the array,
+ * including the timestamp array.
  */
 typedef struct RMW_PUBLIC_TYPE rmw_services_t
 {
   /// The number of services represented by the array.
   size_t service_count;
   /// Pointer to an array of void * pointers of services.
   void ** services;
+  /// Pointer to an array of rcutil timestamps, to be filled in same order as the above
```
wjwwood:
```suggestion
  /// Pointer to an array of rcutils timestamps, to be filled in same order as the above.
```

iluetkeb:
Closed in favor of https://github.com/ros2/rmw/pull/214

PR #207, opened 2020-03-26
add topic endpoint info to rmw

Karsten1987:
https://github.com/ros2/rmw/pull/186 introduces two functions to the RMW interface which are not present when including `rmw/rmw.h`.
For convenience to the rmw implementer, I think it makes sense to include all functions when including `rmw/rmw.h`.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -94,6 +94,7 @@ extern "C"
 #include "rosidl_generator_c/message_type_support_struct.h"
 #include "rosidl_generator_c/service_type_support_struct.h"
 
+#include "rmw/get_topic_endpoint_info.h"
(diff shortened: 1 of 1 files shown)


ivanpauno:
I'm not quite in favor of "huge" headers, as they make compilation times much slower.
IMO, we shouldn't keep adding things to `rmw/rmw.h` and similar, but I'm ok with this if somebody else approves.

emersonknapp:
@ivanpauno @Karsten1987 @wjwwood  should we close this issue? I assume it's no longer relevant.

fujitatomoya (commented):
I will go ahead to close this. user can include the header if they want to use introspection of ROS graph and with https://github.com/ros2/rmw/pull/207#issuecomment-749207431.

PR #208, opened 2020-03-29
add comment about RMW_RET_UNSUPPORTED for loaned_message.

fujitatomoya:
just adding some comment for RMW_RET_UNSUPPORTED corresponds to most of the rmw_implementation.


fujitatomoya:
@Karsten1987 

appreciate for the quick response!

PR #212, opened 2020-04-10
Add API for taking a sequence of messages

mjcarroll:
This will allow users to take multiple messages in one take call if they
are available in the underlying middleware, versus having to make
multiple rmw_take calls in a row.

Signed-off-by: Michael Carroll <michael@openrobotics.org>

PR #214, opened 2020-04-14
Add timestamps to message info

iluetkeb:
See https://github.com/ros2/design/issues/259

This replaces https://github.com/ros2/rmw/pull/200 as discussed.

### DIFF
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -332,11 +332,17 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
 
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
-  const rmw_time_t source_timestamp;
-  const rmw_time_t received_timestamp;
+  rmw_time_t source_timestamp;
+  rmw_time_t received_timestamp;
   rmw_gid_t publisher_gid;
   bool from_intra_process;
 } rmw_message_info_t;
+
+/// Get zero initialized mesage info.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_message_info_t
+rmw_get_zero_initialized_message_info(void);
 
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_sequence_t
 {
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -332,7 +332,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
 
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
-  // const rmw_time_t received_timestamp;
+  const rmw_time_t source_timestamp;
+  const rmw_time_t received_timestamp;
   rmw_gid_t publisher_gid;
   bool from_intra_process;
 } rmw_message_info_t;
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -330,10 +330,12 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_t
   uint8_t data[RMW_GID_STORAGE_SIZE];
 } rmw_gid_t;
 
+typedef rcutils_time_point_value_t rmw_time_point_value_t;
+
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_t
 {
-  rmw_time_t source_timestamp;
-  rmw_time_t received_timestamp;
+  rmw_time_point_value_t source_timestamp;
+  rmw_time_point_value_t received_timestamp;
   rmw_gid_t publisher_gid;
   bool from_intra_process;
 } rmw_message_info_t;
--- a/rmw/CMakeLists.txt
+++ b/rmw/CMakeLists.txt
@@ -38,6 +38,7 @@ set(rmw_sources
   "src/subscription_options.c"
   "src/topic_endpoint_info_array.c"
   "src/topic_endpoint_info.c"
+  "src/types.c"
   "src/validate_full_topic_name.c"
   "src/validate_namespace.c"
   "src/validate_node_name.c"
--- a/rmw/src/types.c
+++ b/rmw/src/types.c
@@ -19,6 +19,6 @@ RMW_WARN_UNUSED
 rmw_message_info_t
 rmw_get_zero_initialized_message_info(void)
 {
-  rmw_message_info_t zero_initialized_message_info = {{0, 0}, {0, 0}, {NULL, {0}}, false};
+  rmw_message_info_t zero_initialized_message_info = {0, 0, {NULL, {0}}, false};
   return zero_initialized_message_info;
 }
--- a/rmw/src/types.c
+++ b/rmw/src/types.c
@@ -1,0 +1,24 @@
+// Copyright 2020 Robert Bosch GmbH. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rmw/types.h"
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_message_info_t
+rmw_get_zero_initialized_message_info(void)
+{
+  rmw_message_info_t zero_initialized_message_info = {{0, 0}, {0, 0}, {NULL, {0}}, false};
+  return zero_initialized_message_info;
+}


rotu:
BTW, I think `rmw_time_t` represents a *duration* - and it might be more appropriate to use `rcutils_time_point_value_t`.

rmw/include/rmw/types.h
```diff
@@ -243,17 +243,25 @@ typedef struct RMW_PUBLIC_TYPE rmw_wait_set_t
...
+typedef struct RMW_PUBLIC_TYPE rmw_time_t
+{
+  uint64_t sec;
+  uint64_t nsec;
+} rmw_time_t;
+
 typedef struct RMW_PUBLIC_TYPE rmw_request_id_t
 {
+  rmw_time_t source_timestamp;
+  rmw_time_t received_timestamp;
```
rotu, line 255: I don't think this is an appropriate place for these timestamps,
1. A timestamp is not part of an "ID" so at the very least it the struct would need to be renamed.
2. These timestamps are only applicable for `rmw_take_request` so it's confusing to have them be part of `rmw_send_response`/`rmw_take_response`.

It might make sense to add an output `rmw_message_info_t *` to the functions `rmw_take_request` and `rmw_take_response` instead. (or add a parallel function like `rmw_take_with_info`)

iluetkeb: > A timestamp is not part of an "ID" so at the very least it the struct would need to be renamed.

Agreed. I saw these structures referred to as "header" in several places, that's why I thought they are appropriate in general, but I agree that a renaming would be be better.

If nobody objects, that would be my suggestion to go forward.

> These timestamps are only applicable for rmw_take_request so it's confusing to have them be part of rmw_send_response/rmw_take_response

Agree on `send_response`, but why should they not be applicable to `take_response`? The response is a regular message and has these timestamps.

> It might make sense to add an output rmw_message_info_t * to the functions rmw_take_request and rmw_take_response instead. (or add a parallel function like rmw_take_with_info)

Using `rmw_message_info_t` would duplicate the sender ID, so that doesn't seem ideal either.

Since we have three functions where the current structure is appropriate (albeit with renaming), and only one where it isn't, I would prefer changing only the one (`send_response`).

Personally, I'm not a big fan of the `with_info` methods. They are only in `rmw` anyway, `rcl` uses a "pass nullptr if you don't want this" idiom, which I think is really a fine approach here and would substantially reduce clutter.

rotu:
> Agree on `send_response`, but why should they not be applicable to `take_response`?

Because these are request ID timestamps (being in the `rmw_request_id_t` struct). If you mean timestamps for the response message in a `rmw_message_info_t`, I support that.

I have to think about duplicating the publisher ID. I think the intent is to identify the *service*, which might or might not correspond with a publisher ID. AFAICT, nobody defined how “services” map to DDS entities, and I don’t know what that publisher IDs are actually used for.

> Personally, I'm not a big fan of the with_info methods.

:+1: maybe there’s a principled reason they exist, but I hate them too. I’d rather keep the number of RMW methods down. Leave them null or just ignore the value. (But please document if the argument is nullable).

rotu:
I’m implementing, it occurs that this breaks the time abstraction https://design.ros2.org/articles/clock_and_time.html since ROS Time is not defined at the RMW level. I’m not sure what if anything should be done about this.

iluetkeb:
> I’m implementing, it occurs that this breaks the time abstraction

Yeah, I've worried a bit about that as well. It's not so easy to fix that.

However, since we're not planning on giving these timestamps to the application, it should not really matter. They are only used by the executor and there, only ordering does matter, really.

In any case, if we intend to implement the clock abstraction at some point, we should be able to do so without breaking ABI. Therefore, right now I'm more thinking about whether we need to change from `rmw_time_t` to `rcutils_time_point_value_t`.

rmw/include/rmw/types.h
wjwwood: I think it should be structured like this (ideally):

- rmw_request_info_t
  - rmw_request_id_t
    - sequence_number
    - requester_guid (was writer_guid)
  - source_timestamp
  - received_timestamp
- rmw_response_info_t
  - rmw_request_id_t
    - sequence_number
    - requester_guid (was writer_guid)
  - source_timestamp
  - received_timestamp

Now, we could change that up (maybe de-duplicate the types since they're the same or w/e), but I think my take away is that we need more discussion on this item. And I wouldn't tie it to the message info changes.

wjwwood: Also, replying to these:

> The response is a regular message and has these timestamps.

It's a message, but not necessarily one that came from a topic. In Connext we don't use topics directly, we go through RTI's RPC API, and we do something similar with Fast-RTPS (both happen to use topics under the hood but that's an implementation detail) and in OpenSplice we implement our own using topics and a wrapper around the user type that include the request id and stuff. Because it is part of the rmw API services could be different though, maybe implemented with an architecture better suited to request-reply, like grpc or something. So don't just assume it is a message from a topic.

I think we can still rely on it having a source and received timestamp, and make the implementation figure out how to get that info, but I just wanted to check that assumption.

> Personally, I'm not a big fan of the with_info methods. They are only in rmw anyway, rcl uses a "pass nullptr if you don't want this" idiom, which I think is really a fine approach here and would substantially reduce clutter.

Agreed, that's a historical thing, I think we should just always take the info.

rotu:
> so we don't put the message info changes at schedule risk

@wjwwood for Earth 2.0, let’s enable `use_sim_time`, okay?

iluetkeb:
>  for Earth 2.0, let’s enable use_sim_time, okay?

Can you start a design issue on that? Otherwise we probably forget.

wjwwood:
> AFAICT, nobody defined how “services” map to DDS entities, and I don’t know what sender IDs are actually used for nor how this is supposed to behave if a service has multiple servers available

Yeah, I didn't write that part of the rmw interface, but my guess is that "writer_guid" should be "requester_guid" and it should identify the sender of the request, this might be a DDS entity guid, like a data writer, or it might be something else, but the goal is that with the writer/requester id and a sequence number that it uniquely identifies a single request.

Separately we might need a responder id so the requester can know who responded, but that's not strictly needed I think. The response does need to include the writer/requester id and the sequence number so it can ensure the response is intended for themselves and to which request the response should be delivered, in the case of concurrent requests.

> I’m implementing, it occurs that this breaks the time abstraction https://design.ros2.org/articles/clock_and_time.html since ROS Time is not defined at the RMW level. I’m not sure what if anything should be done about this.

I don't think we need source and received timestamps in ROS time, I think system time is ok instead of ROS time.

However, we have to keep in mind that system time can slew and jump, and also that it has leap seconds. Perhaps there should be a third timestamp which is `steady_received_timestamp` which uses a monotonically increasing steady clock? You could imagine a fourth `steady_source_timestamp`, but that would only be useful for comparing time deltas in multiple messages as it would not be associated with a specific system time stamp, though it should be "close" to the `source_timestamp`.

I've long wanted to have an option in our software stack for TAI time, which is both a time point (uses unix epoch) but also is monotonically increasing (no leap seconds).

> In any case, if we intend to implement the clock abstraction at some point, we should be able to do so without breaking ABI. Therefore, right now I'm more thinking about whether we need to change from rmw_time_t to rcutils_time_point_value_t.

Since these times are time points and not durations (because we're using system time), using `rcutils_time_point_value_t` is ok. I'm not sure why we are still using the split struct style in rmw, maybe because that's what the messages are using, or maybe I just never had time to push down my "single variable timestamp" ideas into rmw from rcl.

I would recommend doing a `#define rmw_time_point_value_t rcutils_time_point_value_t` and similar aliases for the related time functions, and then use `rmw_time_point_value_t` in the struct. Just in case we need to refactor and not use `rcutils` (or change the name or something) in the future.

wjwwood:
Ok, here's CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10386/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5901/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8470/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10284/)

This CI includes these pull requests:

- https://github.com/ros2/rmw/pull/214
- https://github.com/ros2/rcl/pull/619
- https://github.com/ros2/rclpy/pull/542
- https://github.com/ros2/rmw_fastrtps/pull/368
- https://github.com/ros2/rmw_cyclonedds/pull/157

And builds up to `demo_nodes_cpp` and `demo_nodes_py`, just to check things out quickly.

Also, it only builds with Fast-RTPS static, as the cyclone pr doesn't look to be updated with the new timestamp type.

wjwwood:
I was going to try and do the message info along with the service info pr's, but there's too much missing still for the service info, e.g. connext is not working, and I cannot edit the bosch forks, so I have been spending a lot of time making my own. Also, I cannot edit the cyclone pr that @iluetkeb put forward, but there are issues with it, so I won't take it as-is either.

Instead, here is CI building and testing everything including only this pr, the associated ones above it, and the fast rtps changes:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10423/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5929/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8501/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10316/)

To be specific this includes:

- #214
- ros2/rcl#619
- ros2/rclpy#542
- ros2/rmw_fastrtps#368

wjwwood:
So, maybe services with Connext is further along than I thought, but I was missing this pull request: https://github.com/ros2/rosidl_typesupport_connext/pull/53

Either way, I think trying to get messages tested before nightly's run was the best bet.

wjwwood:
Ok another round of CI, building and testing with `--packages-up-to test_rclcpp demo_nodes_cpp demo_nodes_py`, including cyclone and Connext too:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10451/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5943/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8528/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10334/)

This one now includes @iluetkeb's cyclone pr:

- #214
- ros2/rcl#619
- ros2/rclpy#542
- ros2/rmw_fastrtps#368
- ros2/rmw_cyclonedds#163

iluetkeb:
This has a patch failure with connext. Is that my failure?

iluetkeb:
@wjwwood This fails with a connext patch failure, but I didn't change rmw_connext for these PRs, so it shouldn't be me...

wjwwood:
Not sure I’m checking.

wjwwood:
I don’t think it was my ci setting or your changes. I’m raising it up the chain.

iluetkeb:
https://github.com/ros2/rmw/pull/212 seemed to have a similar issue.

wjwwood:
@iluetkeb there are linter failures...

https://ci.ros2.org/job/ci_linux-aarch64/5943/testReport/junit/(root)/projectroot/flake8/
https://ci.ros2.org/job/ci_linux-aarch64/5943/testReport/junit/(root)/projectroot/uncrustify/

Are you still around to have a look, or should I?

Please try to run the tests (at least the linters) locally before pushing things. Our CI turn around is very long to catch things like this which can easily be caught locally.

wjwwood:
The failure on Linux is apparently some kind of known flaky build issue. I'll just have to re-run it after we fix the linter issues.

iluetkeb:
I'm still around.

> there are linter failures...

The links you provided don't contain errors... Are they the right ones?

> Please try to run the tests (at least the linters) locally before pushing things.

I do, at least before asking for CI. Sometimes I forget, but I think in this case there may be something else going on.

wjwwood:
No, sorry that's yet another bug in Jenkins, sometimes it links you to a different result, here you can check the arm64 linux job directly, under failing tests:

https://ci.ros2.org/job/ci_linux-aarch64/5943/

wjwwood:
> I do, at least before asking for CI. Sometimes I forget, but I think in this case there may be something else going on.

Of course, I just want to remind you in case you weren't. It's not obvious to everyone.

iluetkeb:
So, the rclpy style issues should be addressed now, and I also added returning 0 for received timestamp in rmw_cyclonedds.

ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/7

PR #217, opened 2020-04-21
Service timestamps

iluetkeb:
Proposal to add timestamps on service request and response (for https://github.com/ros2/design/issues/259)

This requires a few related PR's (see below for mentions).

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -861,7 +861,7 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_take_response(
   const rmw_client_t * client,
-  rmw_request_id_t * request_header,
+  rmw_service_info_t * request_header,
   void * ros_response,
   bool * taken);
 
@@ -884,7 +884,7 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_take_request(
   const rmw_service_t * service,
-  rmw_request_id_t * request_header,
+  rmw_service_info_t * request_header,
   void * ros_request,
   bool * taken);
 
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -255,6 +255,14 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
   uint64_t nsec;
 } rmw_time_t;
 
+/// Meta-data for a service-related take.
+typedef struct RMW_PUBLIC_TYPE rmw_service_info_t
+{
+  rmw_time_t source_timestamp;
+  rmw_time_t received_timestamp;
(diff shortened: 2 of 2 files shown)


iluetkeb:
> This looks reasonable, but it will definitely require changes in rclcpp and rclpy.

What kind of changes?

wjwwood:
There is no "nice" way to get this information in C++, we'd need new functions and callback types for rclcpp Clients/Services. It is completely inaccessible in Python atm, AFAICT.

iluetkeb:
> There is no "nice" way to get this information in C++, we'd need new functions and callback types for rclcpp Clients/Services.

The "user", as in a person writing a ROS node, shouldn't need it. AFAICT, the user cannot currently get the message_info_t either, right? At least I could not find a way to create a subscription callback receiving it.

For the executor, the rcl-functions should suffice, shouldn't they?

I mean, I'm happy to add a better way to access this, but currently I don't understand the use-case.

wjwwood:
Without python code to convert the new struct into a pyobj the executor, which is in python, cannot use it either. Either way it needs to be in python so user could use it if they wanted.

iluetkeb:
Got it. That’s in the works already, I’ll push after dinner.

wjwwood:
Sorry you were speaking about c++, still the user should have the option of looking at the time stamps in think. It would require new callback signatures to expose that to them. Also for the case of using it without and executor the user needs all this information to make decisions about when to execute it.

wjwwood:
CI with all rmw's, building and testing `--packages-up-to demo_nodes_cpp demo_nodes_py test_rclcpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10472/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5957/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8540/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10353/)

This includes these pull requests:

- https://github.com/ros2/rmw/pull/217
- https://github.com/ros2/rmw_implementation/pull/96
- https://github.com/ros2/rcl/pull/627
- https://github.com/ros2/rclpy/pull/545
- https://github.com/ros2/rmw_fastrtps/pull/369
- https://github.com/ros2/rmw_connext/pull/412
- https://github.com/ros2/rosidl_typesupport_connext/pull/53
- https://github.com/ros2/rmw_cyclonedds/pull/161

wjwwood:
New CI after linter fixes:

- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/10478/)

ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/7

PR #220, opened 2020-04-22
fix missing target dependency on rosidl_runtime_c

dirk-thomas:
Related to ros2/ros2#904.

Needed for https://github.com/ros2/rmw/blob/21dc0dbf020d7e7aca7a5144465ce10b48af4a55/rmw/include/rmw/rmw.h#L93-L95.

PR #221, opened 2020-04-23
add build dep on rosidl_runtime_c to work with CMake < 3.13

dirk-thomas:
Fixes https://github.com/ros2/rmw/pull/220#discussion_r414063264

Bionic builds:
* Before: [Build Status](https://ci.ros2.org/job/ci_linux/10459/)
* After: [Build Status](https://ci.ros2.org/job/ci_linux/10460/)


rmw/package.xml
```diff
@@ -12,6 +12,8 @@
   <buildtool_depend>ament_cmake_version</buildtool_depend>
 
   <build_depend>rcutils</build_depend>
+  <!-- Only needed because older CMake versions don't support CMP0079 -->
```
clalancette, line 15:
```suggestion
  <!-- Only needed because CMake versions less than 3.13 don't support CMP0079 -->
```

dirk-thomas: Applied in 496f91e606e60a8886a9a23d149142290c35f68b.

PR #204, opened 2020-03-25
Adding doxygen documentation and READMEs to packages

brawner:
Completing the API documentation of rmw. I tried to research as much as I could through rcl and rcutils packages which this package borrows from as well as design.ros2.org documentation.

One question I had for the maintainers was whether items that were strictly defined to a corresponding element in rcutils should have documention, or just link to rcutils documentation? For those, I copied the existing comments, but that does create two locations where that documentation has to be maintained.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

PR #224, opened 2020-04-24
fix linter warning

dirk-thomas:
Fix regression introduced in #204.

@brawner FYI.


brawner:
My bad Dirk. The CI job I ran on that PR did catch the linter error and I missed it. Thanks for fixing it.

PR #227, opened 2020-05-07
Remove MANUAL_BY_NODE liveliness API

ivanpauno:
It was agreed in the discussion [here](https://github.com/ros2/design/pull/250#issuecomment-535161085), that we were going to remove `MANUAL_BY_NODE` related liveliness API.

It would be great to actually do it before `Foxy` release, as it doesn't have more sense after the Participant/Context change.


ivanpauno:
@wjwwood @jacobperron can you review this and the connected PRs?

I will then add a note to Foxy release notes, and update tutorials/design docs.

rotu:
This is not yet deprecated. Shouldn't we be deprecating instead?

https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#deprecation-strategy
https://semver.org/#how-should-i-handle-deprecating-functionality

ivanpauno:
> This is not yet deprecated. Shouldn't we be deprecating instead?
> 
> https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#deprecation-strategy
> https://semver.org/#how-should-i-handle-deprecating-functionality

No, I think it's not going to be deprecated, but directly removed.

The reasons are two:
- It isn't working as it was intended to work.
- It's rarely used, so it won't affect much users.

wjwwood:
I'll defer to @ivanpauno on this, as he has the most context, but my understanding is that the move to one participant per context makes this feature impossible to have working in a way that makes sense. So, it would not satisfy the first conditional of the linked policy, "where possible". But maybe my interpretation is wrong.

wjwwood:
@ivanpauno this is way past the API freeze date though, I think we need approval from several people to make this change at this point...

rotu:
> I'll defer to @ivanpauno on this, as he has the most context, but my understanding is that the move to one participant per context makes this feature impossible to have working in a way that makes sense. So, it would not satisfy the first conditional of the linked policy, "where possible". But maybe my interpretation is wrong.

That seems to be a reason for deprecating it! Deprecate it and stub out the implementations if necessary. But don't introduce binary incompatibility just for the heck of it.

> > This is not yet deprecated. Shouldn't we be deprecating instead?
> > https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#deprecation-strategy
> > https://semver.org/#how-should-i-handle-deprecating-functionality
> 
> No, I think it's not going to be deprecated, but directly removed.
> 
> The reasons are two:
> 
> * It isn't working as it was intended to work.

That's a *great* reason to deprecate and gently redirect users.

> * It's rarely used, so it won't affect much users.

That's not even a consideration of either our developer guidelines nor SemVer

.I looked into this because I got pushback from @clalancette and @sloretz on a completely unused API function https://github.com/ros2/kdl_parser/pull/7#issuecomment-621260007, who both felt it was incorrect to remove a function without first deprecating. We should enforce the policy or update the documented expectations.

ivanpauno:
@rotu The difference with https://github.com/ros2/kdl_parser/pull/7#issuecomment-621260007, is that you can still use the deprecated API, and it will do what's supposed to do.
That's not the case here.

There will be a note in the release notes, so people will have enough information to understand the issue.

@ros2/team for approval

wjwwood:
> > It isn't working as it was intended to work.
> That's a great reason to deprecate and gently redirect users.

Well I can't speak for everyone on our team, that's part of why I said we should include more reviewers, but I see no value in providing an API where the most reliable thing to do is throw "intentionally not implemented". I think it's better to remove it. It's not a "this other version is better" or "we wanted to change the name", it's a "this no longer even works and we can't replicate the old behavior".

rotu:
That's what the message is for - informing the user of why an API is removed and how code should evolve.

C++: `[[deprecated("Broken. Please use XXX instead.")]]`.
C: `RCUTILS_DEPRECATED_WITH_MSG("Broken. Please use XXX instead.")`

Removing a function is disruptive. It breaks the build and blocks direct testing of different versions of the same package. e.g. we couldn't test yesterday's CycloneDDS with tomorrow's rmw_implementation to compare whether a certain commit on a `master` branch broke the nightly.

rmw/include/rmw/types.h
```diff
@@ -394,9 +394,6 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
   /// The signal that establishes a Topic is alive comes from the ROS rmw layer.
   RMW_QOS_POLICY_LIVELINESS_AUTOMATIC,
 
-  /// The signal that establishes a Topic is alive is manually reported by the node
-  RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE,
```
dirk-thomas, line 398: To avoid changing the values of the below enums they could be explicitly set to their previous values.

rotu: Or the value can be deprecated, which would be more self-documenting.

wjwwood:
If the main concern is a message for the user, I'd be fine with having a stub in rmw which is deprecated, but continue with removing anything. The result will be a deprecation warning when using it, and a linker error later since the symbol is not provided.

I do not agree with providing a stub that just asserts or throws at runtime. I think that's just not use for users.

wjwwood:
> Removing a function is disruptive. It breaks the build and blocks direct testing of different versions of the same package. e.g. we couldn't test yesterday's CycloneDDS with tomorrow's rmw_implementation to compare whether a certain commit on a master branch broke the nightly.

You could still do this test, you would just have to include the commit that removes these functions. Btw, if cyclone still implements these functions, but they are removed from rmw there will be no build failure...

The point about the enum's is a good one though, I agree with the suggestion to explicitly set them to avoid changes in the number for now.

rotu:
> You could still do this test, you would just have to include the commit that removes these functions. Btw, if cyclone still implements these functions, but they are removed from rmw there will be no build failure...

Removing the enum breaks the build of `rmw_cyclonedds`. Removing the function definition in `rmw.h` breaks `rcl`. The overall effect of this PR is to require everything to change in lockstep.

The purpose of deprecation in SemVer is for *exactly this reason*. You deprecate the API in the next *minor* release, then adjust downstream consumers, then remove the API in the next *major* release.

If deprecation is not a priority, then we should remove it from the developer guideline and update the quality declaration for this package to state the *actual* versioning policy.

wjwwood:
> Removing the function definition in rmw.h breaks rcl. The overall effect of this PR is to require everything to change in lockstep.

I think that's fine since we're in pre-release. If this were post foxy release, then deprecation would be the only option so that old versions of rcl would continue to work. Before a release we commonly make changes that require synchronized changes between things like rmw and rcl. I don't see any issues with that myself.

> If deprecation is not a priority, then we should remove it from the developer guideline and update the quality declaration for this package to state the actual versioning policy. Let's fix this instead of ignoring it.

We're not ignoring it, in my opinion. Our docs say "where possible". We can expand that to more clearly point out that there is developer discretion in the decision to deprecate or not, but from my perspective it is possible to avoid deprecation with some reasoning behind it. There are cases were it is simply not technically feasible and cases were it just isn't worth it.

Even semver only states *how* you should deprecate (with docs and the like) and that if you deprecate you should at least do the minor version bump. It does not indicate *when* you must deprecate, nor does it advise on how to decide when to do that or not. If you asked on stack overflow I bet you'd get several opinions on this point, but nothing obviously definitive in my opinion.

----

I'm sorry you don't agree with me, but I remain unswayed in my opinion that keeping this functions in a broken or unimplemented state simply adds no value.

I am still open to having the declarations in rmw with deprecation messages and fixing the enums so they don't change values. @ivanpauno would you be willing to do that?

rotu:
> I think that's fine since we're in pre-release. If this were post foxy release, then deprecation would be the only option so that old versions of rcl would continue to work. Before a release we commonly make changes that require synchronized changes between things like rmw and rcl. I don't see any issues with that myself.

ROS2 is in prerelease but rmw_cyclonedds_cpp, rmw_fastrtps, etc. have already released new minor versions. This change would necessarily entail at least a minor version bump for them. SemVer is very particular on this: "Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version."

> > If deprecation is not a priority, then we should remove it from the developer guideline and update the quality declaration for this package to state the actual versioning policy. Let's fix this instead of ignoring it.
> 
> We're not ignoring it, in my opinion. Our docs say "where possible". We can expand that to more clearly point out that there is developer discretion in the decision to deprecate or not, but from my perspective it is possible to avoid deprecation with some reasoning behind it. There are cases were it is simply not technically feasible and cases were it just isn't worth it.

Deprecation *is feasible*. In this case I think it involves a "harder deprecation", e.g. making the function a no-op for one release cycle. I agree that "soft deprecation", i.e. "keep everything working and document that it is going away", is unattainable here.

> Even semver only states _how_ you should deprecate (with docs and the like) and that if you deprecate you should at least do the minor version bump. It does not indicate _when_ you must deprecate, nor does it advise on how to decide when to do that or not. If you asked on stack overflow I bet you'd get several opinions on this point, but nothing obviously definitive in my opinion.

Yes, the SemVer specification on deprecations is all "shoulds" and no "musts". This is so downstream packages like every RMW implementation don't have to do a new release immediately in order to maintain compatibility.

> I'm sorry you don't agree with me, but I remain unswayed in my opinion that keeping this functions in a broken or unimplemented state simply adds no value.

It keeps the build building and allows us to evolve the software rather than "break everything and fix the fallout" that usually accompanies sweeping changes across many repos. Allowing packages to evolve and roll back PRs independently, even in ROS2 "core" packages, is a *massive* value add. 

> I am still open to having the declarations in rmw with deprecation messages and fixing the enums so they don't change values. @ivanpauno would you be willing to do that?

This sounds like exactly what I'm looking for here too. I don't know that we still disagree?

wjwwood:
> ROS2 is in prerelease but rmw_cyclonedds_cpp, rmw_fastrtps, etc. have already released new minor versions. This change would necessarily entail at least a minor version bump for them. SemVer is very particular on this: "Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version."

And what's stopping them from releasing another version?

> This sounds like exactly what I'm looking for here too. I don't know that we still disagree?

Well in my version they would not have implementations, even ones that do nothing. So you'd get a linking error (or possibly a runtime error when dlopening the rmw implementation) after getting a deprecation warning during compile.

rotu:
Maybe instead of thinking about version numbers in terms of what API they change, it's easier to think about them in terms of change propagation:

1. MAJOR changes *require* downstream packages to make changes or break.
2. MINOR changes *allow* downstream packages to make changes.
3. PATCH changes *don't ask anything* of downstream packages.

> > ROS2 is in prerelease but rmw_cyclonedds_cpp, rmw_fastrtps, etc. have already released new minor versions. This change would necessarily entail at least a minor version bump for them. SemVer is very particular on this: "Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version."
> 
> And what's stopping them from releasing another version?

I think it's bad practice, but nothing makes it illegal. In retrospect, maybe these should have all been prerelease versions.

1. For Cyclone DDS, this means the rmw version being out of sync with the eclipse-cyclonedds version.
2. The proposed changes in e.g. https://github.com/ros2/rmw_cyclonedds/pull/178, https://github.com/ros2/rmw_fastrtps/pull/379 are not backwards compatible and would necessitate a MAJOR version bump, not a minor one.

> > This sounds like exactly what I'm looking for here too. I don't know that we still disagree?
> 
> Well in my version they would not have implementations, even ones that do nothing. So you'd get a linking error (or possibly a runtime error when dlopening the rmw implementation) after getting a deprecation warning during compile.

I think you have it backwards - you want to delete the declaration *before* the implementation.

Here's one way to make this change without requiring a synchronized version break:

1. Deprecate in `rmw` by adding a deprecation notice to the declaration and to the enum value. (MINOR change)
2. Stub out the function in `rmw_implementation` and stop loading the function so it's no longer exposed when loading the RMW at runtime. (PATCH change)
3. Remove the declaration in `rmw` so it's no longer exposed when loading the function at compile time. (MAJOR change)
4. Delete the function implementations in each `rmw_...` package. (MAJOR change)
5. Delete the stub in `rmw_implementation` (MAJOR change)

ivanpauno:
> I am still open to having the declarations in rmw with deprecation messages and fixing the enums so they don't change values. @ivanpauno would you be willing to do that?

Sure.

I've already solved the second point (fixing the enums).

I'm not sure of what you mean by the first one.
Do you mean leaving a `rmw_node_assert_liveliness` function that returns `RMW_UNIMPLEMENTED` and is marked with a deprecation attribute?

I can also create a `rmw_node_assert_liveliness` that generates a build error with a message.

wjwwood:
> I think it's bad practice

Why?

> are not backwards compatible and would necessitate a MAJOR version bump, not a minor one.

Why is that an issue?

> I think you have it backwards - you want to delete the declaration before the implementation.

If we were deprecating functionality we wanted to remove in the future, I'd agree. However, the difference here is that the functionality is gone now, no other option. There is no way to preserve the existing behavior, and that point is the reason why we cannot do what you laid out, not because we just don't want the take the steps. There's no opportunity to delete the declaration after the definition, we can either delete both now (original proposal for this pr) or delete the (no longer working) definition and leave the declaration only as a way to inform users why it is no longer there and then remove it later.

The alternative: "replacing the broken definition with a noop, then removing the declaration, then removing the noop (doing a release for each step)" is what I think is not worth the effort. If you want a version with the function available, use the current version, if you want to move forward you have to stop using the function immediately, there is no in between where you get to keep using the function and just have a warning. That's not ideal, but there's no alternative that actually works, as far as I see. If we left the definition with a throw not implemented, it would allow you to compile, but none of your tests for that function would pass anyways. There's no choice but to stop calling, implementing, and/or testing that function right now, so in effect its not deprecated it's already gone.

At any rate, I'm personally done discussing this in circles. My preference is to remove the definition now, leave something behind for users to run into (deprecated declaration or error macro), and do one major version bump release, and I don't think I will be persuaded otherwise. Others can weigh in if they want. Also, we can take what ever steps possible to avoid unnecessary breaks downstream, like the enum fix.

---

> Do you mean leaving a rmw_node_assert_liveliness function that returns RMW_UNIMPLEMENTED and is marked with a deprecation attribute?

No, I mean leaving the declaration in `rmw`, adding a deprecation warning, and deleting it from everywhere else.

> I can also create a rmw_node_assert_liveliness macro that generates a build error with a message.

Is that different than the deprecation warning with no implementation? I guess it's always a compiler error and not a warning? I'm fine with either.

ivanpauno:
> Is that different than the deprecation warning with no implementation? I guess it's always a compiler error and not a warning? I'm fine with either.

Ah, yeah you're right. I will do that.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10673/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6092/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8679/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10572/)

ivanpauno:
@wjwwood I think this is ready.
I've made changes in:
- https://github.com/ros2/rmw/pull/227
- https://github.com/ros2/rmw_cyclonedds/pull/178
- https://github.com/ros2/rclpy/pull/556

rotu:
> > I think it's bad practice
> 
> Why?

> > are not backwards compatible and would necessitate a MAJOR version bump, not a minor one.
> 
> Why is that an issue?

In the end, it just comes down to a personal judgement. I think API compatibility and stability are attractive software features; and that major releases of a package should happen when necessary due to package needs, and not just because an upstream package wants to tie up loose ends. Contrary to a long history of the "ROS way" of big releases spanning many repositories with long periods of instability between releases, I think it's important to decouple development of different, individually versioned and repositoried codebases.

> > I think you have it backwards - you want to delete the declaration before the implementation.
> 
> If we were deprecating functionality we wanted to remove in the future, I'd agree. However, the difference here is that the functionality is gone now, no other option. There is no way to preserve the existing behavior, and that point is the reason why we cannot do what you laid out, not because we just don't want the take the steps. There's no opportunity to delete the declaration after the definition, we can either delete both now (original proposal for this pr) or delete the (no longer working) definition and leave the declaration only as a way to inform users why it is no longer there and then remove it later.
> If you want a version with the function available, use the current version, if you want to move forward you have to stop using the function immediately, there is no in between where you get to keep using the function and just have a warning.

Sorry. I mean the function *declaration* before the function *definition*. You're right that the functionality is already gone. I think API stability and build compatibility are still worth preserving, even if that means gutting a few function implementations to no-ops for a release.

> At any rate, I'm personally done discussing this in circles. My preference is to remove the definition now, leave something behind for users to run into (deprecated declaration or error macro), and do one major version bump release, and I don't think I will be persuaded otherwise. Others can weigh in if they want. Also, we can take what ever steps possible to avoid unnecessary breaks downstream, like the enum fix.

I agree. I think everything has been said and we've both laid out our cases fully. I'm sympathetic to your take. It's very sensible and well-thought out, and nobody want to release broken or ugly code.

jacobperron:
On the topic of bumping the version, I don't think it necessarily has to be a *major* bump as we have not release a 1.0.0 yet. I think a *minor* bump is also acceptable.

From https://semver.org/: 

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

wjwwood:
@jacobperron sure, but the plan is to move this to 1.0.0 for foxy anyways, since as semver recommends:

> How do I know when to release 1.0.0?
>
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

I think some or all of those things are true for this package now.

rmw/include/rmw/types.h
wjwwood: @ivanpauno is it possible to deprecate the enum here in C?

Users can still do `rmw_qos_liveliness_policy_t foo = 2;` right? So the code that take this should have a catch all for "unknown values", so keeping the enum with a deprecation should be safe right?

ros-discourse:

https://discourse.ros.org/t/ros-2-foxy-fitzroy-call-for-testing-and-package-releases/13998/1

ivanpauno:
New CI round:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10692/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6107/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8702/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10590/)

rmw/include/rmw/types.h
wjwwood: @ivanpauno bump on this, not sure anything can be done or needs to be done, but we should resolve this thread one way or the other.

ivanpauno: I've explicitly assigned the enum numbers in https://github.com/ros2/rmw/pull/227/commits/bdea895160edbf3df10a4092944daf0f517276b7.

I can also add a definition of RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE that prints a build warning when used.

ivanpauno: See https://github.com/ros2/rmw/pull/227/commits/f81157128813868c03b13f7726dd9749971d1301.

ivanpauno:
Build up to `rmw_implementation`, test `rmw`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10700/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10598/)

MacOS failed randomly last few times, another try:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8710/)

rmw/include/rmw/types.h
wjwwood: Thanks @ivanpauno

rmw/include/rmw/types.h
```diff
@@ -389,23 +389,41 @@ enum RMW_PUBLIC_TYPE rmw_qos_durability_policy_t
...
+  RMW_QOS_POLICY_LIVELINESS_AUTOMATIC = 1,
+
+#ifdef _WIN32
+  /// Explicitly asserting node liveliness is required in this case.
+  /// This option is deprecated, use RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC if your application
+  /// requires to assert liveliness manually.
+  RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE = 2,
+#else
+  _RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE = 2,
```
rotu, line 403: I think you could do something like:

```
RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE
#ifdef _MSC_VER
#pragma deprecated(RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE)
#else
__attribute__ ((deprecated(RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC_DEPRECATED_MSG)))
#endif
 = 2,
```

Or if you prefer,
```
#ifdef _MSC_VER
#define DECLARE_DEPRECATED(token, msg) token _Pragma(deprecated(token))
#else
#define DECLARE_DEPRECATED(token, msg) token __attribute__((deprecated(msg))
#endif

enum {
   Foo = 1,
   DECLARE_DEPRECATED(Bar, "Use Foo instead") = 2
}
```

ivanpauno: I tried the second one, and it didn't work.
I don't think the first one works either.

rotu: There are typos in the above, but I made and tested a gist. Both approaches work.

https://gist.github.com/rotu/d65b382ef424e97df93d68e7f73d5c0c

ivanpauno: I originally put the deprecation attribute after `= 2` :man_facepalming:.
See https://github.com/ros2/rmw/pull/227/commits/1b188633c970cefd3d06bb98deb799a410650233.

ivanpauno:
Build up to `rmw_implementation` test `rmw`, after updated style in deprecation code:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10710/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6122/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10609/)

Another full macos, because IDK what happened in last one:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8723/)

Issue #226, opened 2020-05-06
Add support for `on sample lost` event


ivanpauno:
#### Feature description

Similar to previously added events like https://github.com/ros2/ros2/issues/822 or `on liveliness lost`, it would be good to add support to `on sample lost` events (which might be really helpful when debugging communication issues).

For `ROS`, `on message lost` might be a more appropriate name.

For DDS based rmw implementations, implementing this should be pretty straightforward:
- Connext: [sample lost status](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Statuses_for_DataReaders.htm#receiving_2076951295_607362), [data reader listener callbacks](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Setting_Up_DataReaderListeners.htm#receiving_2076951295_426659).
- FastRTPS: AFAIS, it's not supported yet: https://github.com/eProsima/Fast-RTPS/blob/8842bdd35bac66b731fe3f2f5e79fe7417355c4b/include/fastdds/rtps/reader/ReaderListener.h#L41.
- CycloneDDS: https://github.com/eclipse-cyclonedds/cyclonedds/blob/623b35d83b9d061187c816f37f4a92798f3b35d1/src/core/ddsc/include/dds/ddsc/dds_public_listener.h#L37.

Detailed list of tasks: https://github.com/ros2/rmw/issues/226#issuecomment-640890428

#### Implementation considerations

For rmw implementations that don't support the feature, they can return [RMW_RET_UNSUPPORTED](https://github.com/ros2/rmw/blob/10bb863176fcd832739eed3077d2fcbe9d619f64/rmw/include/rmw/ret_types.h#L34), as for example fastrtps already does when the creation of an `ON_*_INCOMPATIBLE_QOS` event is requested.

[...]

PR #232, opened 2020-05-28
Add message lost event

ivanpauno:
See https://github.com/ros2/rmw/issues/226.

[First commit](https://github.com/ros2/rmw/commit/79781321003095ce8ac3b17781d1154fdc423cae) reorganizes event statuses headers in a subfolder, so we stop populating the old big `rmw/types.h` (the old headers are still including the new ones, to avoid creating a bunch of PRs together with this one).

[The second commit](https://github.com/ros2/rmw/commit/84e5b29afda22bf4e139b0a1104948c32455ff28) actually implements what the title says.

### DIFF
--- a/rmw/include/rmw/event.h
+++ b/rmw/include/rmw/event.h
@@ -29,13 +29,14 @@ extern "C"
 #include "rmw/ret_types.h"
 #include "rmw/visibility_control.h"
 
-/// Define QoS policy events
+/// Define publisher/subscription events
 typedef enum rmw_event_type_t
 {
   // subscription events
   RMW_EVENT_LIVELINESS_CHANGED,
   RMW_EVENT_REQUESTED_DEADLINE_MISSED,
   RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE,
+  RMW_EVENT_MESSAGE_LOST,
 
   // publisher events
   RMW_EVENT_LIVELINESS_LOST,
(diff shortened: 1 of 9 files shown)


jacobperron:
I'm happy if we distribute the event status definitions into their own header files, but I think it would be best done in it's own PR as it seems completely orthogonal to adding a new event type. At the very least, we should update this PR title to reflect that change and avoid squashing the two changes if we keep them in the same PR.

ivanpauno:
> I'm happy if we distribute the event status definitions into their own header files, but I think it would be best done in it's own PR as it seems completely orthogonal to adding a new event type. At the very least, we should update this PR title to reflect that change and avoid squashing the two changes if we keep them in the same PR.

Yes, I'm not planning to squash both commits.
The header redistribution was to avoid increasing the size of (the already big) `types.h`.
I then added the extra `events_statuses` folder, to avoid having a lots of headers in a flat layout.

ivanpauno:
Last force push edited commit messages without introducing code changes.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11104/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6411/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9061/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11027/)

Issue #234, opened 2020-06-09
test_names_and_types failing on Linux


nuclearsandwich:
## Bug report

**Required Info:**

- Operating System:
  - All platforms
- Installation type:
  - Source (ci.ros2.org)
- Version or commit hash:
  - [vcs export --exact](https://gist.githubusercontent.com/nuclearsandwich/3dbdc7d9adf70bb74c9a9a5a985e9d9e/raw/2b9a945170dd9732334500ff0049452f6a27ab36/1231-exact.repos)
- DDS implementation:
  - N/A
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

```
colcon build --packages-up-to rmw
colcon test --packages-up-to rmw
```

#### Expected behavior

Tests on [test/test_names_and_types.cpp:131](https://github.com/ros2/rmw/blob/b6dcccbb7ec7e1adea6d029868164c5a10910a82/rmw/test/test_names_and_types.cpp#L131) and [test/test_names_and_types.cpp:137](https://github.com/ros2/rmw/blob/b6dcccbb7ec7e1adea6d029868164c5a10910a82/rmw/test/test_names_and_types.cpp#L137) pass.

#### Actual behavior

```

13: [ RUN      ] rmw_names_and_types.rmw_names_and_types_fini
13: /home/jenkins-agent/workspace/nightly_linux-aarch64_debug/ws/src/ros2/rmw/rmw/test/test_names_and_types.cpp:131: Failure
13: Expected equality of these values:
13:   rmw_names_and_types_fini(&names_and_types)
13:     Which is: 0
13:   11
13: /home/jenkins-agent/workspace/nightly_linux-aarch64_debug/ws/src/ros2/rmw/rmw/test/test_names_and_types.cpp:137: Failure
13: Expected equality of these values:
13:   rmw_names_and_types_fini(&names_and_types)
13:     Which is: 0
13:   11
13: [  FAILED  ] rmw_names_and_types.rmw_names_and_types_fini (0 ms)
```

#### Additional information

[...]

PR #235, opened 2020-06-09
Fix rmw_names_and_types_fini test to address issue #234

brawner:
After recent PR in rcutils (https://github.com/ros2/rcutils/pull/247), this test `rmw_names_and_types_init` had some fragile assumptions break, which were reported in issue #234. It also reinitializes names_and_types after it successfully finalizes it the first time, which was another issue.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11085/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6393/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9040/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10995/)

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/rmw/test/test_names_and_types.cpp
+++ b/rmw/test/test_names_and_types.cpp
@@ -108,6 +108,7 @@ TEST(rmw_names_and_types, rmw_names_and_types_init) {
 
   result = rmw_names_and_types_init(&names_and_types, size, &allocator);
   EXPECT_EQ(result, RMW_RET_OK);
+  EXPECT_EQ(rmw_names_and_types_fini(&names_and_types), RMW_RET_OK);
 }
 
 TEST(rmw_names_and_types, rmw_names_and_types_fini) {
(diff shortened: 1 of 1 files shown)


rmw/test/test_names_and_types.cpp
```diff
@@ -118,22 +119,30 @@ TEST(rmw_names_and_types, rmw_names_and_types_fini) {
...
   // Size != 0 and types is null
   auto types_ptr = names_and_types.types;
   names_and_types.types = nullptr;
   EXPECT_EQ(rmw_names_and_types_fini(&names_and_types), RMW_RET_INVALID_ARGUMENT);
   rmw_reset_error();
-  names_and_types.types = types_ptr;
 
   // bad allocator, rcutils fails to finalize string array
   names_and_types.names.allocator.deallocate = nullptr;
+  names_and_types.names.size = 0u;
```
nuclearsandwich, line 139: Why is zeroing the size necessary here?

rmw/test/test_names_and_types.cpp
```diff
@@ -108,6 +108,7 @@ TEST(rmw_names_and_types, rmw_names_and_types_init) {
 
   result = rmw_names_and_types_init(&names_and_types, size, &allocator);
   EXPECT_EQ(result, RMW_RET_OK);
+  EXPECT_EQ(rmw_names_and_types_fini(&names_and_types), RMW_RET_OK);
```
fujitatomoya, line 111: This is the test for rmw_names_and_types_init, not fini. so i guess that if we need we can finalize but not checking the result as test for rmw_names_and_types_init.

rmw/test/test_names_and_types.cpp
```diff
@@ -118,22 +119,30 @@ TEST(rmw_names_and_types, rmw_names_and_types_fini) {
   rmw_ret_t result = rmw_names_and_types_init(&names_and_types, size, &allocator);
   EXPECT_EQ(result, RMW_RET_OK);
 
+  // Ok
+  EXPECT_EQ(rmw_names_and_types_fini(&names_and_types), RMW_RET_OK);
+  result = rmw_names_and_types_init(&names_and_types, size, &allocator);
+  EXPECT_EQ(result, RMW_RET_OK);
```
fujitatomoya, line 125: same here, i would not mix up the test case.

rmw/test/test_names_and_types.cpp
fujitatomoya: okay, understood. I am good to go with it!

PR #243, opened 2020-06-23
Update init/shutdown API documentation.

hidmic:
Constrain behavior for known context states. Alternative to #242.

### DIFF
--- a/rmw/include/rmw/init.h
+++ b/rmw/include/rmw/init.h
@@ -55,10 +55,14 @@ rmw_get_zero_initialized_context(void);
 
 /// Initialize the middleware with the given options, and yielding an context.
 /**
- * The given context must be zero initialized, and is filled with
- * middleware specific data upon success of this function.
+ * Context is filled with middleware specific data upon success of this function.
  * The context is used when initializing some entities like nodes and
- * guard conditions, and is also required to properly call rmw_shutdown().
+ * guard conditions, and is also required to properly call `rmw_shutdown()`.
+ *
+ * \pre The given context must be zero initialized.
+ *
+ * \remarks If context has been already initialized (`rmw_init()` was called on it),
+ *   then `RMW_RET_INVALID_ARGUMENT` is returned.
  *
  * <hr>
  * Attribute          | Adherence
@@ -73,9 +77,9 @@ rmw_get_zero_initialized_context(void);
  * \param[in] options initialization options to be used during initialization
  * \param[out] context resulting context struct
  * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the implementation
  *   identifier does not match, or
- * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are null or invalid, or
  * \return `RMW_RET_ERROR` if an unexpected error occurs.
  */
 RMW_PUBLIC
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/init.h
```diff
@@ -128,8 +136,12 @@ rmw_shutdown(rmw_context_t * context);
  * Lock-Free          | Yes [1]
  * <i>[1] if `atomic_is_lock_free()` returns true for `atomic_uint_least64_t`</i>
  *
- * \return `RMW_RET_OK` if the shutdown was completed successfully, or
+ * This should be defined by the rmw implementation.
```
ivanpauno, line 139: this comment is referring to what?

rmw/include/rmw/init.h
```diff
@@ -55,10 +55,14 @@ rmw_get_zero_initialized_context(void);
...
 /**
- * The given context must be zero initialized, and is filled with
- * middleware specific data upon success of this function.
+ * Context is filled with middleware specific data upon success of this function.
  * The context is used when initializing some entities like nodes and
- * guard conditions, and is also required to properly call rmw_shutdown().
+ * guard conditions, and is also required to properly call `rmw_shutdown()`.
+ *
+ * \pre The given context must be zero initialized.
+ *
```
ivanpauno, line 63: is it possible to ensure that the context will remain zero initialized if `rcl_init` failed?

rmw/include/rmw/init.h
hidmic: It's stating that an RMW implementation has to provide it. Many other functions have it, though we're not entirely consistent about it if you look at the entire API.

rmw/include/rmw/init.h
hidmic: I'll assume you mean `rmw_init()`. If we can provide an adequate return code, we can rollback what we've done and reset the content of the `struct`. Note this talks about the `rmw_context_t` struct and not about any other global state that an `rmw` implementation may initialize as a side effect. We can't make promises there.

ivanpauno: If it doesn't add much burden to the implementation, I would try to always enforce a "strong failure guarantee" (equivalent to the C++ [strong exception safety guarantee](https://en.wikipedia.org/wiki/Exception_safety)), more than only a "basic failure" guarantee).

> I'll assume you mean rmw_init()

Yes, edited.

> Note this talks about the rmw_context_t struct and not about any other global state that an rmw implementation may initialize as a side effect

yes, there might be visible side effects that we can't control.

hidmic: > If it doesn't add much burden to the implementation, I would try to always enforce a "strong failure guarantee"

Yeah, I know where you're going. Because I don't know for sure what is (or will be) out there, I try not to commit to more than it is strictly necessary for it not to be imprecise. In this specific case, I'd rather not make strong guarantees about anything but the struct _just yet_. @wjwwood for feedback.

wjwwood: I don't mind one way or the other. We don't usually reset structures on failure, but we do things like reclaim resources that would otherwise be leaked, which is a little more important in my opinion. The user could always re-zero init the structure if it failed. I think the only important guarantee is that no resources are leaked, so it is not needed to call shutdown or fini on it and it's say to overwrite it.

rmw/include/rmw/init.h
wjwwood: Yeah it would be nice to have this everywhere. I'm fine with adding them as we see them, getting to where we want to be through attrition.

ivanpauno: Oh, I see.

I didn't understand to what "this" was pointing to :joy:.
Maybe, "This function" is more clear.

hidmic:
Alright, going in!

PR #244, opened 2020-06-23
Update init options API documentation.

hidmic:
Clarify pre- and post-conditions. Clean up a bit.

### DIFF
--- a/rmw/include/rmw/init_options.h
+++ b/rmw/include/rmw/init_options.h
@@ -71,7 +71,7 @@ RMW_WARN_UNUSED
 rmw_init_options_t
 rmw_get_zero_initialized_init_options(void);
 
-/// Initialize given init_options with the default values and implementation specific values.
+/// Initialize given init options with the default values and implementation specific values.
 /**
  * The given allocator is used, if required, during setup of the init options,
  * but is also used during initialization.
@@ -79,6 +79,10 @@ rmw_get_zero_initialized_init_options(void);
  * In either case the given allocator is stored in the returned init options.
  *
  * The `impl` pointer should not be changed manually.
+ *
+ * \pre The given init options must be zero initialized.
+ *
+ * \post If initialization fails, init options will remain zero initialized.
  *
  * <hr>
  * Attribute          | Adherence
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/init_options.h
```diff
@@ -108,11 +112,12 @@ rmw_init_options_init(rmw_init_options_t * init_options, rcutils_allocator_t all
  * The allocator from the source is used for any allocations and stored in the
  * destination.
  *
- * The destination should either be zero initialized with
- * `rmw_get_zero_initialized_init_options()` or should have had
- * `rmw_init_options_fini()` called on it.
- * Giving an already initialized init options for the destination will result
- * in a failure with return code `RMW_RET_INVALID_ARGUMENT`.
+ * \pre The destination init options must be zero initialized.
```
ivanpauno, line 115: should we have this requirement?
I think it's ok to make a copy over a no zero initialized init options. We should guarantee that that case doesn't leak and that `*dst` is not modified in the case of failure.

hidmic: > should we have this requirement?

Well, it has to be in a known state. Here I enforce "zero-initialized". We can include "initialized". I guess it shouldn't be much of a burden for most implementations.

Blast545: IMO asking for the dst options to be zero initialized is OK, as it is the expected usage for other `options` over `rcl` and `rmw`

ivanpauno: > IMO asking for the dst options to be zero initialized is OK, as it is the expected usage for other options over rcl and rmw

It's the most common usage we have, and probably the only one.
But I don't think that this precondition is really needed, the implementation will most likely don't change.

hidmic: I will say though that it currently does not provide such feature (i.e. on `master`), so I'd open a separate feature request. I want this to PR to clarify the API contract, not extend it.

hidmic:
@ivanpauno @Blast545 I just constrained the API a bit more e.g. you cannot copy from a zero-initialized source (it's redundant w/ finalization). PTAL.

hidmic:
CI up to `rcl` and `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11262/)

hidmic:
One last sanity CI run after API contract updates:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11281/)

hidmic:
Alright going in!

PR #245, opened 2020-06-24
Add atexit helper

hidmic:
A dumb helper to minimize and simplify cleanup logic in `rmw` implementations.

### DIFF
--- a/rmw/include/rmw/impl/cpp/atexit.hpp
+++ b/rmw/include/rmw/impl/cpp/atexit.hpp
@@ -1,0 +1,63 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


hidmic:
CI up to `rmw`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11292/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6539/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9241/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11196/)

rmw/include/rmw/impl/cpp/atexit.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <utility>
+
+namespace rmw
+{
+namespace impl
+{
+namespace cpp
+{
+
+class atexit final
```
ivanpauno, line 28: I would call this `ScopeExit`. `atexit` sounds like https://www.man7.org/linux/man-pages/man3/atexit.3.html, which is not the same.

wjwwood: Yes, this is definitely a scope exit. I would recommend putting this in rcpputils instead. It could re-use the one in `osrf_testing_tools_cpp`, but it's trivial enough to just copy it around a few times, though that's not ideal.

hidmic:
> Considering that we already have https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/scope_exit.hpp, can we add this to rcpputils and get rid of the duplication?

Aha! I wasn't aware of that one. Ok, I'll improve that one instead. Closing

PR #249, opened 2020-07-07
Update node creation/destruction API documentation.

hidmic:

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -170,7 +170,7 @@ rmw_create_node(
 /**
  * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`
  * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given node handle unchanged.
- * Otherwise, it will proceed despite errors, freeing as much resources as it can, including
+ * Otherwise, it will proceed despite errors, freeing as many resources as it can, including
  * the node handle. Usage of a deallocated node handle is undefined behavior.
  *
  * \pre All publishers, subscribers, services, and clients created from this node must
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -134,7 +134,7 @@ rmw_get_serialization_format(void);
 /**
  * This function can fail, and therefore return `NULL`, if:
  *   - name is not a valid non-null node name
- *   - namespace is not a valid non-null namespace
+ *   - namespace_ is not a valid non-null namespace
  *   - context is not valid i.e. it is zero-initialized, or
  *     its implementation identifier does not match that of
  *     this API implementation, or has been invalidated by
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -133,23 +133,14 @@ rmw_get_serialization_format(void);
 /// Create a node and return a handle to that node.
 /**
  * This function can fail, and therefore return `NULL`, if:
- *   - context, name, namespace_, or security_options is `NULL`
- *   - context, security_options is invalid
+ *   - context is not valid i.e. it has been initialized
+ *     by `rmw_init()` but not yet invalidated by
+ *     `rmw_shutdown()`
  *   - memory allocation fails during node creation
  *   - an unspecified error occurs
  *
- * The context must be non-null and valid, i.e. it has been initialized
- * by `rmw_init()` and has not been finalized by `rmw_shutdown()`.
- *
- * The name and namespace_ should be valid node name and namespace,
- * and this should be asserted by the caller (e.g. `rcl`).
- *
- * The domain ID should be used to physically separate nodes at the
- * communication graph level by the middleware.
- * For RTPS/DDS this maps naturally to their concept of domain id.
- *
- * The security options should always be non-null and encapsulate the
- * essential security configurations for the node and its entities.
+ * \pre The given name must be a valid non-null node name.
+ * \pre The given namespace_ must be a valid non-null node namespace.
  *
  * <hr>
  * Attribute          | Adherence
@@ -166,10 +157,9 @@ rmw_get_serialization_format(void);
  * \param[in] name the node name
  * \param[in] namespace_ the node namespace
  * \param[in] domain_id the id of the domain that the node should join
- * \param[in] security_options the security configurations for the node
- * \param[in] localhost_only whenever to use loopback only for communication or default
- * network interfaces.
- * \return rmw node handle or `NULL` if there was an error
+ * \param[in] localhost_only whenever to use loopback only for communication or
+ *   default network interfaces.
+ * \return rmw node handle, or `NULL` if there was an error
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
@@ -181,7 +171,7 @@ rmw_create_node(
 
 /// Finalize a given node handle, reclaim the resources, and deallocate the node handle.
 /**
- * The method may assume - but should verify - that all publishers, subscribers,
+ * The function may assume - but should verify - that all publishers, subscribers,
  * services, and clients created from this node have already been destroyed.
  * If the rmw implementation chooses to verify instead of assume, it should
  * return `RMW_RET_ERROR` and set a human readable error message if any entity
@@ -189,7 +179,9 @@ rmw_create_node(
  *
  * \param[in] node the node handle to be destroyed
  * \return `RMW_RET_OK` if successful, or
- * \return `RMW_RET_INVALID_ARGUMENT` if node is null, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if node is invalid, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the implementation
+ *   identifier does not match, or
  * \return `RMW_RET_ERROR` if an unexpected error occurs.
  */
 RMW_PUBLIC
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -133,14 +133,13 @@ rmw_get_serialization_format(void);
 /// Create a node and return a handle to that node.
 /**
  * This function can fail, and therefore return `NULL`, if:
+ *   - name is not a valid non-null node name
+ *   - namespace is not a valid non-null namespace
  *   - context is not valid i.e. it has been initialized
  *     by `rmw_init()` but not yet invalidated by
  *     `rmw_shutdown()`
  *   - memory allocation fails during node creation
  *   - an unspecified error occurs
- *
- * \pre The given name must be a valid non-null node name.
- * \pre The given namespace_ must be a valid non-null node namespace.
  *
  * <hr>
  * Attribute          | Adherence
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -133,23 +133,14 @@ rmw_get_serialization_format(void);
...
- * and this should be asserted by the caller (e.g. `rcl`).
- *
- * The domain ID should be used to physically separate nodes at the
- * communication graph level by the middleware.
- * For RTPS/DDS this maps naturally to their concept of domain id.
- *
- * The security options should always be non-null and encapsulate the
- * essential security configurations for the node and its entities.
+ * \pre The given name must be a valid non-null node name.
+ * \pre The given namespace_ must be a valid non-null node namespace.
```
hidmic, line 143: API here departs from the rest. Usually, a `null` arguments means `RMW_RET_INVALID_ARGUMENT` or similar.`rmw_create_node` would do the same, and then require the caller to ensure `name` and `namespace_` to be valid, precluding `null` arguments entirely. 

I chose to keep the precondition. But we could check instead.

ivanpauno: The problem of requiring preconditions that aren't checked at runtime and that the compiler doesn't check is that in the case they aren't satisfied it will probably generate a segfault (in no-release mode, that can be avoided using asserts).
If it's cheap to check, I don't see why not doing that.

rmw/include/rmw/rmw.h
```diff
@@ -183,15 +173,17 @@ rmw_create_node(
 
 /// Finalize a given node handle, reclaim the resources, and deallocate the node handle.
 /**
- * The method may assume - but should verify - that all publishers, subscribers,
+ * The function may assume - but should verify - that all publishers, subscribers,
```
ivanpauno, line 176: I would really delete the `should verify` comment.
We don't verify that.
Verifying that isn't trivial nor cheap.

hidmic: Agreed. Though it doesn't make any promises.

hidmic: Also, this should be documented as a precondition.

rmw/include/rmw/rmw.h
hidmic: Asserts would indeed be required. That still means no checks in production. 

> If it's cheap to check, I don't see why not doing that.

I can't tell what the cost of validation is. 

In this particular case, I don't mind either way as long as it's properly documented.

ivanpauno: > I can't tell what the cost of validation is.

Yeah, but it does sound like something we can do easily.

> In this particular case, I don't mind either way as long as it's properly documented.

I think it was properly documented:

```
*   - context, name, namespace_, or security_options is `NULL`
```

hidmic: Correct, the odd thing being that nullity would be checked but validity would be assumed. Here I went for complete assumption. We can go for complete checking.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/249/commits/6c83764cab3e6a5d4fe1dcc0a9124d369e14e156.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/249/commits/6c83764cab3e6a5d4fe1dcc0a9124d369e14e156 for the alternative.

rmw/include/rmw/rmw.h
```diff
@@ -133,24 +133,14 @@ rmw_get_serialization_format(void);
 /// Create a node and return a handle to that node.
 /**
  * This function can fail, and therefore return `NULL`, if:
- *   - context, name, namespace_, or security_options is `NULL`
- *   - context, security_options is invalid
+ *   - name is not a valid non-null node name
+ *   - namespace is not a valid non-null namespace
+ *   - context is not valid i.e. it has been initialized
+ *     by `rmw_init()` but not yet invalidated by
+ *     `rmw_shutdown()`
```
clalancette, line 140: I'm not sure I understand this sentence.  Shouldn't it be more like:

```
 *   - context is not valid i.e. it has not yet been initialized
 *     by `rmw_init()` or has been invalidated by
 *     `rmw_shutdown()`
```

If not, can you clarify further?

rmw/include/rmw/rmw.h
```diff
@@ -166,10 +156,9 @@ rmw_get_serialization_format(void);
  * \param[in] name the node name
  * \param[in] namespace_ the node namespace
```
clalancette, line 157: I know you didn't change this, but I find it odd that `namespace_` has a trailing underscore.  What do you think about changing that (and updating the documentation to match)?

rmw/include/rmw/rmw.h
```diff
@@ -183,15 +172,16 @@ rmw_create_node(
...
 /**
- * The method may assume - but should verify - that all publishers, subscribers,
- * services, and clients created from this node have already been destroyed.
- * If the rmw implementation chooses to verify instead of assume, it should
- * return `RMW_RET_ERROR` and set a human readable error message if any entity
- * created from this node has not yet been destroyed.
+ * \pre All publishers, subscribers, services, and clients created from this node must
+ *   have been destroyed upon call. Some rmw implementations may choose to verify this,
+ *   returning `RMW_RET_ERROR` and setting a human readable error message if any entity
+ *   created from this node has not yet been destroyed.
```
clalancette, line 178:
```suggestion
 * \pre All publishers, subscribers, services, and clients created from this node must
 *   have been destroyed prior to this call. Some rmw implementations may verify this,
 *   returning `RMW_RET_ERROR` and setting a human readable error message if any entity
 *   created from this node has not yet been destroyed.  However, this is not guaranteed and so
 *   callers should ensure that this is the case before calling this function.
```

hidmic: See https://github.com/ros2/rmw/pull/249/commits/42bb43bc5711eec1dcb3ec177d22875a9f067bcc.

rmw/include/rmw/rmw.h
hidmic: True. See https://github.com/ros2/rmw/pull/249/commits/42bb43bc5711eec1dcb3ec177d22875a9f067bcc.

rmw/include/rmw/rmw.h
hidmic: Yes! That's what I meant, though _not initialized_ is a slipper slope. We can only safely guard against mismatching implementation identifiers or zero-initialized contexts.

rmw/include/rmw/rmw.h
hidmic: Spoke too soon, https://github.com/ros2/rmw/pull/249/commits/244205136b1317f8b2cee7c0a36541e222b0c0ff.

Even though it's an argument name and it is within an external C linkage block, the compiler still barfs at `namespace`.

rmw/include/rmw/rmw.h
ivanpauno: > Correct, the odd thing being that nullity would be checked but validity would be assumed. Here I went for complete assumption. We can go for complete checking. It's the middle ground that I find odd.

I didn't get that part before.
I imagine that for a valid name and namespace you're referring to:
- https://github.com/ros2/rmw/blob/master/rmw/include/rmw/validate_node_name.h
- https://github.com/ros2/rmw/blob/master/rmw/include/rmw/validate_namespace.h

I think we're using that on `rcl` but never on the rmw implementations (not sure why the definitions are in rmw, maybe we wanted to use the functions from the rmw implementations (?)).
Repeating the validity check would be probably unnecessary, as `rcl` already takes care of that.

> It's the middle ground that I find odd.

Yeah, it's a bit odd to check "not NULL but maybe invalid".
I'm comfortable with the middle ground though.

rmw/include/rmw/rmw.h
ivanpauno: In `rcl` we are using `node_namespace` to avoid that problem, which looks a bit nicer that the trailing underscore.
That change sounds a bit unrelated to the PR though.

hidmic: It's slightly redundant though.

rmw/include/rmw/rmw.h
hidmic: So, I'm pushing for a choice because IMO the middle ground makes no sense. If the caller must ensure that names are valid, `null` names already fails validation. If you get a `null`  name, that means the caller broke the precondition and you're in UB land. Checking for `null` doesn't make things any better.

If no one strongly disagrees, I'll keep the checks as it's less awkward.

hidmic:
One thing I changed in RMW implementations but didn't reflect here yet. What should memory be allocated? `rmw_context_t` comes with an `rcutils_allocator_t` in `options` so I went along with it naturally in some places. But there's also `rmw_allocate()` and friends, and C++ `new` / `delete` operators. It's a bit of a mess.

clalancette:
> `rmw_context_t` comes with an `rcutils_allocator_t` in `options` (which I went along with it naturally in some places). But there's also `rmw_allocate()` and friends, and C++ `new` / `delete` operators. I don't mind rolling back some changes, but I think having three different allocators that can be used interchangeably is a mess. CC @wjwwood .

I agree that this is kind of messy.  But it seems orthogonal to this PR; should we go ahead and merge this one, and open an issue for further discussion of this topic?

hidmic:
> But it seems orthogonal to this PR; should we go ahead and merge this one, and open an issue for further discussion of this topic?

Yeah, I guess. I do think that the way memory is allocated should be consistent and documented though.

wjwwood:
I would agree with merge then fix in a follow up, just because there's so much inconsistency already.

I think the "right way" to do this is to always use the passed rcutils allocator, or pass one if there's not one being passed atm. We should deprecate the `rmw_*` allocator functions, and we should convert the rcutils allocator into a C++ allocator if we must use new/delete, otherwise placement new can be used in combination with the rcutils allocator.

hidmic:
Rebased to reflect #248 API changes.

hidmic:
Running CI up to `test_rmw_implementation` and all RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11444/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6659/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9388/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11347/)

rmw/include/rmw/rmw.h
```diff
@@ -181,15 +168,22 @@ rmw_create_node(
...
 /// Finalize a given node handle, reclaim the resources, and deallocate the node handle.
 /**
- * The method may assume - but should verify - that all publishers, subscribers,
- * services, and clients created from this node have already been destroyed.
- * If the rmw implementation chooses to verify instead of assume, it should
- * return `RMW_RET_ERROR` and set a human readable error message if any entity
- * created from this node has not yet been destroyed.
+ * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`
+ * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given node handle unchanged.
+ * Otherwise, it will proceed despite errors, freeing as much resources as it can, including
```
clalancette, line 173:
```suggestion
 * Otherwise, it will proceed despite errors, freeing as many resources as it can, including
```

hidmic: Done in https://github.com/ros2/rmw/pull/249/commits/79648b3bb84aad31fa3a2f134458c1c88b0d3ed3.

hidmic:
Alright, all green. Merging.

PR #251, opened 2020-07-17
Add actual domain id to rmw_context_t

ivanpauno:
See https://github.com/ros2/rcl/pull/718.

### DIFF
--- a/rmw/include/rmw/init.h
+++ b/rmw/include/rmw/init.h
@@ -42,6 +42,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_context_t
   const char * implementation_identifier;
   /// Options used to initialize the context.
   rmw_init_options_t options;
+  /// Domain id that is being used.
+  size_t actual_domain_id;
   /// Implementation defined context information.
   /** May be NULL if there is no implementation defined context information. */
   rmw_context_impl_t * impl;
--- a/rmw/include/rmw/init.h
+++ b/rmw/include/rmw/init.h
@@ -67,6 +67,9 @@ rmw_get_zero_initialized_context(void);
  * \pre The given context must be zero initialized.
  *
  * \post If initialization fails, context will remain zero initialized.
+ * \post `context->actual_domain_id` will be set with the domain id the rmw implementation is using.
+ *  This matches `options->domain_id` if it is not RMW_DEFAULT_DOMAIN_ID.
+ *  In other case, the value is rmw implementation dependent.
  *
  * \remarks If options are zero-initialized, then `RMW_RET_INVALID_ARGUMENT` is returned.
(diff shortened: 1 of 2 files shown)


rmw/src/init.c
```diff
@@ -26,6 +28,9 @@ rmw_get_zero_initialized_context(void)
 {
   return (const rmw_context_t) {
            .instance_id = 0,
+           .implementation_identifier = NULL,
+           .options = rmw_get_zero_initialized_init_options(),
+           .actual_domain_id = RMW_DEFAULT_DOMAIN_ID,
```
hidmic, line 33: @ivanpauno hmm, this is not zero-initialization. And I don't think it's strictly worse to set it to `0`, as the `actual_domain_id` field value is completely meaningless for an uninitialized context.

rmw/include/rmw/init.h
```diff
@@ -42,6 +42,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_context_t
   const char * implementation_identifier;
   /// Options used to initialize the context.
   rmw_init_options_t options;
+  /// Domain id that is being used.
+  size_t actual_domain_id;
```
hidmic, line 46: @ivanpauno consider documenting this as a postcondition of `rmw_init()`.

ivanpauno: Done in https://github.com/ros2/rmw/pull/251/commits/3956c762373ab42d22d4c57937db9125c6272b27.

rmw/src/init.c
ivanpauno: Done in https://github.com/ros2/rmw/pull/251/commits/3956c762373ab42d22d4c57937db9125c6272b27.

> hmm, this is not zero-initialization

That isn't the interpretation of `*_get_zero_initialized` functions that I have.
IMO, the point is to have a non-fail function to initialize the structure to a valid state.
If we would always set everything to zero, I don't see the point of having these functions.
It would be easier to maintain if we would simply do:

```c
memset(&struct_to_be_initialized, 0, sizeof(package_struct_definition_t));
```

It would likely be have better performance too.

hidmic: > the point is to have a non-fail function to initialize the structure to a valid state.

I'd rather say it allows to initialize the structure in a known _invalid_ or not initialized state (of which there're many). The choice of `0` is somewhat arbitrary I guess, and while you're right about `memset`, having a function makes the code more readable :).

ivanpauno: > known invalid or not initialized state

Yeah, that's what I tried to mean.

wjwwood: Actually zero is intentional, and `0` would make more sense here, but it doesn't really matter. The point is to put the structure in a known state that the initialization function can recognize. If the structures were larger I think memset might make sense, but for these sizes, the compiler will lilkely optimize this step heavily. Maybe more so if all the fields are `0`...

PR #252, opened 2020-07-20
Update publisher creation/destruction API documentation.

hidmic:

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -296,8 +296,8 @@ rmw_get_default_publisher_options(void);
  * ------------------ | -------------
  * Allocates Memory   | Yes
  * Thread-Safe        | No
- * Uses Atomics       | No [1]
- * Lock-Free          | No [1]
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
  * <i>[1] rmw implementation defined, check the implementation documentation</i>
  *
  * \param[in] node Handle to node with which to register this publisher
@@ -333,8 +333,8 @@ rmw_create_publisher(
  * ------------------ | -------------
  * Allocates Memory   | No
  * Thread-Safe        | No
- * Uses Atomics       | No [1]
- * Lock-Free          | No [1]
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
  * <i>[1] rmw implementation defined, check the implementation documentation</i>
  *
  * \param[in] node Handle to node with which the given publisher is registered
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -302,7 +302,9 @@ rmw_get_default_publisher_options(void);
  *
  * \param[in] node Handle to node with which to register this publisher
  * \param[in] type_support Type support for the messages to be published
- * \param[in] topic_name Fully qualified name of the topic to publish to
+ * \param[in] topic_name Name of the topic to publish to, often a fully qualified
+ *   topic name unless `qos_profile` is configured to avoid ROS namespace conventions
+ *   i.e. to create a native topic publisher
  * \param[in] qos_profile QoS policies for this publisher
  * \param[in] publisher_options Options to configure this publisher
  * \return rmw publisher handle, or `NULL` if there was an error
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -276,14 +276,36 @@ RMW_WARN_UNUSED
 rmw_publisher_options_t
 rmw_get_default_publisher_options(void);
 
-/// Create and return an rmw publisher.
+/// Create a publisher and return a handle to that publisher.
 /**
- * \TODO(wjwwood): add detailed documentation, adding a not about one of the
- *   arguments for now.
+ * This function can fail, and therefore return `NULL`, if:
+ *   - node is not a valid non-null handle for this rmw implementation,
+ *     as returned by `rmw_create_node()`, or it is associated to a shutdown context
+ *   - type_support is a valid non-null message type support, as returned by
+ *     `ROSIDL_GET_MSG_TYPE_SUPPORT()`
+ *   - topic_name is not a valid non-null topic name, according to
+ *     `rmw_validate_full_topic_name()`
+ *   - qos_profile is a fully specified non-null profile i.e. no UNKNOWN policies
+ *   - publisher_options is not a valid non-null option set, as returned by
+ *     `rmw_get_default_publisher_options()`
+ *   - memory allocation fails during publisher creation
+ *   - an unspecified error occurs
  *
- * The argument `publisher_options` must not be nullptr.
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No [1]
+ * Lock-Free          | No [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
  *
- * \param[in] publisher_options options for configuring the publisher
+ * \param[in] node Handle to node with which to register this publisher
+ * \param[in] type_support Type support for the messages to be published
+ * \param[in] topic_name Name of the topic to publish to
+ * \param[in] qos_profile QoS policies for this publisher
+ * \param[in] publisher_options Options to configure this publisher
+ * \return rmw publisher handle, or `NULL` if there was an error
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -292,9 +314,35 @@ rmw_create_publisher(
...
   const rmw_publisher_options_t * publisher_options);
 
+/// Finalize a given publisher handle, reclaim the resources, and deallocate the publisher handle.
+/**
+ * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`
+ * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given publisher handle unchanged.
+ * Otherwise, it will proceed despite errors, freeing as many resources as it can, including
+ * the publisher handle. Usage of a deallocated publisher handle is undefined behavior.
+ *
+ * \pre Given node must be the one the publisher was registered with.
```
hidmic, line 327: @hidmic this makes me wonder, why don't we keep a reference to the node in the publisher?

rmw/include/rmw/rmw.h
```diff
@@ -276,14 +276,36 @@ RMW_WARN_UNUSED
...
 rmw_get_default_publisher_options(void);
 
-/// Create and return an rmw publisher.
+/// Create a publisher and return a handle to that publisher.
 /**
- * \TODO(wjwwood): add detailed documentation, adding a not about one of the
- *   arguments for now.
+ * This function can fail, and therefore return `NULL`, if:
+ *   - node is not a valid non-null handle for this rmw implementation,
+ *     as returned by `rmw_create_node()`, or it is associated to a shutdown context
```
hidmic, line 283: Behavior on a shutdown context was not specified. I think this makes sense.

ivanpauno: I'm not sure if it makes sense.

One of the possible paths to fix https://github.com/ros2/rclcpp/issues/1139 is to reduce the side-effects that calling `shutdown` has, and this will go against that.

rmw/include/rmw/rmw.h
```diff
@@ -276,14 +276,36 @@ RMW_WARN_UNUSED
...
+ *   - type_support is a valid non-null message type support, as returned by
```
ivanpauno, line 284:
```suggestion
 *   - type_support is not a valid non-null message type support, as returned by
```
?

rmw/include/rmw/rmw.h
```diff
@@ -276,14 +276,36 @@ RMW_WARN_UNUSED
...
+ *     `ROSIDL_GET_MSG_TYPE_SUPPORT()`
+ *   - topic_name is not a valid non-null topic name, according to
+ *     `rmw_validate_full_topic_name()`
+ *   - qos_profile is a fully specified non-null profile i.e. no UNKNOWN policies
```
ivanpauno, line 288:
```suggestion
 *   - qos_profile is not a fully specified non-null profile i.e. no UNKNOWN policies
```

rmw/include/rmw/rmw.h
```diff
@@ -276,14 +276,36 @@ RMW_WARN_UNUSED
...
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No [1]
+ * Lock-Free          | No [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
  *
- * \param[in] publisher_options options for configuring the publisher
+ * \param[in] node Handle to node with which to register this publisher
+ * \param[in] type_support Type support for the messages to be published
+ * \param[in] topic_name Name of the topic to publish to
```
ivanpauno, line 305:
```suggestion
 * \param[in] topic_name Fully qualified name of the topic to publish to
```

rmw/include/rmw/rmw.h
ivanpauno: I understand what you mean, but I think the scope of that suggestion is a little broader than the goal of this PR.

rmw/include/rmw/rmw.h
```diff
@@ -292,9 +314,35 @@ rmw_create_publisher(
...
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No [1]
+ * Lock-Free          | No [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
+ * \param[in] node Handle to node with which the given publisher is registered
+ * \param[in] publisher Handle to publisher to be finalized
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if node or publisher is null, or
```
ivanpauno, line 341:
```suggestion
 * \return `RMW_RET_INVALID_ARGUMENT` if node or publisher is `NULL`, or
```

rmw/include/rmw/rmw.h
hidmic: Yeap, not planning to address it here. Just wondering why we seemingly introduce a point of failure for no apparent reason.

rmw/include/rmw/rmw.h
hidmic: Oof, that's a long winded discussion. Short term, I don't mind as long as behavior is specified. 

Skimming through that thread though, [this comment](https://github.com/ros2/rclcpp/issues/1139#issuecomment-635001222) seems to suggest this is the current expected behavior, nevermind future changes.

ivanpauno: > Skimming through that thread though, this comment seems to suggest this is the current expected behavior, nevermind future changes.

I think that we have settled in that PR that we want to avoid that behavior in the future.

Currently, that behavior is only enforced in `rcl` in "some parts" of the API (not everywhere, some things continue working after shutdown).
IMO, we should avoid introducing similar checks in `rcl`, it will make change to happen much more difficult in the future.

rmw/include/rmw/rmw.h
hidmic: Done in https://github.com/ros2/rmw/pull/252/commits/5f8b5671760c2091c30de3677f4546f279cd8c0c.

rmw/include/rmw/rmw.h
hidmic: Done in https://github.com/ros2/rmw/pull/252/commits/5f8b5671760c2091c30de3677f4546f279cd8c0c.

rmw/include/rmw/rmw.h
hidmic: Done in https://github.com/ros2/rmw/pull/252/commits/5f8b5671760c2091c30de3677f4546f279cd8c0c.

rmw/include/rmw/rmw.h
hidmic: Yes! Done in https://github.com/ros2/rmw/pull/252/commits/5f8b5671760c2091c30de3677f4546f279cd8c0c.

rmw/include/rmw/rmw.h
hidmic: We can settle over there, as I disagree with some of the arguments.

ivanpauno: Sure, but this PR shouldn't be blocked waiting on that decision.
Currently, none of the rmw implementations are checking if `shutdown` has already been called.

hidmic: Fair point. I do want to speed things up.

rmw/include/rmw/rmw.h
```diff
@@ -276,14 +276,36 @@ RMW_WARN_UNUSED
...
+ *   - an unspecified error occurs
  *
- * The argument `publisher_options` must not be nullptr.
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No [1]
+ * Lock-Free          | No [1]
```
ivanpauno, line 300: I would really document this as `Undefined [1]` instead of `No [1]`.
It's true we're using the later everywhere else.

hidmic: Yeah, that'd be better than a qualified `No`. @clalancette @wjwwood thoughts?

rmw/include/rmw/rmw.h
```diff
@@ -302,7 +302,9 @@ rmw_get_default_publisher_options(void);
  *
  * \param[in] node Handle to node with which to register this publisher
  * \param[in] type_support Type support for the messages to be published
- * \param[in] topic_name Fully qualified name of the topic to publish to
+ * \param[in] topic_name Name of the topic to publish to, often a fully qualified
+ *   topic name unless `qos_profile` is configured to avoid ROS namespace conventions
```
ivanpauno, line 306: In that case, it would still be the fully qualified name.

The only difference is that when "avoid ROS namespace conventions" is used our dds mangling rules aren't applied.
"fully qualified" refers to "expansion and remapping has already been done", which is still true when "avoid ROS namespaces conventions" has been set.

hidmic: Fully qualified ROS names are also required to start with a slash (`/`), see [_Topic and Service name mapping to DDS_ design document](http://design.ros2.org/articles/topic_and_service_names.html).

rmw/include/rmw/rmw.h
wjwwood: I think we've used `Maybe [N]` elsewhere.

rmw/include/rmw/rmw.h
wjwwood: I'm not sure, but I think it might encourage the caller to make sure they keep the node around (as you need it to call fini on the publisher, so it's obvious it needs to be kept around), but I don't actually remember the logic behind it. Obviously if they just follow the docs, then this would not be a useful argument.

hidmic: Thanks for the insight. We can revisit at a later point in time. I wasn't planning to change it here anyways.

hidmic:
CI up `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11647/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6777)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9528/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11623/)

Issue #210, opened 2020-04-03
QoS duration APIs need INFINITE constant and an API contract to enforce proper usage by implementations


emersonknapp:
#### Feature description
Right now, when requesting endpoint info from rmw implementations, e.g. `rmw_get_publishers_info_by_topic`, we receive inconsistent results for publishers created with the same QoS settings. This makes it hard to create subscriptions adaptively when using queried QoS profiles. It also means that the reporting is noisy/lossy, we do not know what settings were really given to the endpoints.

* Define a set of constants in the RMW API to define "infinite duration"
* State the contract on the RMW APIs `rmw_get_publishers_info_by_topic` and `rmw_get_subscriptions_info_by_topic` that these infinity values will be reported by RMW implementations
* State the contract on subscription/publisher creation APIs that an input of Zero duration for QoS policies will be interpreted and reported as Infinity.

The DDS Spec's OMG IDL PSM (Platform Specific Model) defines the following (page 139)

```
struct Duration_t {
long sec;
unsigned long nanosec;
};
const long DURATION_INFINITE_SEC = 0x7fffffff;
const unsigned long DURATION_INFINITE_NSEC = 0x7fffffff;

const long DURATION_ZERO_SEC = 0;
const unsigned long DURATION_ZERO_NSEC = 0;
```

We should do something similar. However, since we use a `uint64_t` for both fields of `rmw_time_t`, I'd suggest the infinity value being "uint64_max" for both seconds and nanoseconds.

# Work Log

* [ ] #301
* [ ] https://github.com/ros2/rmw_cyclonedds/pull/288 Change rmw_cyclonedds to conform to API
* [ ] Change rmw_fastrtps to conform to API
* [ ] Change rmw_connext to conform to API

fujitatomoya:
> I am uncomfortable with this contract.

so am I. and I believe we should not have redundant definition as infinity. This could be complicated.

emersonknapp:
Yes, based on these conversations I am thinking to revise the proposal to something like the following:

```
Define a constant RMW_TIME_INFINITY.

A Publisher offering a {0, 0} rmw_time_t for a Duration-based QoS policy is interpreted as "not specified" and be reported as RMW_TIME_INFINITY in endpoint queries.

A Subscription requesting a {0, 0} rmw_time_t for a Duration-based QoS policy is interpreted as "not specified" and be reported as RMW_TIME_INFINITY in endpoint queries. This request should be compatible with all offers.
```

[...]

PR #255, opened 2020-07-31
RMW_DURATION_INFINITE constant - use in QoS API contracts

### LINKED ISSUE #215: rmw_time_t is
`rmw_time_t` is defined as:

```
{
  uint64_t sec;
  uint64_t nsec;
} rmw_time_t;
```

This differs from `rcutils_duration_value_t` which admits a value as *just* a `uint64_t` holding nanoseconds. It probably makes sense to replace `rmw_time_t` with a type alias for `rcutils_duration_value_t`, and to make clear that it is a *duration*, not a *time point*.

emersonknapp:
Resolves https://github.com/ros2/rmw/issues/210
Depends on #298 (which resolves #215)

Create explicit `RMW_DURATION_INFINITE` constant for use in requesting/offering QoS policies with durations, to be used as the default value instead of 0. Enforce using these values in the API contract.

Linked PRs:
* [ ] https://github.com/ros2/rmw_fastrtps/pull/512
* [ ] https://github.com/ros2/rmw_cyclonedds/pull/287
* [ ] https://github.com/ros2/rmw_connext/pull/487

Note: non-linked followups:
* [ ] ros2cli (print "Infinite" instead of constant literal value)
* [ ] rosbag2 (interpret old implementation-specific "infinite" values and rewrite them as RMW_DURATION_INFINITE, for playback)

### Testing

Gist: 
* src: https://gist.github.com/emersonknapp/5f998a7e585c8a7bdc1f4845a67bd27a
* raw: https://gist.githubusercontent.com/emersonknapp/5f998a7e585c8a7bdc1f4845a67bd27a/raw/d743e24087109283bb9f6f3ae21f1feae4436b80/ros2.repos

Build:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13711/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8585/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11424/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13785/)

### DIFF
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -339,6 +339,9 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
 /// A duration of time, measured in nanoseconds.
 typedef rcutils_duration_value_t rmw_duration_t;
 
+#define RMW_DURATION_INFINITE INT64_MAX
+
+
 typedef rcutils_time_point_value_t rmw_time_point_value_t;
 
 /// Meta-data for a service-related take.
@@ -430,14 +433,14 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
   RMW_QOS_POLICY_LIVELINESS_UNKNOWN = 4
 };
 
-/// QoS Deadline default, 0s indicates deadline policies are not tracked or enforced
-#define RMW_QOS_DEADLINE_DEFAULT 0
+/// QoS Deadline default, indicates that no deadline is enforced
+#define RMW_QOS_DEADLINE_DEFAULT RMW_DURATION_INFINITE
 
-/// QoS Lifespan default, 0s indicate lifespan policies are not tracked or enforced
-#define RMW_QOS_LIFESPAN_DEFAULT 0
+/// QoS Lifespan default, indicates that no lifespan is enforced
+#define RMW_QOS_LIFESPAN_DEFAULT RMW_DURATION_INFINITE
 
-/// QoS Liveliness lease duration default, 0s indicate lease durations are not tracked or enforced
-#define RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT 0
+/// QoS Liveliness lease duration default, indicates that no lease duration is enforced
+#define RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT RMW_DURATION_INFINITE
 
 /// ROS MiddleWare quality of service profile.
 typedef struct RMW_PUBLIC_TYPE rmw_qos_profile_t
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -333,7 +333,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
   /// Nanoseconds component of this time point
   uint64_t nsec;
 } RMW_DECLARE_DEPRECATED (rmw_time_t,
-  "rmw_time_t has been deprecated in G-Turtle in favor of rmw_duration_t. "
+  "rmw_time_t has been deprecated in Galactic in favor of rmw_duration_t. "
   "It will be removed in H-Turtle");
 
 /// A duration of time, measured in nanoseconds.
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -39,6 +39,12 @@ extern "C"
 // implementation. It may need to be increased in the future.
 #define RMW_GID_STORAGE_SIZE 24u
 
+#ifndef _WIN32
+# define RMW_DECLARE_DEPRECATED(name, msg) name __attribute__((deprecated(msg)))
+#else
+# define RMW_DECLARE_DEPRECATED(name, msg) name __pragma(deprecated(name))
+#endif
+
 /// Structure which encapsulates an rmw node
 typedef struct RMW_PUBLIC_TYPE rmw_node_t
 {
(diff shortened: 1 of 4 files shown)


rmw/include/rmw/types.h
```diff
@@ -444,12 +447,18 @@ typedef struct RMW_PUBLIC_TYPE rmw_qos_profile_t
   /// Durability QoS policy setting
   enum rmw_qos_durability_policy_t durability;
   /// The period at which messages are expected to be sent/received
+  /// RMW_TIME_INFINITY indicates "no deadline", so deadline is not tracked
+  /// RMW_TIME_ZERO means "not specified" and is changed to RMW_TIME_INFINITY
```
hidmic, line 451: @emersonknapp If "not specified", wouldn't it make sense to let the RMW implementation choose an appropriate default? As stated, `RMW_TIME_ZERO` and `RMW_TIME_INFINITY` seem redundant to me.

emersonknapp: They are redundant in the use case of these 3 qos policies, though other uses may have different meanings - that's why I wanted to specify on each policy rather than making it a "global declaration".

I'm considering, for these policies, RMW_TIME_ZERO to mean 'unspecified' (since we can't actually choose not to specify)

I know that ROS 2 != DDS, but within the DDS spec:
* Deadline defaults to infinite
* Liveliness lease duration defaults to infinite
* Lifespan defaults to infinite

I feel like we should not want to allow different RMW implementations to have different defaults for these policies so that code is more portable between implementations, but maybe I am in the wrong there. In that case, the proposal should be "RMW_TIME_ZERO is interpreted as 'unspecified' and the RMW implementation will use its default value"

hidmic: > They are redundant in the use case of these 3 qos policies, though other uses may have different meanings - that's why I wanted to specify on each policy rather than making it a "global declaration".

I see. I'm somewhat inclined towards having a sentinel default or unspecified value rather than re-purposing `RMW_TIME_ZERO`.

> I feel like we should not want to allow different RMW implementations to have different defaults for these policies so that code is more portable between implementations

Well, we do have system default QoS policies. I _think_ the intent was to support bypassing ROS 2 configuration whenever RMW implementation - specific configuration is used. But @wjwwood or @dirk-thomas may know better.

emersonknapp: I guess it makes sense to have the equivalent of `_SYSTEM_DEFAULT` for durations. What would be a good value for this?

hidmic: If we were to use `rcutils` time and duration types, negative values which are meaningless in this context could be used as sentinels.

hidmic: @emersonknapp friendly ping.

hidmic: @emersonknapp do you plan to circle back?

hidmic: @emersonknapp friendly ping

emersonknapp: Thanks for all the patient pings! I am planning to spend some time with this tomorrow

emersonknapp: And of course I got pulled onto an emergency situation for 2 days immediately after I said that. This is still on my immediate docket and I plan to update it within the coming week.

emersonknapp:
Closed by unintentional language in another PR comment - still working on this

emersonknapp:
I just want to check in that I am actively working on this right now (for any interested parties). I don't have the final API nailed down yet, but I have done some more data gathering to show why it is necessary to solve this for the RMW API to be correct.

The following values are gathered from `get_publishers_info_by_topic` when a single publisher is publishing with unspecified Deadline duration. It doesn't matter which RMW implementation is being used to publish, the value only depends on the implementation of the subscripton - I have tested it with only Cyclone and FastDDS, with all 4 pub/sub combinations.

| Value Type | Fast sec | Fast nsec | Fast total nsec | Cyclone sec | Cyclone nsec | Cyclone total nsec |
| --------------  | -             | -                | -              | -                    | -                     | -                     |
| decimal | 2147483647 | 4294967295 | 2147483651294967295 | 9223372036 | 854775807 | 9223372036854775807 |
| hex | 0x7FFFFFFF | 0xFFFFFFFF | 0x1DCD6500C46535FF | 0x225C17D04 | 0x32F2D7FF | 0x7FFFFFFFFFFFFFFF |
| DDS IDL PSM Constant | `DURATION_INFINITE_SEC` | `TIME_INVALID_SEC` | nothing | nothing | nothing | nothing |
| int limits | `INT32_MAX` | `UINT32_MAX` | nothing | nothing | nothing | `INT64_MAX` |


"DDS IDL PSM Constant" above refers to the values laid out in Section 2.3.3 of the DDS 1.4 spec. From my understanding this PSM is meant to provide common interfaces for programming languages. There are separate docs available as well:
* C language (Cyclone) - https://www.omg.org/spec/DDS/20140501/dds_dcps.idl
* C++ language (FastDDS) - https://www.omg.org/spec/DDS-PSM-Cxx/1.0/PDF
The C++ one doesn't lay out new constants, from what I understand is it is meant to replace the IDL-generated API with more idiomatic language

Based on the above information, the two implementations are providing values that are neither consistent with each other or guidelines from the specification - seems like we just have exposed internal values.

My next step is an updated API proposal, which will come with coupled prototype PRs for the RMW implementations.

rmw/include/rmw/types.h
```diff
@@ -328,6 +328,21 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
...
 
+/// Explicit representation of an infinite duration
+/**
+ * The value INT32_MAX is chosen for the following reasons:
+ * 1. Parity with the DDS IDL Platform Specific Model (PSM).
+ *    This was not a requirement, but it simplifies some conversions.
+ * 2. When summed for total nanoseconds as represented in rcutils_duration_value_t,
+ *    do not overflow 64 bit - as it would if INT64_MAX were chosen.
+ */
+static const rmw_time_t RMW_DURATION_INFINITE = {INT32_MAX, INT32_MAX};
```
mm318, line 339: Does this mean instances of `rmw_time_t` that has any one of its fields greater than `INT32_MAX` will be considered invalid? Perhaps add a utility function for checking that.

emersonknapp: That's not something I want to have to define here. This particular value happens to be non-normalized (nanos > 1s) which makes it "very invalid".

Based on the difficulties and edge cases around managing this split time, I'm starting to feel more inclined to resolve #215 _before_ fixing this, because it would make this API a lot simpler

emersonknapp:
Closing this in favor of #301 for now

PR #256, opened 2020-07-31
Update subscription API documentation

hidmic:
Precisely what the title says. Each commit is to be kept separately.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -687,7 +687,8 @@ rmw_create_subscription(
  * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given subscription handle
  * unchanged.
  * Otherwise, it will proceed despite errors, freeing as many resources as it can, including
- * the subscription handle. Usage of a deallocated subscription handle is undefined behavior.
+ * the subscription handle, and return `RMW_RET_ERROR`. Usage of a deallocated subscription
+ * handle is undefined behavior.
  *
  * \pre Given node must be the one the subscription was registered with.
  *
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -680,10 +680,12 @@ rmw_create_subscription(
   const rmw_qos_profile_t * qos_policies,
   const rmw_subscription_options_t * subscription_options);
 
-/// Finalize a given subscription handle, reclaim the resources, and deallocate the subscription handle.
+/// Finalize a given subscription handle, reclaim the resources, and deallocate the subscription
+/// handle.
 /**
  * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`
- * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given subscription handle unchanged.
+ * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given subscription handle
+ * unchanged.
  * Otherwise, it will proceed despite errors, freeing as many resources as it can, including
  * the subscription handle. Usage of a deallocated subscription handle is undefined behavior.
  *
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -645,10 +645,10 @@ rmw_fini_subscription_allocation(
  *   - type_support is a not valid non-null message type support, as returned by
  *     `ROSIDL_GET_MSG_TYPE_SUPPORT()`
  *   - topic_name is not a valid non-null topic name, according to
- *     `rmw_validate_full_topic_name()`
+ *     `rmw_validate_full_topic_name()` if ROS namespace conventions apply
  *   - qos_profile is not a fully specified non-null profile i.e. no UNKNOWN policies
- *   - subscription_options is not a valid non-null option set, as returned by
- *     `rmw_get_default_subscription_options()`
+ *   - subscription_options is not a valid non-null option set, such as the one
+ *     returned by `rmw_get_default_subscription_options()`
  *   - memory allocation fails during subscription creation
  *   - an unspecified error occurs
  *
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `test_rmw_implementation` and `rcl`, against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11736/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6817/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9574/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/11676/)

rmw/include/rmw/rmw.h
```diff
@@ -637,14 +637,38 @@ rmw_ret_t
...
- *   arguments for now.
+ * This function can fail, and therefore return `NULL`, if:
+ *   - node is not a valid non-null handle for this rmw implementation,
+ *     as returned by `rmw_create_node()`
+ *   - type_support is a not valid non-null message type support, as returned by
+ *     `ROSIDL_GET_MSG_TYPE_SUPPORT()`
+ *   - topic_name is not a valid non-null topic name, according to
+ *     `rmw_validate_full_topic_name()`
+ *   - qos_profile is not a fully specified non-null profile i.e. no UNKNOWN policies
+ *   - subscription_options is not a valid non-null option set, as returned by
```
gbiggs, line 650: It's a minor language thing, but "as returned by" implies that you have to pass the `rmw_subscription_options_t` returned by `rmw_get_default_subscription_options()`.

rmw/include/rmw/rmw.h
```diff
@@ -656,6 +680,34 @@ rmw_create_subscription(
   const rmw_qos_profile_t * qos_policies,
   const rmw_subscription_options_t * subscription_options);
 
+/// Finalize a given subscription handle, reclaim the resources, and deallocate the subscription
+/// handle.
+/**
+ * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`
```
gbiggs, line 686: Can this be specific about what logical errors are allowed to trigger an early return, rather than saying "such as"?

rmw/include/rmw/rmw.h
```diff
@@ -656,6 +680,34 @@ rmw_create_subscription(
...
+ * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given subscription handle
+ * unchanged.
+ * Otherwise, it will proceed despite errors, freeing as many resources as it can, including
```
gbiggs, line 689: > freeing as many resources as it can

This is terribly non-deterministic. After this function returns, the caller won't know what has been freed and what has not.

clalancette: > This is terribly non-deterministic. After this function returns, the caller won't know what has been freed and what has not.

I guess the other alternative is to give up as soon as we hit the first error.  That also seems non-deterministic (and hard to document), but would that be better in your view?

rmw/include/rmw/rmw.h
hidmic: Yes it can, and in this case, it's purely phrasing i.e. replacing "such as" with "namely" will remain true.

rmw/include/rmw/rmw.h
hidmic: Good point.

rmw/include/rmw/rmw.h
hidmic: > This is terribly non-deterministic. After this function returns, the caller won't know what has been freed and what has not.

I wouldn't say non-deterministic, but rather unspecified. And yes, the caller won't know to what extent the entity was finalized, nor can do anything about it as the subscription will have been deallocated. 

As to whether that's a good error handling approach during finalization or not, we've discussed this quite a bit throughout related PRs. To proceed with finalization despite errors precludes the need for strong error (exception) guarantees, which may not even be achievable for a given implementation, and/or dealing with partially finalized entities, likely implementation specific and thus unspecified. The fact that finalization functions are often invoked from object destructors in higher-level language bindings (i.e. at a point where the object is going out of scope anyways) may have also influenced the decision. 

I've had mixed feelings about this in the past. But at this point I think that allowing a finalization function to fail helps no one, let alone returning before completion. If it does, it's because you've hit (a) a transient error, which an implementation could deal with by deferring side effects if need be, (b) a logical error, in which the case the program is already broken, or (c) a system error, in which case you can no longer trust the OS that's supporting you.

gbiggs: Yeah, I guess it's the same as the approach that C++ destructors take: If there's an error in the destructor, give up and assume everything is broken beyond repair.

It is good, though, to be a bit more specific about what errors will cause predictable failure and which will leave things in an indeterminate state. The recent changing of "such as" to "namely" helps there but I think the text could be formatted to make it more obvious, e.g. using a list.

hidmic: > If there's an error in the destructor, give up and assume everything is broken beyond repair.

Yeah, but that's largely dependent on the application. Just like in a C++ destructor, you could suppress/log that error (exception) and keep going, which is by far the common pattern in `rcl` and above.

> It is good, though, to be a bit more specific about what errors will cause predictable failure and which will leave things in an indeterminate state.

Hmm, I'm not exactly sure what you're asking for. If you mean that we should specify all possible return codes, they are listed exhaustively below. If you mean that we should specify  failure modes, we can't as that's RMW implementation specific. If you mean that we should split the generic `RMW_RET_ERROR` return code into a few ones to aid the caller on how to proceed, I agree we could but that's orthogonal to this patch (and a new feature entirely).

ivanpauno: > Yeah, I guess it's the same as the approach that C++ destructors take: If there's an error in the destructor, give up and assume everything is broken beyond repair.

Maybe that will make destruction code simpler (no need to handle chained errors).

hidmic:
@gbiggs PTAL!

rmw/include/rmw/rmw.h
gbiggs: > Hmm, I'm not exactly sure what you're asking for.

If there are errors that could occur but will result in the function not doing anything, those need to be explicitly listed. If there are errors that are known that may leave things in an indefinite state, those also should be listed.

The two are separate from a statement saying "Unknown errors may occur". Basically I'm looking for information that allows a developer to be certain about whether things are now in an unknown state or not.

hidmic: > If there are errors that could occur but will result in the function not doing anything, those need to be explicitly listed. 

They are (though I'm not entirely convinced we should be handling logical errors in run-time in any way but terminating the program).

> If there are errors that are known that may leave things in an indefinite state, those also should be listed.

There are no known errors, only a blanket one.

> Basically I'm looking for information that allows a developer to be certain about whether things are now in an unknown state or not.

See https://github.com/ros2/rmw/pull/256/commits/e6620b03980113bb228f444d1afacce7fce138a7. If you get `RMW_RET_ERROR`, then it's highly likely that something didn't get fully finalized.

PR #258, opened 2020-08-10
Update serialization/deserialization API documentation.

hidmic:
Precisely what the title says. Just clarifying what was already given, plus specifying `rmw_serialized_message_t` (regardless of it being an alias). 

I'll point out that this API departs from the rest, in that every argument MUST be correct or you'll incur UB.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -565,6 +565,18 @@ rmw_publisher_assert_liveliness(const rmw_publisher_t * publisher);
  * \pre Given serialized message must be a valid non-null instance, initialized
  *   by the caller.
  *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | Yes [2]
+ * Uses Atomics       | Yes
+ * Lock-Free          | Yes
+ * <i>[1] if the given serialized message does not have enough capacity to hold
+ *        the ROS message serialization</i>
+ * <i>[2] as long as no two concurrent calls make use of the same `serialized_message`
+ *        object and a given `serialized_message` allocator is thread-safe as well </i>
+ *
  * \param[in] ros_message the typed ROS message
  * \param[in] type_support the typesupport for the ROS message
  * \param[out] serialized_message the destination for the serialize ROS message
@@ -594,6 +606,16 @@ rmw_serialize(
  *   by `rosidl` APIs.
  * \pre Given ROS message must be a valid non-null instance, initialized
  *   by the caller.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | Yes [2]
+ * Uses Atomics       | Yes
+ * Lock-Free          | Yes
+ * <i>[1] if the given ROS message contains unbounded fields</i>
+ * <i>[2] as long as no two concurrent calls make use of the same `ros_message` object </i>
  *
  * \param[in] serialized_message the serialized message holding the byte stream
  * \param[in] type_support the typesupport for the typed ros message
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -570,12 +570,13 @@ rmw_publisher_assert_liveliness(const rmw_publisher_t * publisher);
  * ------------------ | -------------
  * Allocates Memory   | Maybe [1]
  * Thread-Safe        | Yes [2]
- * Uses Atomics       | Yes
- * Lock-Free          | Yes
+ * Uses Atomics       | Maybe [3]
+ * Lock-Free          | Maybe [3]
  * <i>[1] if the given serialized message does not have enough capacity to hold
  *        the ROS message serialization</i>
  * <i>[2] as long as no two concurrent calls make use of the same `serialized_message`
  *        object and a given `serialized_message` allocator is thread-safe as well </i>
+ * <i>[3] rmw implementation defined, check the implementation documentation</i>
  *
  * \param[in] ros_message the typed ROS message
  * \param[in] type_support the typesupport for the ROS message
@@ -612,10 +613,11 @@ rmw_serialize(
  * ------------------ | -------------
  * Allocates Memory   | Maybe [1]
  * Thread-Safe        | Yes [2]
- * Uses Atomics       | Yes
- * Lock-Free          | Yes
+ * Uses Atomics       | Maybe [3]
+ * Lock-Free          | Maybe [3]
  * <i>[1] if the given ROS message contains unbounded fields</i>
  * <i>[2] as long as no two concurrent calls make use of the same `ros_message` object </i>
+ * <i>[3] rmw implementation defined, check the implementation documentation</i>
  *
  * \param[in] serialized_message the serialized message holding the byte stream
  * \param[in] type_support the typesupport for the typed ros message
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -569,14 +569,12 @@ rmw_publisher_assert_liveliness(const rmw_publisher_t * publisher);
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | Maybe [1]
- * Thread-Safe        | Yes [2]
- * Uses Atomics       | Maybe [3]
- * Lock-Free          | Maybe [3]
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [2]
+ * Lock-Free          | Maybe [2]
  * <i>[1] if the given serialized message does not have enough capacity to hold
  *        the ROS message serialization</i>
- * <i>[2] as long as no two concurrent calls make use of the same `serialized_message`
- *        object and a given `serialized_message` allocator is thread-safe as well </i>
- * <i>[3] rmw implementation defined, check the implementation documentation</i>
+ * <i>[2] rmw implementation defined, check the implementation documentation</i>
  *
  * \param[in] ros_message the typed ROS message
  * \param[in] type_support the typesupport for the ROS message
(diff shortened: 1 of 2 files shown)


rmw/include/rmw/serialized_message.h
```diff
@@ -29,9 +29,49 @@ extern "C"
...
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return 'RMW_RET_BAD_ALLOC` if no memory could be allocated correctly
+ * \return `RMW_RET_ERROR` if an unexpected error occurs
+ */
 #define rmw_serialized_message_init rcutils_uint8_array_init
+
+/// Finalize a serialized message.
+/**
+ * Passing an uninitialized instance to this function leads to undefined
```
gbiggs, line 50: Is it better to make this a checked pre-condition?

clalancette: At the very least, it's probably a good idea to add a `\pre` notice like we did above.

rmw/include/rmw/serialized_message.h
```diff
@@ -29,9 +29,49 @@ extern "C"
...
+/**
+ * \param[inout] serialized_message a pointer to the serialized message to initialize
+ * \param[in] message_capacity the size of the memory to allocate
+ * \param[in] allocator the allocator to use for the memory allocation
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return 'RMW_RET_BAD_ALLOC` if no memory could be allocated correctly
+ * \return `RMW_RET_ERROR` if an unexpected error occurs
+ */
 #define rmw_serialized_message_init rcutils_uint8_array_init
```
clalancette, line 46: Just a thought here.  I would find this documentation easier to read if `rmw_serialized_message_init` had arguments.  That is, can we change the macro to:

```
#define rmw_serialized_message_init(serialized_message, message_capacity, allocator) rcutils_uint8_array_init(serialized_message, message_capacity, allocator)
```

The same goes for all of the other `rmw_serialized*` macros in this file.

rmw/include/rmw/serialized_message.h
```diff
@@ -29,9 +29,49 @@ extern "C"
...
+ */
 #define rmw_serialized_message_fini rcutils_uint8_array_fini
+
+/// Resize the internal buffer of the serialized message.
+/**
+ * The internal buffer of the serialized message can be resized dynamically if needed.
+ * If the new size is smaller than the current capacity, then the memory is
+ * truncated.
+ * Be aware, that this might deallocate the memory and therefore invalidates any
+ * pointers to this storage.
```
clalancette, line 66: This sentence is grammatically correct, but I would still find a modification slightly easier to read:

```suggestion
 * Be aware, that this might deallocate the memory and therefore invalidate any
 * pointers to this storage.
```

rmw/include/rmw/serialized_message.h
```diff
@@ -29,9 +29,49 @@ extern "C"
...
+ *
```
clalancette, line 67: Should we add `\pre serialized_message must have been initialized` or similar?

rmw/include/rmw/serialized_message.h
```diff
@@ -29,9 +29,49 @@ extern "C"
...
+ * \param[inout] serialized_message pointer to the serialized message to be resized
+ * \param[in] new_size the new size of the internal buffer
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if new_size is set to zero
```
clalancette, line 71:
```suggestion
 * \return `RMW_RET_INVALID_ARGUMENT` if new_size is set to zero, or
```

rmw/include/rmw/serialized_message.h
```diff
@@ -29,9 +29,49 @@ extern "C"
...
 #define rmw_get_zero_initialized_serialized_message rcutils_get_zero_initialized_uint8_array
+
+/// Initialize a serialized message, zero initializing its contents.
+/**
+ * \param[inout] serialized_message a pointer to the serialized message to initialize
+ * \param[in] message_capacity the size of the memory to allocate
+ * \param[in] allocator the allocator to use for the memory allocation
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return 'RMW_RET_BAD_ALLOC` if no memory could be allocated correctly
```
clalancette, line 43:
```suggestion
 * \return 'RMW_RET_BAD_ALLOC` if no memory could be allocated correctly, or
```

rmw/include/rmw/serialized_message.h
hidmic: Sounds good. At some point we should correct `rcutils_uint8_array_t` API documentation as well.

rmw/include/rmw/serialized_message.h
hidmic: We can constrain possible states to zero-initialized or initialized, but IMO to say that using an uninitialized variable is UB goes without saying.

rmw/include/rmw/serialized_message.h
hidmic: This needs an update. Passing uninitialized arguments will always be UB, and known states (zero-initialized or initialized) will be handled correctly in runtime.

hidmic:
Alright, PTAL! I've also added API attributes based on what current implementations do, thread-safety being the most constraining spec.

hidmic:
CI up to `test_rmw_implementation` and `rmw`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11885/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6930/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9682/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11822/)

rmw/include/rmw/rmw.h
```diff
@@ -556,7 +556,27 @@ rmw_publisher_assert_liveliness(const rmw_publisher_t * publisher);
...
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | Yes [2]
+ * Uses Atomics       | Maybe [3]
+ * Lock-Free          | Maybe [3]
+ * <i>[1] if the given serialized message does not have enough capacity to hold
+ *        the ROS message serialization</i>
+ * <i>[2] as long as no two concurrent calls make use of the same `serialized_message`
```
ivanpauno, line 577: what if `ros_message` is being modified in another thread while being serialized?

IMO, this function is simply NO thread safe.

rmw/include/rmw/rmw.h
```diff
@@ -577,10 +597,27 @@ rmw_serialize(
...
+ * \pre Given typesupport must be a valid non-null instance, as provided
+ *   by `rosidl` APIs.
+ * \pre Given ROS message must be a valid non-null instance, initialized
+ *   by the caller.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | Yes [2]
```
ivanpauno, line 615: same as above, this is not thread safe IMO

rmw/include/rmw/serialized_message.h
gbiggs: The thing here is that we are not just saying "it's undefined", but we are in a way specifying whether implementers of this function declaration need to add a check for null pointers, etc. That, I think, makes it worth being more explicit.

hidmic: I think the added `\pre` and `\remarks` should have us covered.

rmw/include/rmw/rmw.h
hidmic: I think we're discussing in circles across PRs on what we mean by _thread-safe_. If we say that for a function to be thread-safe not only it should be reentrant but the read-write operations it performs on any of its arguments is also thread-safe, then yes, you're right, this function is not thread-safe. And neither are [`rcutils_system_time_now()`](https://github.com/ros2/rcutils/blob/049167533d473cca5c8ae8347becdd2209503927/include/rcutils/time.h#L50-L75) or [`rcl_parse_arguments()`](https://github.com/ros2/rcl/blob/6ff418ebc1dff312a6d8495494b183b998e9c14f/rcl/include/rcl/arguments.h#L61-L125).  OrAnd I highly doubt any of our C APIs can fulfill that requirement (we don't put synchronization mechanisms in user facing API, barely any atomics), and thus we may as well put a blanket statement saying nothing is thread-safe.

Now, if we say that these APIs are thread-safe as long as they are reentrant and we document (or imply) mutually exclusive write access on their output arguments as a requirement, then we can have some thread-safe APIs and this one is.

hidmic: Also, looking at [POSIX safety terminology](https://www.gnu.org/software/libc/manual/html_node/POSIX-Safety-Concepts.html), it doesn't seem it makes such promises on their arguments either. For instance, [`memset`](https://man7.org/linux/man-pages/man3/memset.3.html) is supposedly MT-Safe even though concurrently modyfing it or even realloc'ing may result in unexpected outcomes or memory corruption.

[This blog](https://developers.redhat.com/blog/2014/09/10/multi-thread-async-signal-and-async-cancel-safety-docs-in-gnu-libc/) goes some length discussing POSIX safety concepts.

hidmic: Alright, after an offline discussion with @ivanpauno, we've decided it's best to (1.) relax the thread-safety requirement here, and (2.) kickstart a separate discussion to standardize and document our thread-safety definitions.

gbiggs: Sounds like a good plan to me.

PR #262, opened 2020-08-20
Update publisher/subscription matched count API documentation.

hidmic:

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -448,10 +448,21 @@ rmw_publish_loaned_message(
  * Query the underlying middleware to determine how many subscriptions are
  * matched to a given publisher.
  *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
  * \param[in] publisher the publisher object to inspect
  * \param[out] subscription_count the number of subscriptions matched
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if either argument is null, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if node or publisher
+ *   implementation identifier does not match, or
  * \return `RMW_RET_ERROR` if an unexpected error occurs.
  */
 RMW_PUBLIC
@@ -753,10 +764,21 @@ rmw_destroy_subscription(rmw_node_t * node, rmw_subscription_t * subscription);
  * Query the underlying middleware to determine how many publishers are
  * matched to a given subscription.
  *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
  * \param[in] subscription the subscription object to inspect
  * \param[out] publisher_count the number of publishers matched
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if either argument is null, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if subscription
+ *   implementation identifier does not match, or
  * \return `RMW_RET_ERROR` if an unexpected error occurs.
  */
 RMW_PUBLIC
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -461,7 +461,7 @@ rmw_publish_loaned_message(
  * \param[out] subscription_count the number of subscriptions matched
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if either argument is null, or
- * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if node or publisher
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if publisher
  *   implementation identifier does not match, or
  * \return `RMW_RET_ERROR` if an unexpected error occurs.
  */


hidmic:
Writing tests for this portion of the API brings up an interesting question. What's a _reasonable_ latency for pub/sub match events (within the same process, within the same host, within the same network)? We don't have (nor specify) a guaranteed upper bound to rely on, and there's no API to wait for these events.

hidmic:
Just to clarify, I'm looking for a way to specify within which time interval (in seconds or in received graph events) the API is expected to reflect the state of the system. Otherwise I can't tell which implementation behaves correctly and which doesn't.

rmw/include/rmw/rmw.h
```diff
@@ -448,10 +448,21 @@ rmw_publish_loaned_message(
...
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
  * \param[in] publisher the publisher object to inspect
  * \param[out] subscription_count the number of subscriptions matched
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if either argument is null, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if node or publisher
```
clalancette, line 464:
```suggestion
 * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if publisher
```

Or maybe I'm wrong.  I guess my question is; why is this different than `rmw_subscription_count_matched_publishers`?

wjwwood: This seems like the right change to me. It's not possible AFAIK for the node which is referenced from within the publisher to not match the publisher, so checking the publisher only is fine I think.

hidmic: You're spot on. A left over. See https://github.com/ros2/rmw/pull/262/commits/1083ea1c855f9d56de25ca8b307c6b3a2f7421aa.

hidmic:
Thanks for the prompt reviews guys! What do we do about https://github.com/ros2/rmw/pull/262#issuecomment-677696961? I need a boundary for tests in https://github.com/ros2/rmw_implementation/pull/119 and I think it'd be good we mention that here.

wjwwood:
I don't think we can put a latency on that which will be meaningful for all rmw implementations... Either in terms of time or number of events. Why does that need to be part of the documentation?

hidmic:
> I don't think we can put a latency on that which will be meaningful for all rmw implementations... Either in terms of time or number of events. Why does that need to be part of the documentation?

In my mind, I cannot test for pub/sub matching if I don't have a way to wait for it to happen. If I use the graph guard condition and  say, a retry count, that count becomes an implicit requirement for the implementation. I think that it should be explicit (along with all the necessary constraints, like for `pub/sub within the same process` or for `pub/sub within the same host with < 10% CPU load`).

Or not test pub/sub matching (which is a bit of a bummer).

wjwwood:
I don't understand why we need a number in the docs. Why can we not have a retry loop with an empirically calculated number? That's what we do everywhere.

When we send a message and check to see it is received we have a wait period and/or a retry loop but that's not documented and I don't think it could be given that it would impacted by the underlying system.

I think we shouldn't put numbers on things like this because it depends too much on the underlying hardware and middleware implementation.

hidmic:
> I don't understand why we need a number in the docs. Why can we not have a retry loop with an empirically calculated number? That's what we do everywhere.

I know. And we have a bunch of conditionals in gtests, launch tests, pytests, and CMakeLists.txt to cope with the nitty-gritty of the implementations we have in our source tree.
 
Having `test_rmw_implementation` is good for RMW implementations coverage, but it also gives us the opportunity to refine `rmw` API specifications and have a test suite in place that checks if any given implementation reasonably meets those specifications (all for the same price). But to do that, API specifications have to be complete.

Now that I think about this again, perhaps we can mention the tests themselves, and document in what conditions `test_rmw_implementation` tests are valid.

> and middleware implementation.

Hmm, wait, is that the case? Would it be OK if a middleware doesn't match publishers and subscriptions that don't stay around for more than, say, 5 minutes? I mean, that's fine, but then (a) generic RMW implementation testing is not possible, and I have to separate these from the rest, and (b) we should add that to the docs! Like, check your RMW implementation documentation to understand what to expect from this API.

wjwwood:
Well we can have large upper bounds, like "a subscription and publisher shall be matched within 2 minutes of their creation", but to make that tighter would just make it hard to avoid flaky tests, and to make it broader makes it less and less useful of a number while still not completely eliminating the chance that some hardware or new middleware implementation not meeting the goal.

So if we want to add them that's fine, but I can't give you a number that isn't arbitrary myself. Sorry, I know that's not satisfying but I've been dealing with this while trying to write requirements for rclcpp for Apex and in my opinion there are no satisfying answers.

hidmic:
Yeah, this is a tough one. For worst, skimming through RMW implementations, it doesn't seem like pub/sub matching triggers graph guard conditions at all -- which is curious considering [`rcl` tests](https://github.com/ros2/rcl/blob/6d824bccc95bc238e668b9846eebba56dbe968a7/rcl/test/rcl/test_count_matched.cpp#L80) appear to rely on that.

Perhaps, the conclusion here is that the `rmw` API is lacking the features to enable these tests, and, for the time being, we'll have to move forward without it. That is, loops around checks, with an arbitrary cap like we do elsewhere, and a big TODO for change in the future.

gbiggs:
> So if we want to add them that's fine, but I can't give you a number that isn't arbitrary myself. Sorry, I know that's not satisfying but I've been dealing with this while trying to write requirements for rclcpp for Apex and in my opinion there are no satisfying answers.

In my experience, the best way to deal with requirements like that is to be relative, rather than absolute. "A must happen before B can happen", "A must happen before the system attempts C", "The system shall not be considered started until A has happened", etc. Specific time limits are too dependent on the hardware, the underlying implementation, and the needs of the application (my application might be happy to wait 5 minutes, yours might not) to be worth it.

> I expect (hope for?) them to be the contract and verification process that any RMW implementation has to comply with and go through, respectively, and not just the tests that ensure our current code doesn't regress.

I 100% agree with the sentiment that we want the API documentation to be the enforceable contract. However I don't think that contract must include specific time limits even where we can't provide them in any meaningful way. Specifying it in terms of "if this has not happened, you can't do X" makes more sense to me. It's more stateful, but it's a stateful API anyway.

hidmic:
> Specific time limits are too dependent on the hardware, the underlying implementation, and the needs of the application (my application might be happy to wait 5 minutes, yours might not) to be worth it.

I agree with that. Still, we will have a finite timeout. A timeout that will just be enough to get `rmw_fastrtps*_cpp`, `rmw_cyclonedds_cpp`, and `rmw_connext_cpp` passing tests in our CI machines w/o flakes. And thus test results won't be generic.

---

This was a very interesting discussion, but I will not push it any further. Thank you all for the time!

hidmic:
FYI https://github.com/ros2/rmw/issues/264

hidmic:
CI up to `test_rmw_implementation`, against all Tier 1 RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12070/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7095/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9844/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12001/)

PR #270, opened 2020-09-01
Update publishing API documentation.

hidmic:

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -483,7 +483,8 @@ rmw_return_loaned_message_from_publisher(
  * \pre Given `ros_message` must be a valid message, whose type matches the message type
  *   support the `publisher` was registered with on creation.
  * \pre If not NULL, given `allocation` must be a valid publisher allocation initialized
- *   with rmw_publisher_allocation_init().
+ *   with rmw_publisher_allocation_init() with a message type support that matches the
+ *   one registered with `publisher` on creation.
  *
  * \param[in] publisher Publisher to be used to send message.
  * \param[in] ros_message Type erased ROS message to be sent.
@@ -544,7 +545,8 @@ rmw_publish(
  * \pre Given `ros_message` must be a valid message, borrowed from the same publisher using
  *   rmw_borrow_loaned_message().
  * \pre If not NULL, given `allocation` must be a valid publisher allocation, initialized
- *   with rmw_publisher_allocation_init().
+ *   with rmw_publisher_allocation_init() with a message type support that matches the
+ *   one registered with `publisher` on creation.
  *
  * \param[in] publisher Publisher to be used to send message.
  * \param[in] ros_message Loaned type erased ROS message to be sent.
@@ -642,7 +644,8 @@ rmw_publisher_get_actual_qos(
  *   rmw_serialized_message_init() and containing the serialization of a ROS message whose
  *   type matches the message type support the `publisher` was registered with on creation.
  * \pre If not NULL, given `allocation` must be a valid publisher allocation initialized
- *   with rmw_publisher_allocation_init().
+ *   with rmw_publisher_allocation_init() with a message type support that matches the
+ *   one registered with `publisher` on creation.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -473,9 +473,9 @@ rmw_return_loaned_message_from_publisher(
  *   Publishers are thread-safe objects, and so are all operations on them except for finalization.
  *   Therefore, it is safe to publish using the same publisher concurrently.
  *   However, when publishing regular ROS messages:
- *   - Access to ROS messages is read-only but it is not synchronized.
+ *   - Access to the ROS message is read-only but it is not synchronized.
  *     Concurrent `ros_message` reads are safe, but concurrent reads and writes are not.
- *   - Access to publisher allocations may be synchronized, but it is not required to.
+ *   - Access to the publisher allocation may be synchronized, but it is not required to.
  *     Check the implementation documentation to learn about publisher allocations' thread-safety.
  *
  * \pre Given `publisher` must be a valid publisher, as returned by rmw_create_publisher().
@@ -536,7 +536,7 @@ rmw_publish(
  *   - Ownership of the loaned ROS message is given back to the middleware.
  *     This transfer is not synchronized, and thus it is not safe to publish the
  *     same loaned ROS message concurrently.
- *   - Access to publisher allocations may be synchronized, but it is not required to.
+ *   - Access to the publisher allocation may be synchronized, but it is not required to.
  *     Check the implementation documentation to learn about publisher allocations' thread-safety.
  *
  * \pre Given `publisher` must be a valid publisher, as returned by rmw_create_publisher().
@@ -658,9 +658,9 @@ rmw_publisher_get_actual_qos(
  *   Publishers are thread-safe objects, and so are all operations on them except for finalization.
  *   Therefore, it is safe to publish using the same publisher concurrently.
  *   However, when publishing serialized ROS messages:
- *   - Access to byte streams is read-only but it is not synchronized.
+ *   - Access to the byte stream is read-only but it is not synchronized.
  *     Concurrent `serialized_message` reads are safe, but concurrent reads and writes are not.
- *   - Access to publisher allocations may be synchronized, but it is not required to.
+ *   - Access to the publisher allocation may be synchronized, but it is not required to.
  *     Check the implementation documentation to learn about publisher allocations'
  *     thread-safety.
  *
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -397,15 +397,33 @@ rmw_return_loaned_message_from_publisher(
...
+ * \pre Given ROS message must match the message type support the publisher
+ *   was registered with on creation.
+ * \pre If not NULL, given allocation must be a valid publisher allocation initialized
+ *   with rmw_publisher_allocation_init().
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | No
```
wjwwood, line 414: This function does need to be thread-safe. For example, calling it on a publisher, with different allocation objects, from multiple threads should be fine.

rmw/include/rmw/rmw.h
```diff
@@ -416,23 +434,42 @@ rmw_publish(
...
+ * \pre Given ROS message must have been borrowed from the same publisher using
+ *   rmw_borrow_loaned_message().
+ * \pre If not NULL, given allocation must be a valid publisher allocation initialized
+ *   with rmw_publisher_allocation_init().
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | No
```
wjwwood, line 458: Same thing here, except you need to use unique messages from separate threads as well.

rmw/include/rmw/rmw.h
```diff
@@ -510,18 +547,37 @@ rmw_publisher_get_actual_qos(
...
+ * \pre Given serialized ROS message must match the message type support the publisher
+ *   was registered with on creation.
+ * \pre If not NULL, given allocation must be a valid publisher allocation initialized
+ *   with rmw_publisher_allocation_init().
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe [1]
+ * Thread-Safe        | No
```
wjwwood, line 568: Same thing as the normal publish function.

rmw/include/rmw/rmw.h
```diff
@@ -397,15 +397,33 @@ rmw_return_loaned_message_from_publisher(
...
+ * Uses Atomics       | Maybe [2]
+ * Lock-Free          | Maybe [2]
+ * <i>[1] if no publisher allocation is given, otherwise it will not</i>
```
wjwwood, line 417: This is potentially true, but even when you use a publisher allocation it could still allocate memory, since the rmw layer could either ignore the publisher allocation (not used yet) or if the message has unbounded, dynamically sized fields then it might require memory allocations still. For fixed sized messages or messages will only bounded fields then it could be a stronger guarantee.

rmw/include/rmw/rmw.h
hidmic: Ok, let's try to exercise some of the discussion on https://github.com/ros2/ros2/issues/1004. In general, we require `rmw_publisher_t` operations other than finalization to be thread-safe. In particular, publish API is thread-safe, but ROS message read/write access is not synchronized, and publisher allocation objects use is not required (but could be) thread-safe.

Fitting that into this documentation format, we could mark `Thread-Safe` as `Yes [1]` and add that `[1] it is safe for the same publisher to publish concurrently, but ROS message access is not synchronized and publisher allocations are not required to be`. WDYT?

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/270#discussion_r481328354.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/270#discussion_r481328354.

rmw/include/rmw/rmw.h
hidmic: Alright, let's make that clarification.

rmw/include/rmw/rmw.h
wjwwood: That sounds right to me. Just reiterating to check our shared understanding:

- same publisher: ok
- same message: ok, but not synchronized (so have to wait until every publish call is done to modify it)
- same allocation: could be, but not guaranteed, same as "not thread-safe to reuse allocations"

hidmic: Exactly. I may get a bit creative with the format though. It's a lot of information to pack into a small note.

hidmic: See https://github.com/ros2/rmw/pull/270/commits/f01962ad6de7c27977f1a3955d999894e676c2b3. I think using paragraphs like this makes sense. I moved sections around a bit so that argument and return value documentation isn't so far down. Perhaps having the table float on the right would also be nice, but it's too much HTML for this PR.

hidmic:
Alright, I think this is now complete. PTAL !

hidmic:
Ok, now it is ready. A couple things were either not constrained enough or not generic enough.

rmw/include/rmw/rmw.h
```diff
@@ -510,18 +632,54 @@ rmw_publisher_get_actual_qos(
...
- * The publisher must already be registered with the correct message type
- * support so that it can send serialized data corresponding to that type.
- * This function sends the serialized byte stream directly over the wire without
- * having to serialize the message first.
- * A ROS message can be serialized manually using the rmw_serialize() function.
- *
- * \param[in] publisher The publisher object registered to send the message.
- * \param[in] serialized_message The serialized message holding the byte stream.
- * \param[in] allocation Specify preallocated memory to use (may be NULL).
+ * Send a ROS message serialized in a byte stream directly over the wire to
```
hidmic, line 637: @hidmic what about implementations that do not use serialization (e.g. shared memory-based ones)? Do we allow them to give back the ROS message as a blob?

rmw/include/rmw/rmw.h
```diff
@@ -350,23 +350,47 @@ RMW_WARN_UNUSED
...
+ *
+ * <i>[1] implementation defined, check implementation documentation.</i>
+ *
+ * \par Memory allocation
+ *   It is implementation defined whether memory will be allocated on borrow or not.
+ *   Check the implementation documentation to learn about memory allocation
+ *   guarantees when using ROS message loaning support.
+ *
+ * \par Thread-safety
+ *   Publishers are thread-safe objects, and so are all operations on them except for finalization.
```
ivanpauno, line 375: Unrelated note: It would be great to have this kind of notes together with the object definition

rmw/include/rmw/rmw.h
ivanpauno: I think in those cases they can implement "serialize/deserialize" as a noop, and publish/taking the serialized message would be the same that the non-serialized one.

hidmic: > more than a noop, the serialized message would be an exact copy of the original one.

Exactly. I didn't think we should make any promises either, but I wanted to make sure others agree.

rmw/include/rmw/rmw.h
hidmic: Yeah. I didn't want to depart that much from how we usually document these things. 

Extra points for documentation format proposals on https://github.com/ros2/ros2/issues/1004 :smiley:

rmw/include/rmw/rmw.h
wjwwood: I would just say "publish a serialized ROS message", because several things could occur at this point:

- it could go to the wire
- it could need to be deserialized and converted to a new type (DDS type maybe if serialized publishing is not supported properly)
- it could be sent over shared memory or within the process as a serialized blob
- other things possibly...

So I'd just say "published" where that means "delivered to all matched subscriptions", without saying more than that.

wjwwood (approved):
lgtm other than the one comment thread

wjwwood (approved):
lgtm, thanks this is really a great improvement!

hidmic:
Alright, that should be it.

rmw/include/rmw/rmw.h
```diff
@@ -350,23 +350,54 @@ RMW_WARN_UNUSED
...
+ *   Therefore, it is safe to borrow ROS messages from the same publisher concurrently.
+ *
+ * \pre Given `publisher` must be a valid publisher, as returned by rmw_create_publisher().
+ * \pre Given `type_support` must be a valid `rosidl` message type support, matching the
+ *   one registered with the `publisher` on creation.
```
ivanpauno, line 387: unrelated, but it's weird that we're passing the typesupport again here, when the publisher already stores one internally.

ivanpauno (approved):
lgtm

rmw/include/rmw/rmw.h
hidmic: Yeap, it is weird.

hidmic:
Going in!

PR #271, opened 2020-09-02
Update taking API documentation.

hidmic:
Precisely what the title says. Closely related to #270.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -886,8 +886,10 @@ rmw_subscription_get_actual_qos(
  * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
  *   with rmw_subscription_allocation_init() with a message type support that matches the
  *   one registered with the `subscription` on creation.
- * \post If rmw_take() fails or `taken` is false, `ros_message` will be left in a valid
- *   but unknown state.
+ * \post Given `ros_message` will remain a valid message.
+ *   It will be left unchanged if this function fails early due to a logical error, such as an
+ *   invalid argument, or in an unknown yet valid state if it fails due to a runtime error.
+ *   It will also be left unchanged if this function succeeds but `taken` is false.
  *
  * \param[in] subscription Subscription to take message from.
  * \param[out] ros_message Type erased ROS message to write to.
@@ -959,8 +961,11 @@ rmw_take(
  * \pre If not NULL, given `allocation` must be a valid subscription allocation
  *   initialized with rmw_subscription_allocation_init() with a message type support
  *   that matches the one registered with the `subscription` on creation.
- * \post If rmw_take_with_info() fails or `taken` is false, both `ros_message` and
- *   `message_info` will be left in a valid but unknown state.
+ * \post Given `ros_message` will remain a valid message, and
+ *   `message_info`, valid message metadata.
+ *   Both will be left unchanged if this function fails early due to a logical error, such as
+ *   an invalid argument, or in an unknown yet valid state if it fails due to a runtime error.
+ *   Both will also be left unchanged if this function succeeds but `taken` is false.
  *
  * \param[in] subscription Subscription to take ROS message from.
  * \param[out] ros_message Type erased ROS message to write to.
@@ -989,11 +994,12 @@ rmw_take_with_info(
 
 /// Take multiple incoming ROS messages with their metadata.
 /**
- * Take a sequence of ROS messages received by the given subscription, removing
- * them from internal queues.
+ * Take a sequence of consecutive ROS messages received by the given subscription,
+ * removing them from internal queues.
  * While `count` ROS messages may be requested, fewer messages may have been
  * received by the subscription.
- * This function will succeed even if fewer (or zero) messages were received.
+ * This function will only take what has been already received, and it will
+ * succeed even if fewer (or zero) messages were received.
  * In this case, only currently available messages will be returned.
  * The `taken` flag indicates the number of ROS messages actually taken.
  *
@@ -1023,6 +1029,8 @@ rmw_take_with_info(
  *   Subscriptions are thread-safe objects, and so are all operations on them except for
  *   finalization.
  *   Therefore, it is safe to take from the same subscription concurrently.
+ *   Moreover, the sequence of ROS messages taken is guaranteed to be consecutive and to
+ *   preserve the order in the subscription queues, despite any concurrent takes.
  *   However, when taking a sequence of ROS messages with metadata:
  *   - Access to ROS message sequences is not synchronized.
  *     It is not safe to read or write `message_sequence` while rmw_take_sequence() uses it.
@@ -1048,8 +1056,11 @@ rmw_take_with_info(
  * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
  *   with rmw_subscription_allocation_init() with a message type support that matches the
  *   one registered with `subscription` on creation.
- * \post If rmw_take_sequence() fails or `taken` is false, both `message_sequence` and
- *   `message_info_sequence` will be left in a valid but unknown state.
+ * \post Given `message_sequence` will remain a valid message sequence, and
+ *   `message_info_sequence`, a valid message metadata sequence.
+ *   Both will be left unchanged if this function fails early due to a logical error, such as
+ *   an invalid argument, or in an unknown yet valid state if it fails due to a runtime error.
+ *   Both will also be left unchanged if this function succeeds but `taken` is false.
  *
  * \param[in] subscription Subscription to take ROS message from.
  * \param[in] count Number of messages to attempt to take.
@@ -1146,8 +1157,11 @@ rmw_take_sequence(
  * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
  *   with rmw_subscription_allocation_init() with a message type support that matches the
  *   one registered with `subscription` on creation.
- * \post If rmw_take_serialized_message() fails or `taken` is false, `serialized_message`
- *   will be left in a valid but unknown state.
+ * \post Given `serialized_message` will remain a valid serialized message.
+ *   It will be left unchanged if this function fails early due to a logical error,
+ *   such as an invalid argument, or in an unknown yet valid state if it fails due to a
+ *   runtime error.
+ *   It will also be left unchanged if this function succeeds but `taken` is false.
  *
  * \param[in] subscription Subscription to take ROS message from.
  * \param[out] serialized_message Byte stream to write to.
@@ -1226,8 +1240,12 @@ rmw_take_serialized_message(
  * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
  *   with rmw_subscription_allocation_init() with a message type support that matches the
  *   one registered with `subscription` on creation.
- * \post If rmw_take_serialized_message_with_info() fails or `taken` is false, both
- *   `serialized_message` and `message_info` will be left in a valid but unknown state.
+ * \post Given `serialized_message` will remain a valid serialized message, and `message_info`,
+ *   valid message metadata.
+ *   Both will be left unchanged if this function fails early due to a logical error,
+ *   such as an invalid argument, or in an unknown yet valid state if it fails due to a
+ *   runtime error.
+ *   It will also be left unchanged if this function succeeds but `taken` is false.
  *
  * \param[in] subscription Subscription to take ROS message from.
  * \param[out] serialized_message Byte stream to write to.
@@ -1303,6 +1321,8 @@ rmw_take_serialized_message_with_info(
  * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
  *   with rmw_subscription_allocation_init() with a message type support that matches the
  *   one registered with `subscription` on creation.
+ * \post Given `loaned_message` will remain unchanged, or point to a valid message if
+ *   this function was successful and `taken` is true.
  *
  * \param[in] subscription Subscription to take ROS message from.
  * \param[inout] loaned_message Pointer to type erased ROS message taken
@@ -1370,6 +1390,19 @@ rmw_take_loaned_message(
  *     Check the implementation documentation to learn about subscription allocations'
  *     thread-safety.
  *
+ * \pre Given `subscription` must be a valid subscription, as returned
+ *   by rmw_create_subscription().
+ * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
+ *   with rmw_subscription_allocation_init() with a message type support that matches the
+ *   one registered with `subscription` on creation.
+ * \post Given `loaned_message` will remain unchanged, or point to a valid message if
+ *   this function was successful and `taken` is true.
+ * \post Given `message_info` will remain valid message metadata.
+ *   It will be left unchanged if this function fails early due to a logical error,
+ *   such as an invalid argument, or in an unknown yet valid state if it fails due to a
+ *   runtime error.
+ *   It will also be left unchanged if this function succeeds but `taken` is false.
+ *
  * \param[in] subscription Subscription to take ROS message from.
  * \param[inout] loaned_message Pointer to type erased ROS message taken
  *   and loaned by the middleware.
@@ -1401,7 +1434,9 @@ rmw_take_loaned_message_with_info(
 /// Return a loaned ROS message previously taken from a subscription.
 /**
  * Tells the middleware that previously loaned ROS message is no longer needed by the caller.
- * Ownership of the ROS message is given back to the middleware.
+ * If this function fails early due to a logical error, such as an invalid argument,
+ * the loaned ROS message will be left unchanged.
+ * Otherwise, ownership of the ROS message will be given back to the middleware.
  * It is up to the middleware what will be made of the returned ROS message.
  * It is undefined behavior to use a loaned ROS message after returning it.
  *
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -840,7 +840,7 @@ rmw_subscription_get_actual_qos(
 
 /// Take an incoming ROS message.
 /**
- * Take a ROS message received by the given subscription, removing it from internal queues.
+ * Take a ROS message already received by the given subscription, removing it from internal queues.
  * This function will succeed even if no ROS message was received, but `taken` will be false.
  *
  * \remarks The same ROS message cannot be taken twice.
@@ -855,6 +855,14 @@ rmw_subscription_get_actual_qos(
  * Lock-Free          | Maybe [1]
  *
  * <i>[1] implementation defined, check implementation documentation.</i>
+ *
+ * \par Runtime behavior
+ *   To take a ROS message is a synchronous operation.
+ *   It is also non-blocking, to the extent it will not wait for new ROS messages to arrive,
+ *   but it is not guaranteed to be lock-free.
+ *   Generally speaking, implementations may synchronize access to internal resources using
+ *   locks but are not allowed to wait for events with no guaranteed time bound (barring
+ *   the effects of starvation due to OS scheduling).
  *
  * \par Memory allocation
  *   It is implementation defined whether memory will be allocated on take or not.
@@ -928,6 +936,14 @@ rmw_take(
  *
  * <i>[1] implementation defined, check implementation documentation.</i>
  *
+ * \par Runtime behavior
+ *   To take a ROS message with its metadata is a synchronous operation.
+ *   It is also non-blocking, to the extent it will not wait for new ROS messages to arrive,
+ *   but it is not guaranteed to be lock-free.
+ *   Generally speaking, implementations may synchronize access to internal resources using
+ *   locks but are not allowed to wait for events with no guaranteed time bound (barring
+ *   the effects of starvation due to OS scheduling).
+ *
  * \par Memory allocation
  *   It is implementation defined whether memory will be allocated on take or not.
  *   For instance, implementations that deserialize ROS messages received over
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
...
+ *   A subscription allocation, if provided, may or may not be used.
+ *   Check the implementation documentation to learn about memory allocation
+ *   guarantees when taking ROS messages with and without subscription allocations.
+ *
+ * \par Thread-safety
+ *   Subscriptions are thread-safe objects, and so are all operations on them except for
+ *   finalization.
+ *   Therefore, it is safe to take from the same subscription concurrently.
+ *   However, when taking regular ROS messages:
+ *   - Access to ROS messages is not synchronized.
```
ivanpauno, line 873: nit:
```suggestion
 *   - Access to the same ROS message is not synchronized.
```

I find that less confusing, but maybe it's just that I'm reading it incorrectly

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
...
+ *     It is not safe to read or write `ros_message` while rmw_take() uses it.
+ *   - Access to primitive data-types is not synchronized.
+ *     It is not safe to read or write `taken` while rmw_take() uses it.
```
ivanpauno, line 876: isn't this the same that the above item?

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
...
+ *     thread-safety.
+ *
+ * \pre Given `subscription` must be a valid publisher, as returned by rmw_create_subscription().
+ * \pre Given `ros_message` must be a valid message, whose type matches the message type support
+ *   registered with the `subscription` on creation.
+ * \pre If not NULL, given `allocation` must be a valid subscription allocation initialized
+ *   with rmw_subscription_allocation_init() with a message type support that matches the
+ *   one registered with the `subscription` on creation.
+ * \post If rmw_take() fails or `taken` is false, `ros_message` will be left in a valid
+ *   but unknown state.
```
ivanpauno, line 888: In the case of `taken=false`, I think the passed message should not be modified.
For failures, I'm not sure if we can ensure that ... maybe the deserialization failed in the middle of the process and left the object in an invalid expect (though ensuring that the message is in an valid undefined state would be nice, it's maybe a harder requirement to achieve).

rmw/include/rmw/rmw.h
```diff
@@ -858,16 +909,67 @@ rmw_take(
   bool * taken,
   rmw_subscription_allocation_t * allocation);
 
-/// Take an incoming message from a subscription with additional metadata.
+/// Take an incoming ROS message with its metadata.
```
ivanpauno, line 912: same comments that in `rmw_take`

rmw/include/rmw/rmw.h
hidmic: Yeah, I meant it in that sense. To say `things` and to say `the same thing` sounds equivalent to me, but I may be wrong.

rmw/include/rmw/rmw.h
hidmic: Which item? The one above talks about `ros_message`, the one here talks about `taken`. I wanted to be exhaustive, but perhaps we can shorten these to be less verbose.

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
   const rmw_subscription_t * subscription,
   rmw_qos_profile_t * qos);
 
-/// Take an incoming message from a subscription.
+/// Take an incoming ROS message.
 /**
- * Take an incoming ROS message from a given subscription.
+ * Read a ROS message received by the given subscription, removing it from internal queues.
```
wjwwood, line 843:
```suggestion
 * Take a ROS message received by the given subscription, removing it from internal queues.
```

Read means something else in the context of DDS, so it's better we don't mix these words up.

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
...
+ * This function will succeed even if no ROS message was received.
```
wjwwood, line 844:
```suggestion
 * This function will succeed even if no ROS message was received, but taken will be false.
```

wjwwood: I don't see any reason to say this without mentioning the way someone can know this occurred at the same time.

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
...
- * \param[out] ros_message The ROS message data on success.
- * \param[out] taken Boolean flag indicating if a message was taken or not.
- * \param[in] allocation Preallocated buffer to use (may be NULL).
+ * \remarks The same ROS message cannot be taken twice.
+ *   Callers do not have to deal with duplicates.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Maybe
```
wjwwood, line 852: Why does this one not have `[1]` as well?

rmw/include/rmw/rmw.h
ivanpauno: Ah, I didn't understand that `primitive data-types` was referring to `taken` :joy:.
Yeah, maybe just listing all arguments that are not synchronized in the same item would be ok.

rmw/include/rmw/rmw.h
wjwwood: Well `things` implies a group, whereas `the same thing` implies a specific instance.

I'd say the suggested change is worth doing. Or you could say:

```suggestion
 *   - Access to the ROS message is not synchronized.
```

Because in this case there's only one message interacting with this function.

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,66 @@ rmw_subscription_get_actual_qos(
...
+ * \par Thread-safety
+ *   Subscriptions are thread-safe objects, and so are all operations on them except for
+ *   finalization.
+ *   Therefore, it is safe to take from the same subscription concurrently.
+ *   However, when taking regular ROS messages:
+ *   - Access to ROS messages is not synchronized.
+ *     It is not safe to read or write `ros_message` while rmw_take() uses it.
+ *   - Access to primitive data-types is not synchronized.
+ *     It is not safe to read or write `taken` while rmw_take() uses it.
+ *   - Access to subscription allocations may be synchronized, but it is not required to.
```
wjwwood, line 877: I would go so far as to say "assume it is not thread-safe to use them concurrently" because the standard implementation where resources are used during take/publish cannot be protected easily.

wjwwood: Unless specifically stated otherwise by the middleware.

rmw/include/rmw/rmw.h
```diff
@@ -880,31 +982,90 @@ rmw_take_with_info(
   rmw_message_info_t * message_info,
   rmw_subscription_allocation_t * allocation);
 
-/// Take multiple incoming messages from a subscription with additional metadata.
+/// Take multiple incoming ROS messages with their metadata.
```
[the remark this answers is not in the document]
ivanpauno: > Should taking a sequence of message be synchronized with taking single messages?

IMO yes.
If both are called concurrently, I would expect:
- no segfaults
- `rmw_take_sequence()` returns N consecutive messages from the internal queue (i.e. `rmw_take_with_info` cannot take one "in the middle" while `rmw_take_sequence` is still running).

The first item is the most important one.

rmw/include/rmw/rmw.h
wjwwood: I agree, if taken is false the message should not have been modified at all.

hidmic: > In the case of taken=false, I think the passed message should not be modified.

So, the rationale here is that requiring implementations to not modify the given message when things go south adds little value (if any), and may be difficult (on failure) or inconvenient (on success) to attain. If we enforce the message to not be modified, that means some RMW implementation can't deserialize directly into the message without intermediate buffers.

> maybe the deserialization failed in the middle of the process and left the object in an invalid expect

Hmm... if the ROS message is not left in a valid state then finalization is UB. I think we have to ensure that ROS messages are not corrupted, no matter how hard. We cannot allow `take` to corrupt, say, `std::vector` and `std:string` fields in C++ messages. Fields may end up empty, or half populated, but they must remain valid or the process may fall apart on ROS message destruction.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
wjwwood: I think they should be safe to call at the same time.

It should be the same as calling `rmw_take_with_info()` in a loop, except more efficient because there are less function calls and the middleware can batch operations for efficiency.

rmw/include/rmw/rmw.h
```diff
@@ -880,31 +982,90 @@ rmw_take_with_info(
...
 /**
- * Take a sequence of ROS messgages from a given subscription.
+ * Read a sequence of ROS messages received by the given subscription, removing
```
wjwwood, line 987:
```suggestion
 * Take a sequence of ROS messages received by the given subscription, removing
```

rmw/include/rmw/rmw.h
```diff
@@ -918,21 +1079,75 @@ rmw_take_sequence(
   size_t * taken,
   rmw_subscription_allocation_t * allocation);
 
-/// Take a message without deserializing it.
+/// Take an incoming ROS message without deserializing it.
```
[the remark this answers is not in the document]
wjwwood: No, I think they would need to either raise unimplemented here, or it would need to serialize it before returning it.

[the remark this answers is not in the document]
wjwwood: So in that line of thinking maybe this should be:

```suggestion
/// Take an incoming ROS message as serialized data.
```

Or something similar.

rmw/include/rmw/rmw.h
```diff
@@ -918,21 +1079,75 @@ rmw_take_sequence(
...
- * The subscriber has to be registered for a specific type. But instead of receiving
- * the message as its corresponding message type, it is taken as a byte stream.
- * If needed, this byte stream can then be deserialized in a ROS message with a call to
- * rmw_deserialize.
- *
- * \param[in] subscription Subscription object to take from.
- * \param[out] serialized_message The destination in which to store the serialized message.
- * \param[out] taken Boolean flag indicating if a message was taken or not.
- * \param[in] allocation Preallocated buffer to use (may be NULL).
+ * Read a ROS message received by the given subscription, removing it from internal queues.
```
wjwwood, line 1084: I'm not gonna point all these out I think, but we should explicitly avoid read in the take methods.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
hidmic: This was settled on #270.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
hidmic: No segfaults is a given. But this:

> `rmw_take_sequence()` returns N consecutive messages from the internal queue

is not the same as this:

> It should be the same as calling `rmw_take_with_info()` in a loop,

There's no guarantee that you'll get N consecutive messages if you just run take in a loop.

I think that not having that guarantee reduces `rmw_take_sequence()` value somewhat, but I don't mind if we relax it.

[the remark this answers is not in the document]
hidmic: Hmm wait, consecutive != contiguous. Which one did you mean @ivanpauno ?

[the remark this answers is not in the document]
wjwwood: Oh you're right, I meant calling it in a loop while preventing new data from coming in, which is an important detail.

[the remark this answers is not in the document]
wjwwood: I'd say consecutive. Contiguous is odd to use here because the user has no idea how the data is laid out inside the sequence only that they can be accessed in the order presented.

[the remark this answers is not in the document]
ivanpauno: > Hmm wait, consecutive != contiguous. Which one did you mean @ivanpauno ?

with "contiguous": do you mean they are stored in memory one after the other?
I didn't meant that, but I meant "calling it in a loop while preventing new data from coming in" as @wjwwood said.

rmw/include/rmw/rmw.h
hidmic: Fair enough.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
hidmic: I'll go with `as a byte stream`, to use the same terminology everywhere.

rmw/include/rmw/rmw.h
hidmic: Yeah, I took a look at [RTI's docs](https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/AccessingData_Samples_ReadTake.htm) for a good definition of `take` (to avoid recursive definitions) and it's essentially a destructive read. But I'm fine with keeping `take` too.

rmw/include/rmw/rmw.h
hidmic: I didn't add small notes for attributes that have their own section. I didn't find a nice way of hyperlinking them to those sections though.

rmw/include/rmw/rmw.h
hidmic: Alright, we can rephrase it so that _not safe_ is the default.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
hidmic: Ok, for a moment I was unsure consecutive meant what I thought it meant :joy: 

Alright, so we agree it should be consecutive (i.e. one after the other with no interruption). So that if we have 10 messages in the queue, and two threads each taking a sequence of 5 messages, one thread will take the first 5 messages and the other will take the last 5 messages (instead of some random but ordered combination like 0, 1, 2, 6, 7 and 3, 4, 5, 8, 9).

When you say no new data, I assume you mean it should only pay attention to what's on the queue at the very beginning.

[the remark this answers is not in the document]
hidmic: > Alright, so we agree it should be consecutive 

BTW it looks like DDS (well, RTI Connext at least) has a few QoS policies to control this: [PRESENTATION](https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/PRESENTATION_QosPolicy.htm#sending_2410472787_1237370) and [DESTINATION_ORDER](https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/DESTINATION_ORDER_QosPolicy.htm#sending_2410472787_644578) QoS policies.

rmw/include/rmw/rmw.h
ivanpauno: > So, the rationale here is that requiring implementations to not modify the given message when things go south adds little value

Taken being "false" is normal, it only means you have called take with no data available.
That can happen because of race conditions between two takes, a subscription spuriously marked available by a wait set, or because the data is not more there due to some QoS condition.

> Hmm... if the ROS message is not left in a valid state then finalization is UB

Ah, you're right!

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
ivanpauno: > BTW it looks like DDS (well, RTI Connext at least) has a few QoS policies to control this: PRESENTATION and DESTINATION_ORDER QoS policies.

Yeah, I only meant that `rmw_take_sequence` ordering should match the ordering on the internal queue.
IIUC, by default each data reader preserves ordering of the received messages, but that can be different to the order of publication.

If we want to support different "presenstation/ordering" guarantees, we could add that to the QoS profile.
But that's a different issue, in all cases the data taken by `rmw_take_sequence` should be "consecutive" (i.e. preserve the order of the data in the reader queue, and not be interrupted by another `rmw_take/rmw_take_sequence`).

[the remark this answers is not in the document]
hidmic: > If we want to support different "presenstation/ordering" guarantees, we could add that to the QoS profile.
> But that's a different issue

Absolutely/ I don't care what DDS does or can do, only what ROS 2 is supposed to do.

> in all cases the data taken by rmw_take_sequence should be "consecutive" (i.e. preserve the order of the data in the reader queue, and not be interrupted by another rmw_take/rmw_take_sequence).

Agreed. @wjwwood do you agree ?

rmw/include/rmw/rmw.h
hidmic: > Taken being "false" is normal, it only means you have called take with no data available.

Alright, 2 vs 1 :sweat_smile: 

I'll constrain the spec.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
wjwwood: That's right. It doesn't matter how DDS chooses to present the data, but it is consecutive as far as the user is concerned. It doesn't matter if it is contiguous or not and it should not be mutate while taking the sequence.

In the future we could provide the ability to control the presentation order, but for now, we could just say "in the order they are received", which is the default for DDS and typical for most communication systems.

hidmic:
Oof, this is getting big. On top of what we discussed, I added a few more clarifications as to what to expect when a logical/runtime error occurs.

hidmic:
Another thing we should also document is if `take` (and `publish`) calls are: blocking or not, synchronous or not.

rmw/include/rmw/rmw.h
wjwwood: It is, but in the future we will most likely have an `rmw_read()` as well and using the term `read` here is going to be confusing.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
wjwwood: > When you say no new data, I assume you mean it should only pay attention to what's on the queue at the very beginning.

Yeah, I mean the work needed to take out the sequence is synchronized with adding/removing data from the history due to QoS or new data from the wire.

ivanpauno:
>  blocking or not, synchronous or not. OR leave that door open.

`publish` depends on the rmw implementation now.

I think `take` is always non-blocking and synchronous.
Considering how we use it, I think that that should be an API requirement.

rmw/include/rmw/rmw.h
```diff
@@ -838,15 +838,70 @@ rmw_subscription_get_actual_qos(
...
+ *     It is not safe to read or write `ros_message` while rmw_take() uses it.
+ *   - Access to primitive data-types is not synchronized.
+ *     It is not safe to read or write `taken` while rmw_take() uses it.
+ *   - Access to subscription allocations is not synchronized, unless specifically stated
+ *     otherwise by the implementation.
+ *     Thus, it is generally not safe to read or write `allocation` while rmw_take() uses it.
+ *     Check the implementation documentation to learn about subscription allocations'
+ *     thread-safety.
+ *
+ * \pre Given `subscription` must be a valid publisher, as returned by rmw_create_subscription().
```
ivanpauno, line 883:
```suggestion
 * \pre Given `subscription` must be a valid subscription, as returned by rmw_create_subscription().
```

hidmic: Whoops.

hidmic:
> I think `take` is always non-blocking and synchronous.

I agree with it being synchronous, but non-blocking is a tricky one. Well, it depends on how you define it. `take` will not wait for the queue to fill up, that for sure, and the documentation already says so. But it may have to contend with other threads for the reader queue (or the message loaning pool if any). Presumably for a limited amount of time, but the thread will get re-scheduled anyways.

I'd say we allow implementations to wait on locks, but we don't allow them to wait for external events (like a message arriving, or an empty message loaning pool getting refilled). Some implementation may choose to go full non-blocking, but that's a bit too much for all, I think.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/271/commits/6a711a56245d791dfc6b3c7eafc5d33557581118.

ivanpauno:
> I agree with it being synchronous, but non-blocking is a tricky one. Well, it depends on how you define it. take will not wait for the queue to fill up, that for sure, and the documentation already says so. But it may have to contend with other threads for the reader queue (or the message loaning pool if any). Presumably for a limited amount of time, but the thread will get re-scheduled anyways.

I think that non-blocking isn't the same as lock-free, but maybe I'm wrong.
But yeah, I agree with your definition above.

ivanpauno (approved):
lgtm, great work @hidmic!!!

Maybe in the future, we can only have a single `Thread-safety`, `Memory-Allocation`, `Runtime-Behavior` paragraph for the whole "take data" API, and link to that single paragraph from the documentation of each function.
I'm not quite sure on how to do that in doxygen.

rmw/include/rmw/rmw.h
hidmic: Whoops! See https://github.com/ros2/rmw/pull/271/commits/005ba4062a7508129ba0d4bcc592b6160684e15a.

ivanpauno (approved):

hidmic:
Alright, going in!

Issue #264, opened 2020-08-24
Push notification on publisher/subscription match


hidmic:
#### Feature description

Currently, the API does not provide (or documents) a generic mechanism to wait for publishers (subscriptions) to match subscriptions (publishers) -- only the current match count is available for query. Thus, code that monitors these counts has to sleep and retry. That's inefficient and potentially brittle. It'd be nice to enable such waits.

#### Implementation considerations

It seems to me that a new pub/sub event would be the simplest solution. All Tier 1 RMW implementations have a form of pub/sub matching listener that can be leveraged:

- eProsima Fast-DDS has it, and both `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp` already use it (see [here](https://github.com/ros2/rmw_fastrtps/blob/0959c84f9cca4b2005c9dffe7e18ec71656e5e74/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp#L66) and [here](https://github.com/ros2/rmw_fastrtps/blob/0959c84f9cca4b2005c9dffe7e18ec71656e5e74/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp#L69)).
- RTI Connext has it, and `rmw_connext_cpp` already uses it (see [here](https://github.com/ros2/rmw_connext/blob/2c3f725b6050186e7a215d0cd13634c084e05571/rmw_connext_cpp/src/connext_static_publisher_info.hpp#L61) and [here](https://github.com/ros2/rmw_connext/blob/2c3f725b6050186e7a215d0cd13634c084e05571/rmw_connext_cpp/src/connext_static_subscriber_info.hpp#L57)).
- ADLINK CycloneDDS has it, but `rmw_cyclonedds_cpp` doesn't use it yet (see [here](https://github.com/eclipse-cyclonedds/cyclonedds/blob/623b35d83b9d061187c816f37f4a92798f3b35d1/src/core/ddsc/include/dds/ddsc/dds_public_listener.h#L195-L203)). 

[...]

PR #272, opened 2020-09-11
Update graph API documentation.

hidmic:
Precisely what the title says. First pass.

### DIFF
--- a/rmw/include/rmw/get_node_info_and_types.h
+++ b/rmw/include/rmw/get_node_info_and_types.h
@@ -25,10 +25,10 @@ extern "C"
 #include "rmw/types.h"
 #include "rmw/visibility_control.h"
 
-/// Return all topics known to be subscribed by a given node, along with their types.
+/// Return all topic names and types for which a given remote node has subscriptions.
 /**
- * This function returns an array of topic names and types known to be subscribed by a
- * given remote node, as discovered so far by the given local node.
+ * This function returns an array of topic names and types for which a given remote
+ * node has subscriptions, as discovered so far by the given local node.
  *
  * <hr>
  * Attribute          | Adherence
@@ -48,7 +48,7 @@ extern "C"
  *
  * \par Thread-safety
  *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
- *   Therefore, it is to query the ROS graph using the same node concurrently.
+ *   Therefore, it is safe to query the ROS graph using the same node concurrently.
  *   However, when querying subscribed topic names and types:
  *   - Access to the array of names and types is not synchronized.
  *     It is not safe to read or write `topic_names_and_types`
@@ -56,7 +56,7 @@ extern "C"
  *   - Access to node name and namespace is read-only but it is not synchronized.
  *     Concurrent `node_name` and `node_namespace` reads are safe, but concurrent reads and
  *     writes are not.
- *   - Allocators are generally thread-safe objects, but the given `allocator` may not be.
+ *   - The default allocators are thread-safe objects, but any custom `allocator` may not be.
  *     Check your allocator documentation for further reference.
  *
  * \pre Given `node` must be a valid node handle, as returned by rmw_create_node().
@@ -69,8 +69,9 @@ extern "C"
  * \param[in] node_namespace Namespace of the remote node to get information for.
  * \param[in] no_demangle Whether to demangle all topic names following ROS conventions or not.
  * \param[out] topic_names_and_types Array of topic names and types the remote node has created
- *   a subscription for, populated on success.
- *   It is up to the caller to finalize this array later on, using rmw_names_and_types_fini().
+ *   a subscription for, populated on success but left unchanged on failure.
+ *   If populated, it is up to the caller to finalize this array later on
+ *   using rmw_names_and_types_fini().
  * \return `RMW_RET_OK` if the query was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `node` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `allocator` is not valid,
@@ -99,10 +100,10 @@ rmw_get_subscriber_names_and_types_by_node(
   bool no_demangle,
   rmw_names_and_types_t * topic_names_and_types);
 
-/// Return all topics known to be published by a given node, along with their types.
+/// Return all topic names and types for which a given remote node has publishers.
 /**
- * This function returns an array of topic names and types known to be published by a
- * given remote node, as discovered so far by the given local node.
+ * This function returns an array of topic names and types for which a given remote
+ * node has created publishers, as discovered so far by the given local node.
  *
  * <hr>
  * Attribute          | Adherence
@@ -122,7 +123,7 @@ rmw_get_subscriber_names_and_types_by_node(
  *
  * \par Thread-safety
  *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
- *   Therefore, it is to query the ROS graph using the same node concurrently.
+ *   Therefore, it is safe to query the ROS graph using the same node concurrently.
  *   However, when querying published topic names and types:
  *   - Access to the array of names and types is not synchronized.
  *     It is not safe to read or write `topic_names_and_types`
@@ -130,7 +131,7 @@ rmw_get_subscriber_names_and_types_by_node(
  *   - Access to node name and namespace is read-only but it is not synchronized.
  *     Concurrent `node_name` and `node_namespace` reads are safe, but concurrent reads and
  *     writes are not.
- *   - Allocators are generally thread-safe objects, but the given `allocator` may not be.
+ *   - The default allocators are thread-safe objects, but any custom `allocator` may not be.
  *     Check your allocator documentation for further reference.
  *
  * \pre Given `node` must be a valid node handle, as returned by rmw_create_node().
@@ -143,8 +144,9 @@ rmw_get_subscriber_names_and_types_by_node(
  * \param[in] node_namespace Namespace of the remote node to get information for.
  * \param[in] no_demangle Whether to demangle all topic names following ROS conventions or not.
  * \param[out] topic_names_and_types Array of topic names and types the remote node has created
- *   a publisher for, populated on success.
- *   It is up to the caller to finalize this array later on, using rmw_names_and_types_fini().
+ *   a publisher for, populated on success but left unchanged on failure.
+ *   If populated, it is up to the caller to finalize this array later on
+ *   using rmw_names_and_types_fini().
  * \return `RMW_RET_OK` if the query was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `node` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `allocator` is not valid,
@@ -173,11 +175,10 @@ rmw_get_publisher_names_and_types_by_node(
   bool no_demangle,
   rmw_names_and_types_t * topic_names_and_types);
 
-/// Return all services known to be served by a given node, along with their types.
+/// Return all service names and types for which a given remote node has servers.
 /**
- * This function returns an array of service names and types known to be served by
- * a given remote node, as discovered so far by the given local node.
- * A node serving a service means it created a service server for it.
+ * This function returns an array of service names and types for which a given remote
+ * node has servers, as discovered so far by the given local node.
  *
  * <hr>
  * Attribute          | Adherence
@@ -197,7 +198,7 @@ rmw_get_publisher_names_and_types_by_node(
  *
  * \par Thread-safety
  *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
- *   Therefore, it is to query the ROS graph using the same node concurrently.
+ *   Therefore, it is safe to query the ROS graph using the same node concurrently.
  *   However, when querying served service names and types:
  *   - Access to the array of names and types is not synchronized.
  *     It is not safe to read or write `service_names_and_types`
@@ -205,7 +206,7 @@ rmw_get_publisher_names_and_types_by_node(
  *   - Access to node name and namespace is read-only but it is not synchronized.
  *     Concurrent `node_name` and `node_namespace` reads are safe, but concurrent reads and
  *     writes are not.
- *   - Allocators are generally thread-safe objects, but the given `allocator` may not be.
+ *   - The default allocators are thread-safe objects, but any custom `allocator` may not be.
  *     Check your allocator documentation for further reference.
  *
  * \param[in] node Handle to local node to query the ROS graph.
@@ -213,8 +214,9 @@ rmw_get_publisher_names_and_types_by_node(
  * \param[in] node_namespace Namespace of the remote node to get information for.
  * \param[in] no_demangle Whether to demangle all topic names following ROS conventions or not.
  * \param[out] service_names_and_types Array of service names and types the remote node has
- *   created a service server for, populated on success.
- *   It is up to the caller to finalize this array later on, using rmw_names_and_types_fini().
+ *   created a service server for, populated on success but left unchanged on failure.
+ *   If populated, it is up to the caller to finalize this array later on
+ *   using rmw_names_and_types_fini().
  * \return `RMW_RET_OK` if the query was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `node` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `allocator` is not valid,
@@ -242,11 +244,10 @@ rmw_get_service_names_and_types_by_node(
   const char * node_namespace,
   rmw_names_and_types_t * service_names_and_types);
 
-/// Return all services known to be used by a given node, along with their types.
+/// Return all service names and types for which a given remote node has clients.
 /**
- * This function returns an array of service names and types known to be served by
- * a given remote node, as discovered so far by the given local node.
- * A node using a service means it created a service client for it.
+ * This function returns an array of service names and types for which a given remote
+ * node has clients, as discovered so far by the given local node.
  *
  * <hr>
  * Attribute          | Adherence
@@ -266,15 +267,15 @@ rmw_get_service_names_and_types_by_node(
  *
  * \par Thread-safety
  *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
- *   Therefore, it is to query the ROS graph using the same node concurrently.
+ *   Therefore, it is safe to query the ROS graph using the same node concurrently.
  *   However, when querying served service names and types:
  *   - Access to the array of names and types is not synchronized.
  *     It is not safe to read or write `service_names_and_types`
  *     while rmw_get_client_names_and_types_by_node() uses it.
- *   - Access to node name and namespace is read-only but it is not synchronized.
+ *   - Access to C-style string arguments is read-only but it is not synchronized.
  *     Concurrent `node_name` and `node_namespace` reads are safe, but concurrent reads and
  *     writes are not.
- *   - Allocators are generally thread-safe objects, but the given `allocator` may not be.
+ *   - The default allocators are thread-safe objects, but any custom `allocator` may not be.
  *     Check your allocator documentation for further reference.
  *
  * \param[in] node Handle to local node to query the ROS graph.
@@ -282,8 +283,9 @@ rmw_get_service_names_and_types_by_node(
  * \param[in] node_namespace Namespace of the remote node to get information for.
  * \param[in] no_demangle Whether to demangle all topic names following ROS conventions or not.
  * \param[out] service_names_and_types Array of service names and types the remote node has
- *   created a service client for, populated on success.
- *   It is up to the caller to finalize this array later on, using rmw_names_and_types_fini().
+ *   created a service client for, populated on success but left unchanged on failure.
+ *   If populated, it is up to the caller to finalize this array later on
+ *   using rmw_names_and_types_fini().
  * \return `RMW_RET_OK` if the query was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `node` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `allocator` is not valid,
--- a/rmw/include/rmw/get_node_info_and_types.h
+++ b/rmw/include/rmw/get_node_info_and_types.h
@@ -17,7 +17,7 @@
 
 #ifdef __cplusplus
 extern "C"
-
+{
 #endif
 
 #include "rmw/macros.h"
--- a/rmw/include/rmw/get_node_info_and_types.h
+++ b/rmw/include/rmw/get_node_info_and_types.h
@@ -63,7 +63,7 @@ extern "C"
  * \pre Given `topic_names_and_types` must be a zero-initialized array of names and types,
  *   as returned by rmw_get_zero_initialized_names_and_types().
  *
- * \param[in] node Handle to local node to query the ROS graph.
+ * \param[in] node Local node to query the ROS graph.
  * \param[in] allocator Allocator to be used when populating the `topic_names_and_types` array.
  * \param[in] node_name Name of the remote node to get information for.
  * \param[in] node_namespace Namespace of the remote node to get information for.
@@ -138,7 +138,7 @@ rmw_get_subscriber_names_and_types_by_node(
  * \pre Given `topic_names_and_types` must be a zero-initialized array of names and types,
  *   as returned by rmw_get_zero_initialized_names_and_types().
  *
- * \param[in] node Handle to local node to query the ROS graph.
+ * \param[in] Local node to query the ROS graph.
  * \param[in] allocator Allocator to be used when populating the `topic_names_and_types` array.
  * \param[in] node_name Name of the remote node to get information for.
  * \param[in] node_namespace Namespace of the remote node to get information for.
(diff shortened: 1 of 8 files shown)


rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -17,36 +17,73 @@
...
+ *
+ * \par Runtime behavior
+ *   To query the ROS graph is a synchronous operation.
+ *   It is also non-blocking, but it is not guaranteed to be lock-free.
+ *   Generally speaking, implementations may synchronize access to internal resources using
+ *   locks but are not allowed to wait for events with no guaranteed time bound (barring
+ *   the effects of starvation due to OS scheduling).
+ *
+ * \par Thread-safety
+ *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
```
hidmic, line 50: We need to come up with a way to document object thread-safety and not just that of functions. This is way more duplication I'm comfortable with.

rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -17,36 +17,73 @@
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | Yes
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
+ * \par Runtime behavior
+ *   To query the ROS graph is a synchronous operation.
```
hidmic, line 43: We need to come up with a way to document whole features and abstractions like the ROS graph.

rmw/include/rmw/get_topic_names_and_types.h
```diff
@@ -25,49 +25,74 @@ extern "C"
...
+ * may be returned as `rt/foo`, and a non-ROS topic (i.e. a DDS topic) with a partition list
+ * `['foo', 'bar']` and topic `baz` may be returned as `foo/baz` (note that only the first
+ * partition is used but it is still concatenated to the topic).
  *
- * There may be some demangling that occurs when listing the topics from the
- * middleware implementation.
- * This is the mechanism by which this function can discriminate between ROS
- * topics, non-ROS topics, and topics which may be used to implement other
- * concepts like ROS Services.
+ * \sa <a href="http://design.ros2.org/articles/topic_and_service_names.html">Topic naming design</a>
```
hidmic, line 50: You can see the original documentation had two links to two separate sections. I simplified it to a single URL, but I'm not entirely sure we want to put references to design documents here.

rmw/include/rmw/rmw.h
```diff
@@ -200,20 +200,16 @@ rmw_node_assert_liveliness(const rmw_node_t * node);
...
- * The returned handle is made invalid if the node is destroyed or if
- * rmw_shutdown() is called.
- *
- * The guard condition will be triggered anytime a change to the ROS graph
- * occurs.
- * A ROS graph change includes things like (but not limited to) a new publisher
+ * The guard condition will be triggered anytime a change to the ROS graph occurs.
+ * A ROS graph change includes but it is not limited to a new publisher
  * advertises, a new subscription is created, a new service becomes available,
  * a subscription is canceled, etc.
```
hidmic, line 206: We need to sit and defined what a _ROS graph change_ is. Exhaustively. The few examples it gives don't even apply to current implementations.

rmw/include/rmw/rmw.h
```diff
@@ -200,20 +200,16 @@ rmw_node_assert_liveliness(const rmw_node_t * node);
...
  *
+ * The guard condition is owned and internally held by the `node`.
+ * It will be invalidated if `node` is finalized using rmw_destroy_node(),
+ * or if rmw_shutdown() is called on the context `node` was created in.
```
hidmic, line 210: Invalidation on node destruction is probably a given, but I know no implementation invalidates guard conditions on context shutdown. We need to sort out what _shutdown_ means to begin with.

rmw/include/rmw/topic_endpoint_info.h
```diff
@@ -24,8 +24,9 @@ extern "C"
 #include "rmw/types.h"
 #include "rmw/visibility_control.h"
```
hidmic, line 26: These APIs expect the same `allocator` to be passed around to not hit UB, presumably to spare storage. I wonder if these shouldn't be explicitly tagged for internal implementation use only.

hidmic: Also, documentation duplication makes me cry.

rmw/include/rmw/topic_endpoint_info_array.h
```diff
@@ -24,58 +24,81 @@ extern "C"
...
+ * \param[in] allocator Allocator to be used to populate `names_and_types`.
+ * \returns `RMW_RET_OK` if successful, or
+ * \returns `RMW_RET_INVALID_ARGUMENT` if `topic_endpoint_info_array` is NULL, or
+ * \returns `RMW_RET_INVALID_ARGUMENT` if `topic_endpoint_info_array` is not
+ *   a zero initialized array, or
+ * \returns `RMW_RET_INVALID_ARGUMENT` if `allocator` is invalid,
+ *   by rcutils_allocator_is_valid() definition, or
  * \returns `RMW_BAD_ALLOC` if memory allocation fails, or
  * \returns `RMW_RET_ERROR` when an unspecified error occurs.
+ * \remark This function sets the RMW error state on failure.
```
hidmic, line 101: This is actually nice. Unless we agree all APIs should set an error message on failure, it's useful.

rmw/include/rmw/names_and_types.h
```diff
@@ -79,23 +111,31 @@ rmw_names_and_types_init(
...
 
-/// Finalize a rmw_names_and_types_t object.
+/// Finalize an array of names and types.
 /**
- * The names_and_types_t objects are populated by one of the
- * rmw_get_*_names_and_types() functions.
- * During which memory is allocated to store the names and types.
- * This function will reclaim any resources within the object so it is safe
- * to destroy without leaking memory.
+ * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`,
```
clalancette, line 116: I'd rather this description start with what this function does and expects, rather than the unlikely case of failure.  That is, something like:

```
This function frees a string array for topics and names previously allocated by rmw_names_and_types_init().
```

rmw/include/rmw/names_and_types.h
```diff
@@ -79,23 +111,31 @@ rmw_names_and_types_init(
...
+ * ensues, leaving the given array unchanged.
+ * Otherwise, it will proceed despite errors, freeing as much resources as it can and zero
```
clalancette, line 118:
```suggestion
 * Otherwise, it will proceed despite errors, freeing as many resources as it can and zero
```

rmw/include/rmw/rmw.h
clalancette: Agreed.  This also seems like higher-level documentation that needs to live somewhere else (see my previous comment).

In the meantime, if the examples given here don't apply, should we remove them?

rmw/include/rmw/rmw.h
clalancette: This is another place for higher-level documentation.

rmw/include/rmw/rmw.h
```diff
@@ -1996,31 +1994,60 @@ rmw_wait(
   rmw_wait_set_t * wait_set,
   const rmw_time_t * wait_timeout);
 
-/// Return a list of node name and namespaces discovered via a node.
+/// Return all known nodes' names and namespaces.
```
clalancette, line 1997: This is grammatically correct, but hard to read (in my opinion).  Maybe:

```suggestion
/// Return the name and namespace of all known nodes.
```

rmw/include/rmw/rmw.h
```diff
@@ -2031,15 +2058,63 @@ rmw_get_node_names(
   rcutils_string_array_t * node_names,
   rcutils_string_array_t * node_namespaces);
 
-/// Return a list of node name and namespaces discovered via a node with its enclave.
+/// Return all known nodes' names, namespaces, and enclave names.
```
clalancette, line 2061:
```suggestion
/// Return the name, namespace, and enclave name of all known nodes.
```

rmw/include/rmw/rmw.h
```diff
@@ -2031,15 +2058,63 @@ rmw_get_node_names(
...
 /**
- * Similar to \ref rmw_get_node_names, but it also provides the enclave name.
+ * Similar to rmw_get_node_names(), but it also provides enclave names.
```
clalancette, line 2063: Not really a sentence.  Maybe:

```suggestion
 * This is similar to rmw_get_node_names(), but it also provides enclave names.
```

rmw/include/rmw/topic_endpoint_info.h
```diff
@@ -42,26 +43,44 @@ typedef struct RMW_PUBLIC_TYPE rmw_topic_endpoint_info_t
...
 rmw_get_zero_initialized_topic_endpoint_info(void);
 
-/// Finalize a rmw_topic_endpoint_info_t object.
+/// Finalize a topic endpoint info data structure.
 /**
- * The rmw_topic_endpoint_info_t struct has members which require memory to be allocated to them before
- * setting values.
- * This function reclaims any allocated resources within the object and zeroes out all other
- * members.
+ * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`,
```
clalancette, line 58: Again, would be nice to start here by describing what it does, and then talk about the internals.

rmw/include/rmw/topic_endpoint_info.h
```diff
@@ -42,26 +43,44 @@ typedef struct RMW_PUBLIC_TYPE rmw_topic_endpoint_info_t
...
+ * ensues, leaving the given data structure unchanged.
+ * Otherwise, it will proceed despite errors, freeing as much resources as it can and zero
```
clalancette, line 60:
```suggestion
 * Otherwise, it will proceed despite errors, freeing as many resources as it can and zero
```

clalancette (commented):
I've got a bunch of comments in here.

I think there are 2 major things that this has pointed out.

The first is that there is a lot of duplication in the comments.  Partly this is because the API could be improved.  One example is `rmw_get_node_names()` and `rmw_get_node_names_with_enclaves()`.  In my opinion, there is no reason to have both of those APIs.  Instead we could have `rmw_get_node_names` take an `enclave` argument, and don't attempt to fill in enclave information if it is NULL.  Partly the duplication is because we are missing some place to point to that has higher-level documentation about concepts.

The second major point is that we need a place to put higher-level documentation about concepts.  My first inclination would be to start putting some conceptual documents at https://index.ros.org/doc/ros2/Tutorials/ or http://docs.ros2.org/foxy/ , but I'm not sure.

I honestly think neither is blocking; I would go with duplication for now, and we can review both of them later.  But if we do that, then we should open up issues and/or add roadmap items to do so.

rmw/include/rmw/get_node_info_and_types.h
hidmic: > This is something of a philosophical debate, honestly. In some sense, there is no such thing as a thread-safe object; thread safety only comes into play when accessing data, and that can only be done through functions or methods.

Sure, but what is an object if not state (data) coupled with behavior (functions)? At any rate, I think https://github.com/ros2/ros2/issues/1004 would benefit from your POV.

> It seems to me that that can mostly be resolved by pointing the reader to the documentation about the node API, which presumably talks about that.

Do we have node API documentation? Or are you talking about something that we should have at some point?

rmw/include/rmw/get_node_info_and_types.h
hidmic: I'd be inclined towards having it here, in this repository, close to the API specifications that rely on it. External web sites can then consume that content.

rmw/include/rmw/get_topic_names_and_types.h
hidmic: > I honestly think that this section (and the linked design article) do a bit too much mixing of ROS concepts and DDS concepts.

I agree. At the same time, I'm not put off by an `rmw` API specification making an example out of another specification that applies to all `rmw` implementation that are based on DDS. Meaning, I can see the value. We often don't have examples of how our APIs are supposed to be used, which would be a great complement to the spec IMHO.

>  From my perspective, I would try to rewrite the above paragraph saying something like

I like the new phrasing, let's go with that.

rmw/include/rmw/rmw.h
hidmic: > This also seems like higher-level documentation that needs to live somewhere else

Agreed.

> In the meantime, if the examples given here don't apply, should we remove them?

Hmm, I think having just a few examples (or not having examples at all) is a problem. If we don't define what a ROS graph change is, then it's up to implementations to trigger (or not trigger) the graph guard condition when and where they see fit. That makes the graph guard condition impossible to test and couples user code to the implementation of choice (e.g. if you choose another implementation, the scenarios that used to trigger the graph guard condition may no longer apply).

rmw/include/rmw/rmw.h
hidmic: Even before that, there's [a discussion](https://github.com/ros2/rclcpp/issues/1139) that has to be settled. To leave this piece of documentation here means we're gearing towards a shutdowns with side-effects and not just a mere notification. I agree with that view, but I know there're arguments against it.

rmw/include/rmw/get_service_names_and_types.h
```diff
@@ -25,25 +25,56 @@ extern "C"
 #include "rmw/types.h"
 #include "rmw/visibility_control.h"
 
-/// Return a list of service names and their types.
+/// Return all known services and their types.
```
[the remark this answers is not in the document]
hidmic: If I'm not mistaken, this API gives you the services for which there are servers and/or clients, which is not the same as listing available known servers and clients.

rmw/include/rmw/names_and_types.h
hidmic: Sounds good.

rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -25,28 +25,65 @@ extern "C"
...
+ *   It is also non-blocking, but it is not guaranteed to be lock-free.
+ *   Generally speaking, implementations may synchronize access to internal resources using
+ *   locks but are not allowed to wait for events with no guaranteed time bound (barring
+ *   the effects of starvation due to OS scheduling).
+ *
+ * \par Thread-safety
+ *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
+ *   Therefore, it is to query the ROS graph using the same node concurrently.
```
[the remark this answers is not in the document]
hidmic: If you don't mind, I'll go with:
```suggestion
 *   Therefore, it is safe to query the ROS graph using the same node concurrently.
```
to stay consistent with the rest of the thread-safety documentation.

rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -25,28 +25,65 @@ extern "C"
...
+ *   However, when querying subscribed topic names and types:
+ *   - Access to the array of names and types is not synchronized.
+ *     It is not safe to read or write `topic_names_and_types`
+ *     while rmw_get_subscriber_names_and_types_by_node() uses it.
+ *   - Access to node name and namespace is read-only but it is not synchronized.
+ *     Concurrent `node_name` and `node_namespace` reads are safe, but concurrent reads and
+ *     writes are not.
+ *   - Allocators are generally thread-safe objects, but the given `allocator` may not be.
```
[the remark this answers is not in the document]
hidmic: Sounds good.

rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -25,28 +25,65 @@ extern "C"
...
+ * \pre Given `topic_names_and_types` must be a zero-initialized array of names and types,
+ *   as returned by rmw_get_zero_initialized_names_and_types().
+ *
+ * \param[in] node Handle to local node to query the ROS graph.
+ * \param[in] allocator Allocator to be used when populating the `topic_names_and_types` array.
+ * \param[in] node_name Name of the remote node to get information for.
+ * \param[in] node_namespace Namespace of the remote node to get information for.
+ * \param[in] no_demangle Whether to demangle all topic names following ROS conventions or not.
+ * \param[out] topic_names_and_types Array of topic names and types the remote node has created
+ *   a subscription for, populated on success.
```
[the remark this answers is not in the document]
hidmic: Fair point.

rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -99,27 +173,61 @@ rmw_get_publisher_names_and_types_by_node(
   bool no_demangle,
   rmw_names_and_types_t * topic_names_and_types);
 
-/// Return a list of service topic names and their types.
+/// Return all services known to be served by a given node, along with their types.
```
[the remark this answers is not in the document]
hidmic: Sounds good.

rmw/include/rmw/get_node_info_and_types.h
```diff
@@ -134,28 +242,61 @@ rmw_get_service_names_and_types_by_node(
   const char * node_namespace,
   rmw_names_and_types_t * service_names_and_types);
 
-/// Return a list of service client topic names and their types.
+/// Return all services known to be used by a given node, along with their types.
```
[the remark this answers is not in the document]
hidmic: Same as above.

rmw/include/rmw/get_topic_names_and_types.h
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/get_node_info_and_types.h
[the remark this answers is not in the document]
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/topic_endpoint_info.h
hidmic: Sound reasonable, see https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/topic_endpoint_info.h
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/get_node_info_and_types.h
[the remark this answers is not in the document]
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/get_node_info_and_types.h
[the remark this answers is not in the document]
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/get_service_names_and_types.h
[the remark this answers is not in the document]
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

rmw/include/rmw/names_and_types.h
hidmic: See https://github.com/ros2/rmw/pull/272/commits/cbabbacaa854d3087e67a57788b6eaf1333aed98.

hidmic:
Comments addressed, PTAL!

rmw/include/rmw/rmw.h
hidmic: This is one and the same with https://github.com/ros2/rmw/issues/264.

gbiggs:
@clalancette Regarding writing a higher level concepts document, "write a tutorial on how to write an RMW implementation" is one of the outputs we planned for the Zenoh work. I'm happy to help out with writing the higher-level concepts document, provided someone with more experience with the API is willing to review it in detail.

hidmic:
FYI I'll start writing tests and adapting implementation to comply with this. We can sort out where to put documentation later on.

rmw/include/rmw/get_node_info_and_types.h
wjwwood: I think here is fine, but it depends on the purpose. If it is to make this API clear, then here, but it's goal is to explain the concept to users, then higher up the stack is more appropriate, but some details will not, like the thread-safety of this API, as an example. Because higher level APIs may give different thread-safety guarantees and the users are more likely to use those interfaces.

rmw/include/rmw/get_node_info_and_types.h
wjwwood: We can have higher level sections in doxygen (using the file for example), or we can document the struct, like I do with context:

https://github.com/ros2/rcl/blob/8bcada7b884693644d7ba2e3bd2c22fd9541eba7/rcl/include/rcl/context.h#L43-L143

rmw/include/rmw/rmw.h
wjwwood: I actually think the graph change should be described here. And the high level docs can give some examples and reference this function for the exhaustive list.

rmw/include/rmw/rmw.h
wjwwood: The guard condition isn't invalidated when shutdown is called, I think. I think you cannot create new ones, but you can check and trigger and wait on existing ones, IIRC.

rmw/include/rmw/topic_endpoint_info.h
wjwwood: They really should store the allocator given during init and reuse it during fini rather than having the user pass it...

wjwwood: That is to say, I agree with you.

rmw/include/rmw/topic_endpoint_info_array.h
wjwwood: I think if a function returns an error code it should set the error state too. However, mentioning explicitly doesn't hurt anything, but is quite redundant.

wjwwood (commented):
lgtm, it's a great improvement!

I commented on most of the conversations, but you guys picked out everything I noticed already, so no extra stuff from me I think.

rmw/include/rmw/get_node_info_and_types.h
hidmic: > If it is to make this API clear, then here, but it's goal is to explain the concept to users, then higher up the stack is more appropriate

Hmm, higher up the stack where? IMHO we cannot use terms and concepts in API documentation without an explanation or a link to an explanation. Besides, isn't the ROS middleware interface the best place to introduce ROS middleware concepts (barring tutorials and conceptual overviews that we can have elsewhere for educational purposes)?

rmw/include/rmw/topic_endpoint_info_array.h
hidmic: In general I agree with you that the error state should be set consistently. But for some functions it may be convenient if it is not set e.g. those that perform sanity checks on behalf of another function. For now, for this, I'll keep the remarks as they were here already.

rmw/include/rmw/rmw.h
hidmic: Alright, I removed the bit talking about shutdown contexts in c92a73df1974cf9ff9f665155f1029199de64f9a. We can revisit if need be when we get to have that discussion.

rmw/include/rmw/rmw.h
hidmic: Done in c92a73df1974cf9ff9f665155f1029199de64f9a. If we agree, I'll open issues on each implementation to ensure they comply (when I or someone else has time to do it :sweat_smile:).

rmw/include/rmw/get_node_info_and_types.h
hidmic: Any objections to proceed without relocating documentation? For the sake of speed, and to keep it consistent with previous API spec updates that do things this way.

clalancette (approved):

rmw/include/rmw/get_node_info_and_types.h
wjwwood: Fine with me.

rmw/include/rmw/get_node_info_and_types.h
wjwwood: Yes, I was just saying that tutorials and conceptual guides (things actual users will need) shouldn't depend on this except obliquely.

rmw/include/rmw/topic_endpoint_info_array.h
wjwwood: Ok.

The alternative is that we could document when a function does not explicitly set the state.

But it's fine as-is for me.

wjwwood (approved):

hidmic:
Alright, going in!

PR #275, opened 2020-09-22
Update wait and wait sets' API documentation.

hidmic:
Precisely what the title says. It includes some remarks about how time is measured to document https://github.com/ros2/rcl/issues/687.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -1917,72 +1917,139 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_trigger_guard_condition(const rmw_guard_condition_t * guard_condition);
 
-/// Create a wait set to store conditions that the middleware will block on.
+/// Create a wait set to store conditions that the middleware can wait on.
 /**
  * This function can fail, and therefore return `NULL`, if:
  *   - context is `NULL`
- *   - context is invalid
+ *   - context is zero initialized, as provided by rmw_get_zero_initialized_context()
+ *   - context does not belong to this implementation i.e. does not have a matching
+ *     implementation identifier
  *   - memory allocation fails during wait set creation
  *   - an unspecified error occurs
  *
- * If `max_conditions` is `0`, the wait set can store an unbounded number of
- * conditions to wait on.
- * If `max_conditions` is greater than `0`, the number of conditions that can
- * be attached to the wait set is bounded at `max_conditions`.
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | Yes
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
  *
- * \param[in] context init context that this node should be associated with
+ * \par Thread-safety
+ *   Contexts are thread-safe objects, and so are all operations on them except for finalization.
+ *   Therefore, it is safe to create multiple wait sets in the same context concurrently.
+ *
+ * \pre Given `context` must be a valid context, initialized by rmw_init() or
+ *   otherwise zero initialized (or finalized).
+ *
+ * \param[in] context Context to associate the wait set with.
  * \param[in] max_conditions
- *   The maximum number of conditions that can be attached to the wait set.
- * \return A pointer to the created wait set, `NULL` if an error occurred.
+ *   The maximum number of conditions that can be attached to, and stored by, the wait set.
+ *   Can be set to zero (0) for the wait set to support an unbounded number of conditions.
+ * \return An rmw wait set, or `NULL` if an error occurred.
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_wait_set_t *
 rmw_create_wait_set(rmw_context_t * context, size_t max_conditions);
 
-/// Destroy and free memory of this wait_set
+/// Finalize a wait set.
 /**
- * \param[in] wait_set The wait_set object to destroy
- * \return RMW_RET_OK if successful, otherwise an appropriate error code
+ * This function will reclaim all associated resources, including the wait set handle.
+ * Use of a deallocated wait set handle is undefined behavior.
+ * This function will return early if a logical error, such as `RMW_RET_INVALID_ARGUMENT`
+ * or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, ensues, leaving the given wait set
+ * handle unchanged.
+ * Otherwise, it will proceed despite errors.
+ *
+ * <hr>
+ * Attribute          | Adherence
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -1917,72 +1917,139 @@ RMW_WARN_UNUSED
...
 
-/// Create a wait set to store conditions that the middleware will block on.
+/// Create a wait set to store conditions that the middleware can wait on.
 /**
  * This function can fail, and therefore return `NULL`, if:
  *   - context is `NULL`
- *   - context is invalid
+ *   - context is zero initialized, as provided by rmw_get_zero_initialized_context()
+ *   - context does not belong to this implementation i.e. does not have a matching
+ *     implementation identifier
```
hidmic, line 1926: What if the context is shutdown?

rmw/include/rmw/rmw.h
```diff
@@ -1917,72 +1917,139 @@ RMW_WARN_UNUSED
...
+ *   To wait is a reentrant procedure, but:
+ *   - It is not safe to use the same wait set to wait in two or more threads concurrently.
+ *   - It is not safe to wait for the same entity using different wait sets in two or
+ *     more threads concurrently.
+ *   - Access to the given timeout is read-only but it is not synchronized.
+ *     Concurrent `wait_timeout` reads are safe, but concurrent reads and writes are not.
+ *
+ * \pre Given `wait_set` must be a valid wait set, as returned by rmw_create_wait_set().
+ * \pre All given entities must be associated with nodes that, in turn, were registered
+ *   with the same context the given `wait_set` was registered with on creation.
```
hidmic, line 2030: This precondition is a bit of a presumption of mine. It makes sense to me though.

rmw/include/rmw/rmw.h
```diff
@@ -1917,72 +1917,139 @@ RMW_WARN_UNUSED
...
+ *
+ * \param[inout] subscriptions Array of subscriptions to wait on.
+ *   Can be `NULL` if there're no subscriptions to wait on.
```
clalancette, line 2033: `there're` isn't really used in written communication :).  Just expand to `there are` (same below).

clalancette (commented):
A few questions and changes that I have.

rmw/include/rmw/rmw.h
```diff
@@ -1917,72 +1917,139 @@ RMW_WARN_UNUSED
...
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
+ * \par Thread-safety
+ *   Contexts are thread-safe objects, and so are all operations on them except for finalization.
+ *   Therefore, it is safe to create multiple wait sets in the same context concurrently.
  *
- * \param[in] context init context that this node should be associated with
+ * \pre Given `context` must be a valid context, initialized by rmw_init() or
+ *   otherwise zero initialized (or finalized).
```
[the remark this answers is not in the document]
hidmic: So, the precondition is that the context must be in a valid state, where valid state means _zero initialized_, _initialized_, or _shutdown_. That doesn't mean the function will succeed, it just means that the function behavior is well defined. Does that make sense?

rmw/include/rmw/rmw.h
hidmic: @clalancette WDYT?

rmw/include/rmw/rmw.h
hidmic: @clalancette WDYT?

rmw/include/rmw/rmw.h
```diff
@@ -1917,72 +1917,139 @@ RMW_WARN_UNUSED
...
 rmw_wait_set_t *
 rmw_create_wait_set(rmw_context_t * context, size_t max_conditions);
 
-/// Destroy and free memory of this wait_set
+/// Finalize a wait set.
 /**
- * \param[in] wait_set The wait_set object to destroy
- * \return RMW_RET_OK if successful, otherwise an appropriate error code
+ * This function will reclaim all associated resources, including the wait set handle.
+ * Use of a deallocated wait set handle is undefined behavior.
```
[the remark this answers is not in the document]
hidmic: Sounds reasonable.

rmw/include/rmw/rmw.h
hidmic: I see. We'll keep that in mind from now on.

rmw/include/rmw/rmw.h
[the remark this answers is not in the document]
hidmic: Perhaps it's redundant. I'll backtrack. We can revisit in the future how to document this if need be.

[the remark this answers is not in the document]
clalancette: I see what you are saying.  Its the difference between the precondition, and whether the function will succeed.

I think your simplified version is good enough for now; as you say, we can always revisit if we really need to later.

rmw/include/rmw/rmw.h
clalancette: Sorry, it wasn't clear to me if that was a question to the reviewer or to yourself :).

It seems to me that if the context is shutdown, then calling this should fail.  At the moment, I can't think of a situation where you would want this to succeed (but maybe somebody else can).

The follow-up question is: what do the implementations currently do in that case?

hidmic: > It seems to me that if the context is shutdown, then calling this should fail.

I'd think the same.

> what do the implementations currently do in that case?

About the context being shutdown? Nothing :)

clalancette: In that case, we need to answer the question about whether this documentation is descriptive (that is, it describes what is already there), or prescriptive (that is, it is an ambition that we may not be currently achieving).

I'm content to just leave this out and leave it implementation-defined for now.  The next time we do a review of the API, we can bring the question up again.

clalancette (approved):

rmw/include/rmw/rmw.h
wjwwood: Summarizing my opinion from the thread about this issue, I think this should continue to work during shutdown and only the finalization of the context should cause this to fail. With the logic being that some small tasks may need to be done after shutdown but before quitting, and that might involve creating a wait set.

rmw/include/rmw/rmw.h
wjwwood: Yeah, it's definitely required, but I also don't know if it is being asserted anywhere.

wjwwood (approved):
lgtm as-is

rmw/include/rmw/rmw.h
hidmic: Then it makes sense to keep it as a precondition, so that users know that not satisfying its predicate means UB.

wjwwood (approved):

hidmic:
Alright, in the interest of time, I'll get this in. Tests will come soon after.

PR #276, opened 2020-09-22
Update service server/client creation/destruction API documentation.

hidmic:

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -1749,13 +1749,42 @@ rmw_return_loaned_message_from_subscription(
   const rmw_subscription_t * subscription,
   void * loaned_message);
 
-/// Create an rmw client to communicate with the specified service
+/// Create a service client that can send requests to a service server.
 /**
- * \param[in] node Handle to node with which to register this client
- * \param[in] type_support The type_support of this rosidl service
- * \param[in] service_name The name of the ROS 2 service to connect with
- * \param[in] qos_policies The QoS profile policies to utilize for this connection
- * \return The initialized client if successful, `nullptr` if not
+ * This function can fail, and therefore return `NULL`, if:
+ *   - `node` is `NULL`, or
+ *   - `node` does not belong to this implementation i.e. it does not have
+ *     a matching implementation identifier, or
+ *   - `type_support` is `NULL`, or
+ *   - `service_name` is `NULL`, or
+ *   - `service_name` is an empty string, or
+ *   - (if ROS namespace conventions apply) `service_name` is invalid by
+ *     rmw_validate_full_topic_name() definition, or
+ *   - `qos_profile` is `NULL`, or
+ *   - `qos_profile` has invalid or unknown policies, or
+ *   - memory allocation fails during service server creation, or
+ *   - an unspecified error occurs.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] rmw implementation defined, check the implementation documentation</i>
+ *
+ * \pre Given `node` must be a valid node, as returned by rmw_create_node().
+ * \pre Given `type_support` must be a valid `rosidl` service type support, as
+ *   returned by ROSIDL_GET_SRV_TYPE_SUPPORT().
+ *
+ * \param[in] node Node with which to register this service client.
+ * \param[in] type_support Type support of the service to be used.
+ * \param[in] service_name Name of the service to be used, often a fully qualified
+ *   service name unless `qos_profile` is configured to avoid ROS namespace conventions
+ *   i.e. to create a native service client.
+ * \param[in] qos_profile QoS policies for this service client's connections.
+ * \return rmw service client handle, or `NULL` if there was an error.
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -1749,13 +1749,42 @@ rmw_return_loaned_message_from_subscription(
...
+ *   - `qos_profile` is `NULL`, or
+ *   - `qos_profile` has invalid or unknown policies, or
+ *   - memory allocation fails during service client creation, or
+ *   - an unspecified error occurs.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
```
hidmic, line 1772: I'd expect `Yes` though. Anything arguments against?

wjwwood: I would say no? create publisher isn't: https://github.com/ros2/rmw/blob/f48ebcf0abc9e27f90733046bbffffb18d5e0638/rmw/include/rmw/rmw.h#L298

rmw/include/rmw/rmw.h
```diff
@@ -1809,13 +1863,42 @@ rmw_take_response(
   void * ros_response,
   bool * taken);
 
-/// Create an rmw service server that responds to requests
+/// Create a service server that can reply to client requests.
```
wjwwood, line 1866:
```suggestion
/// Create a service server that can receive and reply to client requests.
```

rmw/include/rmw/rmw.h
```diff
@@ -1749,13 +1749,42 @@ rmw_return_loaned_message_from_subscription(
   const rmw_subscription_t * subscription,
   void * loaned_message);
 
-/// Create an rmw client to communicate with the specified service
+/// Create a service client that can send requests to a service server.
```
wjwwood, line 1752:
```suggestion
/// Create a service client that can send requests to and receive replies from a service server.
```

wjwwood (approved):
Two small suggestions, then lgtm

rmw/include/rmw/rmw.h
hidmic: Yeah, I know it says it's not. I added it :)

But reflecting a bit on it, don't we assume it is everywhere else? `rclcpp` doesn't do much to ensure not two services get created concurrently (see [here](https://github.com/ros2/rclcpp/blob/974772e2abfdfdf7fa0ee92ed0a3577d11a0930b/rclcpp/include/rclcpp/service.hpp#L204-L209), no locks). Same for publishers. And looking at implementations, they do seem thread-safe (take that with a ~grain~ bag of salt, I haven't audited that code).

rmw/include/rmw/rmw.h
hidmic: Made yet another tweak. See https://github.com/ros2/rmw/pull/276/commits/7a37488e7f19efc941724bc2ceed6d60e20543cf.

rmw/include/rmw/rmw.h
hidmic: See https://github.com/ros2/rmw/pull/276/commits/7a37488e7f19efc941724bc2ceed6d60e20543cf.

rmw/include/rmw/rmw.h
wjwwood: Well, that might be a bug then? I'm hesitant to place a lot of thread-safety requirements on the rmw API, because it might make it more difficult to implement on various systems, and in scenarios like real-time systems where blocking is bad. That's the entire reason for mentioning locks and atomics in this stanza originally. For the major cases like `publish` and `take`, it is (imo) unavoidable to ask for it to be thread-safe and niche systems may choose to address that with polling or other lock-free operations/datastructures, but I don't really want to put that requirement in too many places. It's easier and perhaps more efficient to do the locking in the client library in most cases.

wjwwood (approved):

rmw/include/rmw/rmw.h
hidmic: Fair enough. By the same token, I wonder if even client libraries should be locking (by default it's fine, but in general it could limit usage and/or hurt performance).

clalancette (approved):

hidmic:
CI up to `test_rmw_implementation` and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12435/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7402/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10145/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12348/)

rmw/include/rmw/rmw.h
ivanpauno: `rclcpp` isn't mutexing the access to `rcl_node_t`, so either we should fix that or make these functions thread safe.

e.g.: weird race condition when registering the same type in rmw_connext https://github.com/ros2/rmw_connext/pull/442.
Maybe, I should have made access to `rcl_node_t` from `rclcpp` mutually exclusive instead of that.

ivanpauno: In fastrtps, the situation is a bit different:

https://github.com/ros2/rmw_fastrtps/blob/721591b4fd849e9b30374e1d6afe7b7db8c06874/rmw_fastrtps_cpp/src/publisher.cpp#L120-L131

Worst case, that [will log an error] because we're ignoring the return value of the "registerType" function.
But if we want to avoid that TOCTTOU race, mutexed access to the node will not solve the problem as different nodes share the same participant.

I would say that access with the same `rcl_node_t` doesn't need to be thread safe (that should be guaranted by rclcpp/rclpy/rcl<another_l>), but the function should be re-entrant for different nodes (i.e. if the function is making access to state somehow shared between the nodes, the implementation must make sure that access is safe).

Does that make sense?

hidmic:
CI up to `test_rmw_implementation` and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12448/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7415/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10158/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12363/)

hidmic:
Alright, going in !

PR #286, opened 2020-10-21
Add EventsExecutor

irobot-ros:
This PR introduces the changes required to implement the `EventsExecutor` design in rmw.
See [design](https://github.com/ros2/design/pull/305) and [Discourse post](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863).

The new executor uses an events queue and a timers manager as opposed to waitsets, to efficiently execute entities with work to do.

This new executor greatly reduces CPU usage of a ROS 2 application.
See the blog post for more details on the tests that we run.

The bulk of the changes for this implementation are in the rclcpp layer, with some minor changes in other repositories (rcl, rmw, rmw_implementation) for forwarding entities, the declaration of some data types in rcutils, and finally some additional changes in the vendor specific rmw implementations..
We currently implemented this only on top of the default ROS middleware fastrtps, while we provided stubs for other middlewares.

See the main PR to rclcpp https://github.com/ros2/rclcpp/pull/1416.

The current implementation does not support ROS 2 actions, which will be added in a follow up PR.

Developed by iRobot
Mauro Passerino
Lenny Story
Alberto Soragna

----

Connects to:

- design:
  - https://github.com/ros2/design/pull/305


ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/20

ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/29

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,58 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_set_log_severity(rmw_log_severity_t severity);
 
+/// Add documentation
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_subscription_set_events_executor_callback(
```
ivanpauno, line 2795: nitpick: can we avoid using "executor" in the name?

I know that the main goal of this work is to implement an executor avoiding the use of waitsets, but this feature itself could be called "callback" or "listener".

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,58 @@ RMW_WARN_UNUSED
...
+  const void * executor_context,
+  EventsExecutorCallback executor_callback,
+  const void * subscription_handle,
+  void * rmw_subscription);
```
ivanpauno, line 2799: do we need both?
can `rmw_subscription` be a `rmw_subscription_t *`?

I would rather use a different signature for each callback, appropriate to the entity where it's registered.

e.g.:

```c
  typedef void (*rmw_subscription_listener_callback_t) (const rmw_subscription_t * sub, void * context);
```

Things like `ExecutorEvent` can be part of `context`.

nitpick: use `snake_case_t` for types.

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,58 @@ RMW_WARN_UNUSED
...
+/// Add documentation
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_event_set_events_executor_callback(
+  const void * executor_context,
+  EventsExecutorCallback executor_callback,
+  const void * event_handle,
+  void * rmw_event,
+  bool use_previous_events);
```
ivanpauno, line 2841: what's the intention of this boolean?

ivanpauno (commented):
Thanks for working on this!

I have left some minor comments about the API, but I think it doesn't make a big difference.

rmw/include/rmw/rmw.h
ivanpauno: We have overloaded the world `context` quite a lot already, maybe just `argument` or similar is a good idea.

rmw/include/rmw/rmw.h
mauropasse: The idea was to differenciate this callback (belonging to the events executor) from the usual subscription callback (which process the message). That's the reason why the name is so explicit, but as this callback in the future could not only belong to the events executor, it could be renamed as `rmw_subscription_set_listener_callback`

rmw/include/rmw/rmw.h
mauropasse: This boolean was created originally for guard conditions, due they are used for different purposes. For example:
* Intra-process subscriptions are implemented as waitables, which are notified of a new msg by the triggering of their guard condition.
* The entities collector is also a waitable, when its guard condition is triggered it has to update everything.

When a IPC subscription (a waitable) is created, it can start receiving msgs even if is still not registered in the executor. By the time we set the `events_executor_callback`, we might have already some events for it. So if the flag `use_previous_events` is true, we push to the queue as many events as messages received while wasn't initialized, so we don't miss any message.
This is not desired for the entities collector waitable, because a node guard condition is triggered many times before the node is added to the executor, we usually have many events by the time we set the `events_executor_callback` and we don't care about previous events, as the action is always the same: reset the system. So the `use_previous_events` is false for the entities collector waitable.

In the case of events, we might want (or not) to take care of events happened before setting the `events_executor_callback`, so that's the reason of the boolean `use_previous_events`.

rmw/include/rmw/rmw.h
mauropasse: Yes, `rmw_subscription` can be a `rmw_subscription_t *`. I'll update the PR

Codematic71: If we used a custom signature for each callback it would require a custom callback function for each new event to match on the executor side. I think it reduces the utility of the service, by making it more difficult to expand.

Context is the top level entity data that mimics the behavior of a "this" pointer for the top level executor or structure that your connecting to.  It is part of the event propagation mechanism, rather than the data it carries.

rmw/include/rmw/rmw.h
ivanpauno: `rmw_subscription_set_listener_callback` is perfect, thanks for the change!

rmw/include/rmw/rmw.h
ivanpauno: > If we used a custom signature for each callback it would require a custom callback function for each new event to match on the executor side.

IIUC using a typed signature here doesn't make things more difficult, as you can do the following:

```c
typedef void (*rmw_subscription_listener_cb_t)(rmw_subscription_t * rmw_subscription, const void * listener_callback_context);

void function_taking_callback(rmw_subscription_listener_cb_t * subscription_cb);

void generic_callback(void * rmw_handle, const void * listener_callback_context) {
  // callback context has type information about what `rmw_handle` is.
  // ..  IMPLEMENTATION HERE ..
}

function_taking_callback(generic_callback); // this is fine, IIRC it's promoted without an explicit cast.
```

The only important thing is that all callbacks need to have the same number of arguments (and ordered equally).

---

On the other hand, the current signature allows getting directly the "rclcpp" or "whatever client library" handle directly instead of the "rmw handle" (e.g. see [here](https://github.com/ros2/rmw_fastrtps/blob/fbc75da5243ae087e1011f49f40c82ba104e2f4d/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp#L196)).

So maybe the current approach is better (and in order to get the "rclcpp" handle directly, type erasure is required).

@wjwwood any opinion?

rmw/include/rmw/rmw.h
ivanpauno: mmm, IMO the problem is that "guard conditions" should not trigger a callback.

A guard condition is only supposed to be used to awake a waitset, if you trigger it twice in a row and the waitset isn't awaken in the middle it will only take effect once.

We were using guard conditions for custom events in the waitset based executor, but I'm not sure if that fits perfectly here ... I have to put a bit of more thought into the issue

---

The idea of "counting" how many events happened before attaching the entity to an executor and then triggering the same amount of events also doesn't seem ideal for subscriptions, etc.
e.g.: if the history depth is 1.

But maybe it's an ok solution.

rmw/include/rmw/rmw.h
wjwwood: I think it's ok to have a type erased pointer that the user can provide, something like the C equivalent of `bind`, but I think we shouldn't call it context since that term is overloaded already in ROS 2. Maybe `user_payload`, because we don't know it's "context", it could be context or it could be nullptr or anything else...

We have to be careful about ownership in this case, since we're storing the "context" and assuming it stays in scope as long as we need it.

mauropasse: For clarifying. Currently, for example, we have:
```cpp
rmw_subscription_set_listener_callback(
  const void * callback_context,
  rmw_listener_cb_t listener_callback,
  const void * subscription_handle,
  rmw_subscription_t * rmw_subscription);
```
The listener_callback `rmw_listener_cb_t` (what will be called from the listeners, and is an executor function) can't have in its signature types like `rmw_subscription_t` since it is out of scope in the executor whom the function belongs (besides, it is not needed). Also, being a generic "listener callback", no particular entity types should be used in the signature, currently:

```cpp
typedef void (*rmw_listener_cb_t)(const void * callback_context, rmw_listener_event_t event);
```
The event `rmw_listener_event_t` has all the info needed by the executor to process the event:

```cpp
struct rmw_listener_event_t {
  const void * entity;
  rmw_listener_event_type_t type;
};
```

Regarding the name `callback_context` I think it makes sense since it's the context of the callback, and the owner of it is who must know how to cast this void * to whatever context the callback needs (in this case, to an executor pointer). But if you think it is confusing because the word `context` is already too much overloaded we can rename it as `user_payload` or `user_data`, but I'm not sure if this contributes to the clarity on the purpose of the word.

rmw/include/rmw/rmw.h
mauropasse: We agree that the purpose of guard conditions are waking up a waitset. But as intra-process subscriptions are implemented as waitables using guard conditions, we need to generate an event when they are triggered for an events based executor to handle it. This is only possible calling the callback from the guard condition.
Maybe attaching a "guard condition" to a waitable is what is not correct on first instance, it could be instead an "rclcpp event" attached to it (which will be the one calling the callback).

-----
Regarding the countings of events happened before setting the callback to a listener with a history depth = 1, currently is not the problem: the executor will have "no ops" subscription events. Notice that this same issue can happen also after setting the callback to the subscription listener, if it receives more events than what the executor can process. In this case we also end up with more events in the executor queue than messages in the DDS subscription queue, which again, results on "no ops" events. The discussion should be how to handle this empty events, if the current way creates a big decrease on performance.

ivanpauno: > Maybe attaching a "guard condition" to a waitable is what is not correct on first instance

The current implementation doesn't seem to be using a guard condition correctly:

- Each time a new intraprocess message is ready in the buffer, the guard condition is triggered.
- Executing the intraprocess subscription only delivers one message.

There's a possible issue in that approach:

- Two messages are ready, guard condition triggered twice.
- One message is delivered, guard condition is now `false` and the other message is still waiting ...

(I guess this wasn't noticed, as in general you have "something else" awaking the executor)

> it could be instead an "rclcpp event" attached to it (which will be the one calling the callback).

I think having some sort of custom event in `rclcpp` would be more appropriate, but I'm failing to see how that will still work in the wait-set based executor.

---

> Regarding the countings of events happened before setting the callback to a listener with a history depth = 1, currently is not the problem: the executor will have "no ops" subscription events. Notice that this same issue can happen also after setting the callback to the subscription listener, if it receives more events than what the executor can process

Yes, that's true.

rmw/include/rmw/rmw.h
ivanpauno: > But if you think it is confusing because the word context is already too much overloaded we can rename it as user_payload or user_data,

:+1:

rmw/include/rmw/rmw.h
mauropasse: > There's a possible issue in that approach:
   Two messages are ready, guard condition triggered twice.
    One message is delivered, guard condition is now false and the other message is still waiting ...

Yes, you're right. But this issue affects the wait set based executors (which BTW I haven't noticed it before, as you say in general there's "something else" awaking the executor), but it doesn't affect the events executor. So I'm wondering if the issue is outside the scope of this PR, so a new issue should be created and fixed in a different PR?

alsora: The issue that you mention with Intra-process communication is not present in the standard executor (or at least it was not present when I implemented the IPC stuff).

Once the executor is awaken, it checks all the entities in the waitset for work to do.
Guard conditions are mapped to rclcpp::Waitable objects and they implement a custom `is_ready()` function.
It does not matter how many times the guard condition has been triggered and/or how many times the executor woke up, since as long as the `is_ready` function returns true, it will keep executing callbacks.

In the case of IPC waitable the is_ready function checks if the IPC buffer is empty or not.

alsora: Actually I think that the sentence `Guard conditions are mapped to rclcpp::Waitable objects` is not correct.
The truth is that IPC subscriptions are mapped to waitables and the waitable is made of a guard condition and a buffer.
They are both needed in order to make the stuff work.

The important concept is that `rclcpp::Waitable` are an abstraction that can map to any RMW entity and or to other generic "things that you can wait on".
This requirement is respected by our events executor design and implementation, as every class that derives from waitable needs to define a custom function for registering the executor listeners.

rmw/include/rmw/rmw.h
mauropasse: Renamed `callback_context`  to `user_data` in https://github.com/ros2/rmw/pull/286/commits/4599137920bd4480a26f5f6d155bd4fbc648e672

rmw/include/rmw/rmw.h
ivanpauno: > The issue that you mention with Intra-process communication is not present in the standard executor (or at least it was not present when I implemented the IPC stuff).

I will have to double check, but AFAIR the executor is executing entities that are ready only once (i.e. taking one IPC message), and then going to sleep again.
The first time the `wait set` is awaken that "consumes" the guard condition, and if only one message is taken maybe you go to sleep again while having still a message in the IPC buffer and never get a new notification.

But I haven't experienced the issue, so I would have to try to reproduce it.

wjwwood: I think the way to fix this is to check all the intra process waitables that are ready and triggering their GC before waiting each time.

Basically we're needing to emulate the "Read Condition" from dds, which isn't just an event, it stays true (and therefore wakes the wait set) so long as data is available.

I think it's still important to wait (even if we know it will wake immediately) because it will also fill the wait set with other things that are now ready, whereas if you short circuited to the ready waitables, you might miss those things.

ivanpauno: Besides I think we will need to fix in rclcpp how the intraprocess subscription works, that's a bit unrelated to this PR.
Always triggering a guard condition when an intraprocess message is available might cause spurious awakes, but that's also something that can happen in other cases (as mentioned above).

So overall, I think that having a listener for the guard condition is ok.

rmw/include/rmw/types.h
```diff
@@ -207,6 +207,35 @@ typedef struct RMW_PUBLIC_TYPE rmw_guard_condition_t
...
+/// Encapsulate the RMW event implementation, data, and type.
+typedef struct RMW_PUBLIC_TYPE rmw_event_t
+{
+  /// Implementation identifier, used to ensure two different implementations are not being mixed.
+  const char * implementation_identifier;
+  /// Data specific to this event type from either the publisher or subscriber.
+  void * data;
+  /// The event type that occurred.
+  rmw_event_type_t event_type;
+} rmw_event_t;
```
ivanpauno, line 237: https://github.com/ros2/rmw/blob/e02978028f1756cc4c057d06f462ceac8acafcab/rmw/include/rmw/event.h#L33-L48

we will need some renaming here

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,117 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_set_log_severity(rmw_log_severity_t severity);
 
+/// Set callback function of the rmw subscription listener.
+/**
+ * This function sets the callback function which will be called whenever the
+ * subscription listener is notified about a new message for the subscription.
```
ivanpauno, line 2794: It would be nice to document that the callback will be called from background thread(s) that the rmw implementation uses.

ivanpauno (approved):
I think this looks good.

I still have a minimal preference for not having an `rmw_listener_event_t` argument in the listener callbacks and making that part of `user_data` (which will require registering not exactly the same function in all listeners of the event executor, but it can be an small wrapper of a common function).
Without the `rmw_listener_event_t` argument, the extra `void * <type>_handle` is also not required in the functions that register a listener.

Anyways, I think that adding this makes sense.
I would recommend ignoring my suggestions until @wjwwood also reviews (to avoid unnecessary iterations).

rmw/include/rmw/types.h
```diff
@@ -207,6 +207,35 @@ typedef struct RMW_PUBLIC_TYPE rmw_guard_condition_t
   rmw_context_t * context;
 } rmw_guard_condition_t;
 
+/// Define publisher/subscription events
+typedef enum rmw_event_type_t
```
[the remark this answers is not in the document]
ivanpauno: > Now this new location makes more sense also, it matches the types definition location for the rest of the entities (subscriptions, services, clients, etc). They are all in the same place now, i.e. here.

Ok. I think we actually wanted to do the opposite, split the giant `types.h` header in many headers (and only keep the `types.h` one for compatibility).

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,117 @@ RMW_WARN_UNUSED
...
+ * \return `RMW_RET_UNSUPPORTED` if the API is not implemented in the dds implementation
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_subscription_set_listener_callback(
+  const void * user_data,
+  rmw_listener_cb_t listener_callback,
+  const void * subscription_handle,
+  rmw_subscription_t * rmw_subscription);
```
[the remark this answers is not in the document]
mauropasse: Done in a0e92588055b1d7b070c999ec56de049c24708c6 and applied same approach on other layers for consistency

rmw/include/rmw/rmw.h
mauropasse: Done in b056b6d14c39ad22f81001f30234e6ba6a32f4ce

rmw/include/rmw/types.h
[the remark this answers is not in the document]
mauropasse: Currently if I leave the event type definitions on `rmw/event.h` I need to `#include "rmw/event.h"` on `rmw.h`. But as `event.h` already has `#include "rmw/types.h"` there's a conflict of unknown type names. I think is a good idea to split the `types.h` in many headers, but maybe is a little bit outside the scope of this PR.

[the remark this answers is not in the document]
wjwwood: I agree with @ivanpauno, maybe that's something I can do.

rmw/include/rmw/listener_event_types.h
```diff
@@ -0,0 +1,46 @@
...
+typedef enum rmw_listener_event_type_t rmw_listener_event_type_t;
+
+struct rmw_listener_event_t {
+  const void * entity;
+  rmw_listener_event_type_t type;
+};
+
+typedef struct rmw_listener_event_t rmw_listener_event_t;
+
+typedef void (*rmw_listener_cb_t)(const void * user_data, rmw_listener_event_t event);
```
wjwwood, line 39: nitpick: I know the names get long, but let's avoid abbreviations (beyond rmw) in the type names, so:

```suggestion
typedef void (*rmw_listener_callback_t)(const void * user_data, rmw_listener_event_t event);
```

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,127 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_set_log_severity(rmw_log_severity_t severity);
 
+/// Set callback function of the rmw subscription listener.
+/**
+ * This API sets the callback function which will be called whenever the
+ * subscription listener is notified about a new message for the subscription.
+ * The callback will be called from background thread(s) that the rmw
+ * implementation uses to wait (listen) for events to happen.
```
wjwwood, line 2796: Technically we cannot enforce that, nor does it make sense to require it. I'd say instead:

```suggestion
 * The callback may be called from a thread that the rmw implementation
 * created, rather than a thread owned by the user, i.e. some thread other
 * than user owned threads calling rmw functions such as rmw_wait() or
 * rmw_publish().
```

Basically, we should let the user know that what thread this comes from is not known, and in this way we do allow implementations to do what ever they need in order to implement this. For example, they may choose to call directly from within the `rmw_wait()` function in some cases or perhaps call subscription callbacks directly from `rmw_publish()`. I don't know if this advisable, but rather I would not want to box in what the rmw can do in the documentation of this function.

This is not mutually exclusive (in my opinion) to documenting that these callbacks must be serviced even if `rmw_wait()` (or any other rmw function) is never called.

What do you guys think? @ivanpauno

rmw/include/rmw/rmw.h
```diff
@@ -2787,6 +2788,127 @@ RMW_WARN_UNUSED
...
+ * \return `RMW_RET_UNSUPPORTED` if the API is not implemented in the dds implementation
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_subscription_set_listener_callback(
+  rmw_subscription_t * rmw_subscription,
+  rmw_listener_cb_t listener_callback,
+  const void * user_data,
+  const void * subscription_handle);
```
wjwwood, line 2812: Why is this needed? I see that it is put in as part of the event in the `entity` field, but why is that needed? I'm just curious because this information could be in the user data if needed right?

alsora: Yes, in theory it may be possible to use a single field, however I still see some value in having two separate ones.
The entity handle is a field that was meant to be required to be set by all users of this API, while on the other hand `user_data` is meant to contain any additional data that may or may not be needed depending on the listener that you are setting.

Without using an entity handle, I think that it becomes more tricky to implement a useful listener function (especially since this has to be a C-style function pointer).

IIRC this was the only reason.

rmw/include/rmw/rmw.h
ivanpauno: That sounds good to me

rpaaron:
I see there is some discussion over on https://github.com/ros2/design/pull/305 about executors being supported out of tree.  Will this PR get merged even if the other is still being discussed?

rmw/include/rmw/types.h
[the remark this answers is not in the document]
wjwwood: So with this patch (and rebasing) it works for me:

```diff
diff --git a/rmw/include/rmw/event.h b/rmw/include/rmw/event.h
index b051dde..8857fb6 100644
--- a/rmw/include/rmw/event.h
+++ b/rmw/include/rmw/event.h
@@ -29,6 +29,43 @@ extern "C"
 #include "rmw/ret_types.h"
 #include "rmw/visibility_control.h"
 
+/// Define publisher/subscription events
+typedef enum rmw_event_type_t
+{
+  // subscription events
+  RMW_EVENT_LIVELINESS_CHANGED,
+  RMW_EVENT_REQUESTED_DEADLINE_MISSED,
+  RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE,
+  RMW_EVENT_MESSAGE_LOST,
+
+  // publisher events
+  RMW_EVENT_LIVELINESS_LOST,
+  RMW_EVENT_OFFERED_DEADLINE_MISSED,
+  RMW_EVENT_OFFERED_QOS_INCOMPATIBLE,
+
+  // sentinel value
+  RMW_EVENT_INVALID
+} rmw_event_type_t;
+
+/// Define publisher/subscription event types
+typedef enum rmw_event_data_type_t
+{
+  RMW_SUBSCRIBER_EVENT,
+  RMW_PUBLISHER_EVENT
+} rmw_event_data_type_t;
+
+/// Encapsulate the RMW event implementation, data, and type.
+typedef struct RMW_PUBLIC_TYPE rmw_event_t
+{
+  /// Implementation identifier, used to ensure two different implementations are not being mixed.
+  const char * implementation_identifier;
+  /// Data specific to this event type from either the publisher or subscriber.
+  void * data;
+  /// The event type that occurred.
+  rmw_event_type_t event_type;
+  rmw_event_data_type_t event_data_type;
+} rmw_event_t;
+
 /// Return a zero initialized event structure.
 RMW_PUBLIC
 RMW_WARN_UNUSED
diff --git a/rmw/include/rmw/rmw.h b/rmw/include/rmw/rmw.h
index acccd7d..09038bd 100644
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -98,6 +98,7 @@ extern "C"
 #include "rosidl_runtime_c/service_type_support_struct.h"
 #include "rosidl_runtime_c/sequence_bound.h"
 
+#include "rmw/event.h"
 #include "rmw/init.h"
 #include "rmw/listener_callback_type.h"
 #include "rmw/macros.h"
diff --git a/rmw/include/rmw/types.h b/rmw/include/rmw/types.h
index 1ea7865..fa14c66 100644
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -207,43 +207,6 @@ typedef struct RMW_PUBLIC_TYPE rmw_guard_condition_t
   rmw_context_t * context;
 } rmw_guard_condition_t;
 
-/// Define publisher/subscription events
-typedef enum rmw_event_type_t
-{
-  // subscription events
-  RMW_EVENT_LIVELINESS_CHANGED,
-  RMW_EVENT_REQUESTED_DEADLINE_MISSED,
-  RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE,
-  RMW_EVENT_MESSAGE_LOST,
-
-  // publisher events
-  RMW_EVENT_LIVELINESS_LOST,
-  RMW_EVENT_OFFERED_DEADLINE_MISSED,
-  RMW_EVENT_OFFERED_QOS_INCOMPATIBLE,
-
-  // sentinel value
-  RMW_EVENT_INVALID
-} rmw_event_type_t;
-
-/// Define publisher/subscription event types
-typedef enum rmw_event_data_type_t
-{
-  RMW_SUBSCRIBER_EVENT,
-  RMW_PUBLISHER_EVENT
-} rmw_event_data_type_t;
-
-/// Encapsulate the RMW event implementation, data, and type.
-typedef struct RMW_PUBLIC_TYPE rmw_event_t
-{
-  /// Implementation identifier, used to ensure two different implementations are not being mixed.
-  const char * implementation_identifier;
-  /// Data specific to this event type from either the publisher or subscriber.
-  void * data;
-  /// The event type that occurred.
-  rmw_event_type_t event_type;
-  rmw_event_data_type_t event_data_type;
-} rmw_event_t;
-
 /// Allocation of memory for an rmw publisher
 typedef struct RMW_PUBLIC_TYPE rmw_publisher_allocation_t
 {

```

So, I don't see why including `event.h` is an issue in `rmw.h`, but if it is, then I'd say, put the event types in their own header rather than in `types.h`.

wjwwood:
I've been testing these changes with some rebasing. I think things are looking good. I still have one item to follow up on design wise, but in the mean time we need to clean up the commit history. Ideally, we could organize the commits (maybe squash them into one or at least fewer commits), but at the very least we need to have all the commits passing the DCO checker. This goes for the other pull requests too.

I would offer to do this for you, but I really need you guys to do the DCO signing.

I cannot test it until master is merged in or rebased against master (my preference).

wjwwood:
> I see there is some discussion over on ros2/design#305 about executors being supported out of tree. Will this PR get merged even if the other is still being discussed?

That's the plan. Merge the rmw/rcl changes related to "listener" style API, and then iterate on the c++ executor designs asynchronously.

rmw/include/rmw/rmw.h
wjwwood: > So overall, I think that having a listener for the guard condition is ok.

I can go along with this, but I'm a bit concerned we're abusing this when we could essentially just directly add an event to the "event executor" from the intra process subscription waitable (rather than trigger a guard condition to do that).

However, I'm still not convinced by the `use_previous_events` boolean.

> When a IPC subscription (a waitable) is created, it can start receiving msgs even if is still not registered in the executor. By the time we set the events_executor_callback, we might have already some events for it. So if the flag use_previous_events is true, we push to the queue as many events as messages received while wasn't initialized, so we don't miss any message.

Is this the only reason to have it? In my opinion, this isn't the way to solve this. Instead we should optionally include the listeners with the creation of the entity if we want to avoid this kind of race. I actually don't see the issue with not solving the race though. It is totally reasonable to me to have documentation that says "you will not receive callbacks for events that happened before the listener callback was specified".

Am I misunderstanding something here?

If not, I'd say we should not have this flag.

rmw/include/rmw/event.h
```diff
@@ -56,6 +63,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_event_t
   void * data;
   /// The event type that occurred.
   rmw_event_type_t event_type;
+  rmw_event_data_type_t event_data_type;
```
wjwwood, line 66: Briefly, what is the purpose of this field? I actually don't see it used anywhere (in https://github.com/ros2/rclcpp/pull/1416 or https://github.com/ros2/rclcpp/pull/1579 or https://github.com/ros2/rcl/pull/839)

rmw/include/rmw/rmw.h
wjwwood: @mauropasse can you elaborate on the things you were running into that this option resolves? I'm trying to tie together the discussion https://github.com/ros2/rmw_cyclonedds/pull/256#issuecomment-797599455, are they related?

On the one hand you have things like the QoS mismatch that happen at creation, but on the other hand that doesn't apply to guard conditions (events and guard conditions are the two things that have this option).

rmw/include/rmw/event.h
wjwwood: I see it used in https://github.com/ros2/rmw_cyclonedds/pull/256, but not in https://github.com/ros2/rmw_fastrtps/pull/468. Is it just to support Cyclone's implementation?

mauropasse: We had to use it because of the `rmw_event_t` type erased data field:
```
typedef struct RMW_PUBLIC_TYPE rmw_event_t
{
  const char * implementation_identifier;
  void * data; // What does data represent? For CycloneDDS, a publisher or a subscriber
  rmw_event_type_t event_type;
  rmw_event_data_type_t event_data_type;
} rmw_event_t;
```
There was no need for this field until we had to implement listeners callbacks for CycloneDDS.

For FastDDS case there's no issue, since when setting the listener callbacks for events, the `void * data` field [is casted to the event type](https://github.com/ros2/rmw_fastrtps/pull/468/files#diff-d87580e9fb67fa5de35008b4fa2d054a34cc456ed0782aae31cbe5106df817f3R78):
`auto fastrtps_event = static_cast<CustomEventInfo *>(rmw_event->data);`

But in the case of `rmw_cyclonedds_cpp` this type represents either a [publisher](https://github.com/ros2/rmw_cyclonedds/pull/256/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R698) or a [subscriber](https://github.com/ros2/rmw_cyclonedds/pull/256/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R690), and AFAIK there's no way to identify this data type from an `rmw_event_t`, when calling `rmw_event_set_listener_callback(rmw_event_t * rmw_event,)`

```
// rmw_event->data can be a CddsPublisher* or CddsSubscription*
auto pub_event = static_cast<CddsPublisher *>(rmw_event->data);
auto sub_event = static_cast<CddsSubscription *>(rmw_event->data);
```
Do you see any other alternative for this?

wjwwood: Seems like Cyclone should have their `rmw_event->data` point to a structure that contains this flag and then either the publisher or subscription? Unless this setting is useful above the rmw layer. Adding fields to rmw structures for rmw implementations is odd. I can imagine that this field is never used beyond cyclone and that would be undesirable I think.

wjwwood: I looks like the `CustomEventInfo` structure in rmw_fastrtps_cpp uses inheritance and a virtual function, so the equivalent would be what I described, something like this:

```c
enum cyclonedds_custom_event_info_type
{
  PUBLISHER_TYPE,
  SUBSCRIPTION_TYPE
}

struct cyclonedds_custom_event_info
{
  cyclonedds_custom_event_info_type type;
  CddsPublisher * publisher;
  CddsSubscription * subscription;
}
```

rmw/include/rmw/rmw.h
mauropasse: > I'm a bit concerned we're abusing this when we could essentially just directly add an event to the "event executor" from the intra process subscription waitable (rather than trigger a guard condition to do that).

Waitables are implemented differently depending on the waitable. Some of them wait on guard conditions to be triggered to do something. Others are waiting on events.
For the waitables that are waiting on guard conditions to be triggered to perform an action, the only place for pushing a `WAITABLE_EVENT` into the queue is where the guard condition is triggered. The idea is that when a guard condition is triggered, if it has assigned the listener callback it doesn't trigger the DDS guard condition, but calls the callback (pushes an event into the queue). See [here](https://github.com/ros2/rmw_fastrtps/blob/f9b62aef2366bb8a29a5716118fc6dea2749c8c8/rmw_fastrtps_shared_cpp/src/types/guard_condition.hpp#L41-L60).

We have currently many waitables:

- Entities collector: When a new entity is added to a node, the node's guard condition is triggered (`WAITABLE_EVENT`: The collector sets listener callback to the new entity).
- Intra-process subscription: When new message, the subscription's guard condition is triggered (`WAITABLE_EVENT`: The message is then taken).
- Events: When an event happens, like `on_requested_deadline_missed` the callback is called so there's a `WAITABLE_EVENT` in the queue for that specific event Waitable.
- EventsExecutorNotifyWaitable: When Ctrl+c or executor guard condition is triggerd, a `WAITABLE_EVENT` is pushed so spin can exit.

> However, I'm still not convinced by the `use_previous_events` boolean. Is this the only reason to have it?

There's another important reason. For example the Intra-process subscriptions Waitable.
Suppose 3 messages `A/B/C` arrive before setting the listener callback. Then the listener callback is assigned but we don't use the previous events: `use_previous_events=false`.
Then a new message arrives `D`, so a single event is pushed into the queue. A message is then taken but is not `D`, is `A` (the oldest). So we'll be always behind for 3 messages taking older ones, which won't happen if previous events were pushed into the queue.
On the other hand for the case of the Entities Collector Waitable, at start up a bunch of nodes+entities are created. The node's guard condition is triggered many times. By the time we assign the listener callback to each node's guard condition, if `use_previous_events=true` a lot of events would be pushed into the queue, which refreshes all the entities registered in the executor, multiple times. This takes a good amount of CPU, blocking the system for a while, and has bad consequences for a resource-constrained platform as the ones we use at iRobot. So in this case, we just care about new events.
This is why we need the `use_previous_events` boolean. Does this make sense?

wjwwood: Yes, I understand better now, thanks for the thorough explanation.

----

I am concerned about how we're using guard condition "listener" callbacks. Specifically:

> The idea is that when a guard condition is triggered, if it has assigned the listener callback it doesn't trigger the DDS guard condition, but calls the callback (pushes an event into the queue).

I don't really like that, because I feel like we don't need the guard condition at all then. We've introduced this alternative behavior for the guard condition that could be achieved (I think) in rclcpp (or higher) by overriding what happens when a Waitable should become ready, e.g. in cases like the intra process subscription where it triggers a guard condition atm. We could abstract that so that it triggers a guard condition for the current executor but pushes an event onto the event queue for the events executor design, never touching rcl/rmw/etc... in the latter case.

So I'm wondering if we should remove the guard condition callbacks. Or if we keep them, maybe they should be done in addition to triggering them, though I'm not sure about that either. I feel like we should stick more strictly to "guard conditions can asynchronously interrupt (wake) a waiting wait set", and that's it. If we need another concept we can create that, but I don't care much for expanding the scope of the guard condition as-is.

On the other hand, I want to think about whether or not this is needed if we want to create the "callback driven, rclcpp wait set" that we discussed in the past, where we maintain a wait set in rclcpp, updating it using the callbacks for entities rather than constructing rmw wait set objects and waiting on them. I _think_, like the event executor design, we could just update the rclcpp wait set directly rather than "trigger" a guard condition, which in turn calls a callback, which then updates the wait set.

I see the guard conditions different because they are typically triggered from the above the rmw layer, whereas the other things with callbacks are triggered from within the rmw layer (driven by external/asynchronous activity). The exception to that case is the graph guard conditions, which are created and triggered by the middleware, and waited on by the client library. I don't think this is a problem though, because we can continue to use those guard conditions as-is. They don't _have_ to be driven by the callbacks, like subscriptions, events, etc... would be.

----

Also, I was thinking about `use_previous_events` and how that works. I agree it may be needed for events, but I was wondering if might be better that instead of calling the callback once for every "stored" event, we instead have an argument to the callback signature that tells you how many have happened since the callback was called. This way you can do batch processing if you want. So code like this:

https://github.com/ros2/rmw_fastrtps/blob/f9b62aef2366bb8a29a5716118fc6dea2749c8c8/rmw_fastrtps_shared_cpp/src/custom_subscriber_info.cpp#L102-L115
```c++
  if (callback) {
    if (use_previous_events) {
      // Push events arrived before setting the executor's callback
      for(uint64_t i = 0; i < unread_events_count_; i++) {
        callback(user_data);
      }
    }
    user_data_ = user_data;
    listener_callback_ = callback;
  } else {
    user_data_ = nullptr;
    listener_callback_ = nullptr;
    return;
  }
```

Could instead be:

```c++
  if (callback) {
    if (use_previous_events) {
      // Push events arrived before setting the executor's callback
      callback(user_data, unread_events_count_);
    }
    user_data_ = user_data;
    listener_callback_ = callback;
  } else {
    user_data_ = nullptr;
    listener_callback_ = nullptr;
    return;
  }
```

And all other times you could just pass one (based on the assumption that the callback is called once for each "event").

In the case you described:

> On the other hand for the case of the Entities Collector Waitable, at start up a bunch of nodes+entities are created. The node's guard condition is triggered many times. By the time we assign the listener callback to each node's guard condition, if use_previous_events=true a lot of events would be pushed into the queue, which refreshes all the entities registered in the executor, multiple times. This takes a good amount of CPU, blocking the system for a while, and has bad consequences for a resource-constrained platform as the ones we use at iRobot. So in this case, we just care about new events.

You would get just one callback with some number as the second argument which you could use, ignore, log, or whatever.

----

Sorry for these late thoughts, it was only going through and thinking about documentation to add that I thought about these things. I'll bring them up at the middleware working group meeting tomorrow.

wjwwood: Summary from the middleware working group meeting:

- Seems like dropping callbacks for guard conditions is 👍 (please comment here if you still disagree)
- Seems like exchanging `use_previous_events` for having an `unread_count` as part of the callback is 👍

wjwwood:
@irobot-ros or @mauropasse, is it ok if I push directly (maybe force push to rebase) to this pull request branch? Or should I make my own?

irobot-ros:
@wjwwood yes sure, go ahead and push to this branch if you prefer.

FYI we just pushed a commit (here and in other PRs) that removes guard condition listeners as discussed in last Middleware WG.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14198/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8996/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11887/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14301/)

wjwwood:
@irobot-ros @mauropasse there are some test failures (I can fix the uncrustify one for `rmw_fastrtps`), but the others need some investigation, so if you have any time that would be great:

https://ci.ros2.org/job/ci_linux/14198/testReport/

I know it must be getting close to you guy's weekend, so if not I'll try to investigate tomorrow some if I have time.

irobot-ros:
I fixed the fastrtps uncrustify errors.
I had a quick look at the other failures, but I don't really know the internals of how QoS events work.
@mauropasse should be able to address them, but he will be back on Tuesday.

rmw/include/rmw/rmw.h
```diff
@@ -98,11 +98,13 @@ extern "C"
 #include "rosidl_runtime_c/service_type_support_struct.h"
 #include "rosidl_runtime_c/sequence_bound.h"
 
+#include "rmw/event.h"
 #include "rmw/init.h"
+#include "rmw/event_callback_type.h"
```
fujitatomoya, line 103:
```suggestion
#include "rmw/event_callback_type.h"
#include "rmw/init.h"
```

nitpick, header files should be alphabetical order?

mauropasse:
> @irobot-ros @mauropasse there are some test failures (I can fix the uncrustify one for rmw_fastrtps), but the others need some investigation, so if you have any time that would be great: https://ci.ros2.org/job/ci_linux/14198/testReport/

`projectroot.test.rclcpp.test_qos_event` fails because assigning listener callbacks in `CycloneDDS` breaks the wait_set based executors notify mechanisms: see https://github.com/ros2/rmw_cyclonedds/pull/256#issuecomment-791478380 (**Issue 1**). 
A temporary workaround is done in this PR: https://github.com/eclipse-cyclonedds/cyclonedds/pull/699
The `test_qos_event` is based on the `SingleThreadedExecutor` so is never notified about a happened event. 

I'll take a look to the other failing tests.

wjwwood:
I'm rerunning CI for this without the rclcpp pull request (https://github.com/ros2/rclcpp/pull/1579):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9053/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11945/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14361/)

If it looks good, then we can at least take those API's in, and maybe we can do ABI stable additions to `rclcpp` after the release. I'm going to propose we extend our pimpl pattern usage before the release to make that possible.

mauropasse:
I went through the failing tests 
```
projectroot.test.test_graph__rmw_cyclonedds_cpp
projectroot.test.test_events__rmw_cyclonedds_cpp
projectroot.uncrustify
rcl.NodeGraphMultiNodeFixture.test_node_info_subscriptions
rcl.NodeGraphMultiNodeFixture.test_node_info_publishers
rcl.NodeGraphMultiNodeFixture.test_node_info_services
rcl.NodeGraphMultiNodeFixture.test_node_info_clients
rcl.test_events__rmw_cyclonedds_cpp.gtest.missing_result
```
They are all fixed with my temporary workaround on CycloneDDS eclipse-cyclonedds/cyclonedds#699 , the issue is what I explained on https://github.com/ros2/rmw/pull/286#issuecomment-813443853

PR #288, opened 2020-10-28
[Backport Foxy] Lost subscription events

ahcorde:


ivanpauno:
Do we need the feature backported because some special reason?
I would rather not backport a feature.

ahcorde:
I was backporting some tests. But there is no specific reason to add a new feature. Closing this then.

PR #298, opened 2021-02-11
Deprecate rmw_time_t

### LINKED ISSUE #215: rmw_time_t is

emersonknapp:
Resolves #215

Removes `rmw_time_t` from the RMW API in favor of an int-based nanosecond type.
This simplifies most time-based code in the core, since rcl, rclcpp, and rclpy (and cyclonedds) all use an int64-based nanosecond timestamp already - it's a net-negative code change.

This should make #255 a lot easier to define - no special allowances will be needed around conversions to and from duration representations, nor can there be non-normalized durations (nanoseconds > 1billion)

Linked PRs:
* [ ] https://github.com/ros2/rmw_implementation/pull/178
* [ ] https://github.com/ros2/rmw_dds_common/pull/43
* [ ] https://github.com/ros2/rmw_cyclonedds/pull/283
* [ ] https://github.com/ros2/rmw_fastrtps/pull/508
* [ ] https://github.com/ros2/rmw_connext/pull/485
* [ ] https://github.com/ros2/rcl/pull/889
* [ ] https://github.com/ros2/rclcpp/pull/1549
* [ ] https://github.com/ros2/rclpy/pull/679
* [ ] https://github.com/ros2/rosbag2/pull/646
* [ ] https://github.com/ros2/demos/pull/484

Gist: https://gist.githubusercontent.com/emersonknapp/32eb5733086c7f6876e4bf9ece5d5fef/raw/caa6d31a0a84866cbc4e51afb60ad8b551a05acd/ros2.repos
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13700/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8577/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11416/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/13783/)

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -2414,7 +2414,7 @@ rmw_destroy_wait_set(rmw_wait_set_t * wait_set);
  * \param[inout] events Array of events to wait on.
  *   Can be `NULL` if there are no events to wait on.
  * \param[in] wait_set Wait set to use for waiting.
- * \param[in] wait_timeout If `NULL`, block indefinitely until an entity becomes ready.
+ * \param[in] wait_timeout If negative, block indefinitely until an entity becomes ready.
  *   If zero, do not block -- check only for immediately available entities.
  *   Else, this represents the maximum amount of time to wait for an entity to become ready.
  * \return `RMW_RET_OK` if successful, or
@@ -2435,7 +2435,7 @@ rmw_wait(
   rmw_clients_t * clients,
   rmw_events_t * events,
   rmw_wait_set_t * wait_set,
-  const rmw_time_t * wait_timeout);
+  rmw_duration_t wait_timeout);
 
 /// Return the name and namespace of all nodes in the ROS graph.
 /**
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -333,7 +333,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
   /// Nanoseconds component of this time point
   uint64_t nsec;
 } RMW_DECLARE_DEPRECATED (rmw_time_t,
-  "rmw_time_t has been deprecated in G-Turtle in favor of rmw_duration_t. "
+  "rmw_time_t has been deprecated in Galactic in favor of rmw_duration_t. "
   "It will be removed in H-Turtle");
 
 /// A duration of time, measured in nanoseconds.
--- a/rmw/include/rmw/types.h
+++ b/rmw/include/rmw/types.h
@@ -39,6 +39,12 @@ extern "C"
 // implementation. It may need to be increased in the future.
 #define RMW_GID_STORAGE_SIZE 24u
 
+#ifndef _WIN32
+# define RMW_DECLARE_DEPRECATED(name, msg) name __attribute__((deprecated(msg)))
+#else
+# define RMW_DECLARE_DEPRECATED(name, msg) name __pragma(deprecated(name))
+#endif
+
 /// Structure which encapsulates an rmw node
 typedef struct RMW_PUBLIC_TYPE rmw_node_t
 {
@@ -326,7 +332,12 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
 
   /// Nanoseconds component of this time point
   uint64_t nsec;
-} rmw_time_t;
+} RMW_DECLARE_DEPRECATED (rmw_time_t,
+  "rmw_time_t has been deprecated in G-Turtle in favor of rmw_duration_t. "
+  "It will be removed in H-Turtle");
+
+/// A duration of time, measured in nanoseconds.
+typedef rcutils_duration_value_t rmw_duration_t;
 
 typedef rcutils_time_point_value_t rmw_time_point_value_t;
 
@@ -389,11 +400,6 @@ enum RMW_PUBLIC_TYPE rmw_qos_durability_policy_t
   "RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE is deprecated. " \
   "Use RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC if manually asserted liveliness is needed."
 
-#ifndef _WIN32
-# define RMW_DECLARE_DEPRECATED(name, msg) name __attribute__((deprecated(msg)))
-#else
-# define RMW_DECLARE_DEPRECATED(name, msg) name __pragma(deprecated(name))
-#endif
 
 /// QoS liveliness enumerations that describe a publisher's reporting policy for its alive status.
 /// For a subscriber, these are its requirements for its topic's publishers.
@@ -425,13 +431,13 @@ enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_t
 };
 
 /// QoS Deadline default, 0s indicates deadline policies are not tracked or enforced
-#define RMW_QOS_DEADLINE_DEFAULT {0, 0}
+#define RMW_QOS_DEADLINE_DEFAULT 0
 
 /// QoS Lifespan default, 0s indicate lifespan policies are not tracked or enforced
-#define RMW_QOS_LIFESPAN_DEFAULT {0, 0}
+#define RMW_QOS_LIFESPAN_DEFAULT 0
 
 /// QoS Liveliness lease duration default, 0s indicate lease durations are not tracked or enforced
-#define RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT {0, 0}
+#define RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT 0
 
 /// ROS MiddleWare quality of service profile.
 typedef struct RMW_PUBLIC_TYPE rmw_qos_profile_t
@@ -444,13 +450,13 @@ typedef struct RMW_PUBLIC_TYPE rmw_qos_profile_t
   /// Durability QoS policy setting
   enum rmw_qos_durability_policy_t durability;
   /// The period at which messages are expected to be sent/received
-  struct rmw_time_t deadline;
+  rmw_duration_t deadline;
   /// The age at which messages are considered expired and no longer valid
-  struct rmw_time_t lifespan;
+  rmw_duration_t lifespan;
   /// Liveliness QoS policy setting
   enum rmw_qos_liveliness_policy_t liveliness;
   /// The time within which the RMW node or publisher must show that it is alive
-  struct rmw_time_t liveliness_lease_duration;
+  rmw_duration_t liveliness_lease_duration;
 
   /// If true, any ROS specific namespacing conventions will be circumvented.
   /**
--- a/rmw/test/test_topic_endpoint_info.cpp
+++ b/rmw/test/test_topic_endpoint_info.cpp
@@ -15,6 +15,7 @@
 #include "gmock/gmock.h"
 #include "osrf_testing_tools_cpp/scope_exit.hpp"
 #include "rcutils/allocator.h"
+#include "rcutils/time.h"
 
 #include "rmw/error_handling.h"
 #include "rmw/topic_endpoint_info.h"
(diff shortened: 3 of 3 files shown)


emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/32eb5733086c7f6876e4bf9ece5d5fef/raw/759013304e76ee83c9da9803fe102e1f889b8486/ros2.repos
BUILD args: 
TEST args: 
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7630
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13647/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8527/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11362/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13711/)

rmw/include/rmw/types.h
```diff
@@ -326,7 +332,12 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
 
   /// Nanoseconds component of this time point
   uint64_t nsec;
-} rmw_time_t;
+} RMW_DECLARE_DEPRECATED (rmw_time_t,
+  "rmw_time_t has been deprecated in G-Turtle in favor of rmw_duration_t. "
```
jacobperron, line 336: minor suggestion; might be a bit more clear
```suggestion
  "rmw_time_t has been deprecated in Galactic in favor of rmw_duration_t. "
```

rmw/test/test_topic_endpoint_info.cpp
```diff
@@ -146,10 +146,10 @@ TEST(test_topic_endpoint_info, set_qos_profile) {
   qos_profile.depth = 0;
   qos_profile.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
   qos_profile.durability = RMW_QOS_POLICY_DURABILITY_VOLATILE;
-  qos_profile.deadline = {1, 0};
-  qos_profile.lifespan = {2, 0};
+  qos_profile.deadline = RCUTILS_S_TO_NS(1);
```
jacobperron, line 149: The macro definition is being transitively included. It would be better to include `rcutils/time.h` explicitly.

rmw/test/test_topic_endpoint_info.cpp
```diff
@@ -173,19 +173,14 @@ TEST(test_topic_endpoint_info, set_qos_profile) {
   EXPECT_EQ(
     topic_endpoint_info.qos_profile.durability,
     RMW_QOS_POLICY_DURABILITY_VOLATILE) << "Unequal durability";
-  EXPECT_EQ(topic_endpoint_info.qos_profile.deadline.sec, 1u) << "Unequal deadline sec";
-  EXPECT_EQ(topic_endpoint_info.qos_profile.deadline.nsec, 0u) << "Unequal deadline nsec";
-  EXPECT_EQ(topic_endpoint_info.qos_profile.lifespan.sec, 2u) << "Unequal lifespan sec";
-  EXPECT_EQ(topic_endpoint_info.qos_profile.lifespan.nsec, 0u) << "Unequal lifespan nsec";
+  EXPECT_EQ(topic_endpoint_info.qos_profile.deadline, RCUTILS_S_TO_NS(1)) << "Unequal deadline";
+  EXPECT_EQ(topic_endpoint_info.qos_profile.lifespan, RCUTILS_S_TO_NS(2u)) << "Unequal lifespan";
```
jacobperron, line 177: nit: it looks odd to use an unsigned literal here and not elsewhere. Ultimately, [it is being cast to `long long`](https://github.com/ros2/rcutils/blob/d413e135d9d2a520f84b70ce3b0fd9625592ec50/include/rcutils/time.h#L30), so it would be better to use the literal `LL`, though it probably doesn't matter and we can drop it altogether.

rmw/test/test_topic_endpoint_info.cpp
```diff
@@ -173,19 +173,14 @@ TEST(test_topic_endpoint_info, set_qos_profile) {
...
   EXPECT_EQ(
     topic_endpoint_info.qos_profile.liveliness,
     RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC) << "Unequal liveliness";
   EXPECT_EQ(
-    topic_endpoint_info.qos_profile.liveliness_lease_duration.sec,
-    3u) << "Unequal liveliness lease duration sec";
-  EXPECT_EQ(
-    topic_endpoint_info.qos_profile.liveliness_lease_duration.nsec,
-    0u) << "Unequal liveliness lease duration nsec";
+    topic_endpoint_info.qos_profile.liveliness_lease_duration, RCUTILS_S_TO_NS(3u)
```
jacobperron, line 182: Same nit here.

jacobperron (commented):
A couple minor comments, otherwise LGTM.

Would be good to get a second approval before moving forward with the deprecation.

emersonknapp:
> A couple minor comments, otherwise LGTM.
> 
> Would be good to get a second approval before moving forward with the deprecation.

Thanks for the comments - I'll take care of them.

Do you think we should deprecated the existing `rmw_wait` signature (to be removed in H-turtle) and add this new signature as an override, or is replacing the existing function OK? I'm not actually sure if RMW implementations are allowed to not implement RMW APIs. If it's an error to not implement the new signature, then I wouldn't see a reason to deprecate+overload, otherwise it could make sense.

jacobperron:
> Do you think we should deprecate the existing `rmw_wait` signature (to be removed in H-turtle) and add this new signature as an override? Or, is replacing the existing function OK?

How do you propose to deprecate the signature? C doesn't support overloading, so I'm not sure how best to deprecate the existing signature short of introducing a new symbol (e.g. `rmw_wait_v2`).

emersonknapp:
> C doesn't support overloading,

Of course, I was still thinking in C++. I don't have an answer then - introducing a function with a different name is not a great alternative.

ivanpauno:
> Of course, I was still thinking in C++. I don't have an answer then - introducing a function with a different name is not a great alternative.

You can achieve "overloading" in C with the `_Generic` macro (part of the C11 standard).
See for example https://stackoverflow.com/a/25026358 or https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/.

I'm not sure if we want to use that thingy or not (I'm pretty sure MSVC/gcc/clang support that, but probably not other compilers).

If we want to move forward with this change, I also don't mind about having a `rcl_wait2` or `rmw_wait2` (or if considered ok, breaking API).

emersonknapp:
> Would be good to get a second approval before moving forward with the deprecation.

@jacobperron who would be best to weigh in with a second opinion? I have the followup of https://github.com/ros2/rmw/pull/255, which I'd like to get these both done in time for Galactic API freeze - just want to make sure we have a good buffer, I am not a fan of the scramble at distro freeze time.

rmw/include/rmw/rmw.h
```diff
@@ -2435,7 +2435,7 @@ rmw_wait(
   rmw_clients_t * clients,
   rmw_events_t * events,
   rmw_wait_set_t * wait_set,
-  const rmw_time_t * wait_timeout);
+  rmw_duration_t wait_timeout);
```
clalancette, line 2438: The thing I'm concerned about here is that this is an API change at the RMW layer.  While we can fix all of the in-tree rmws, this still is a hard break for [all](https://github.com/continental/rmw_ecal) [of](https://github.com/ros2/rmw_iceoryx) [the](https://github.com/ros2/rmw_dps) [out-of-tree](https://github.com/ros2/rmw_gurumdds) ones.  I'd be much happier with this change if we deprecated the old signature, added a new signature, and had a shim to convert between them.  Then we can remove the old signature for H-Turtle.  Thoughts?

emersonknapp: Is this what you're suggesting?

Galactic
```
rmw_wait(const rmw_time_t * wait_timeout) DEPRECATED { 
    rmw_wait2(time_to_duration(wait_timeout));
}

rmw_wait2(rmw_duration_t duration);
```

H-Turtle

```
rmw_wait(rmw_duration_t duration);

rmw_wait2(rmw_duration_t duration) DEPRECATED {
    rmw_wait(duration);
}
```

I-Turtle

```
rmw_wait(rmw_duration_t duration);
```

emersonknapp: Or more along the lines of just introducing a new name entirely (can't think what it should be called... `rmw_wait_on_set`?), then just deprecate `rmw_wait` forever in I-turtle.

Or, do you think this change is not warranted and that we should maintain the `rmw_time_t` type?

emersonknapp: Oh - another note: changing the QoS profiles to use rmw_duration_t instead of rmw_time_t is also is a breaking change for anything interacting with them directly - e.g. in https://github.com/ros2/rmw_cyclonedds/pull/283/files the `.sec` and `.nsec` usages have to be fixed. How would we handle this?

clalancette: > Or, do you think this change is not warranted and that we should maintain the `rmw_time_t` type?

That's hard for me to say.  If I take a look at how the `const rmw_time_t * wait_timeout` parameter to `rmw_wait` is used by the various RMWs, it looks like this:

* Fast-DDS - maps it to a [Duration_t](https://github.com/eProsima/Fast-DDS/blob/8ff9e0dd4531ef759897519e9c4a36087135cace/include/fastdds/rtps/common/Time_t.h#L78), which uses an int32_t/uint32_t to represent the same data.  There's clearly an impedance mis-match there (and the source of the `clamp_rmw_time_to_dds_time` method in rmw_dds_common).
* CycloneDDS - maps it to an internal [dds_time_t](https://github.com/eclipse-cyclonedds/cyclonedds/blob/b8c0657e63717aee1b02138e67a87b8ea691b053/src/ddsrt/include/dds/ddsrt/time.h#L45).  There is some possibility of impedance mismatch there (since in theory you could have very large numbers in both sec and nsec, and hence overflow the int64_t that dds_time_t is), but in real life that would be a pathological case.  (Incidentally, we should probably check for that pathological case before doing the conversion at https://github.com/ros2/rmw_cyclonedds/blob/d024823043504ea40af24bf22365a21cd203df55/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3384 , but that is orthogonal to this) (Also, it is kind of weird that rmw_cyclonedds maps this to a `dds_time_t` and not a `dds_duration_t`.  Practically it doesn't matter since they are both int64_t, but that seems more natural).
* Connext - maps it to an internal [DDS::Duration_t](https://github.com/ros2/rmw_connext/blob/fd3e6a71fee0f7eb26f84505d16169ce489d6235/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/wait.hpp#L324).  Internally this seems to be an int32/uint32 sec/nsec in Connext: https://community.rti.com/rti-doc/500/ndds.5.0.0/doc/html/api_cpp/structDDS__Duration__t.html
* (extra credit) eCal maps it to `std::duration` [here](https://github.com/continental/rmw_ecal/blob/abe07ae9c0839a91982a6bc91739a189dd3c6c92/rmw_ecal_dynamic_cpp/src/rmw.cpp#L836) .  Implicitly, the way that they are using it as waiting on a condition variable seems to be using it as an int64.
* (extra credit) iceoryx maps it to a `struct timespec` [here](https://github.com/ros2/rmw_iceoryx/blob/9ef51d3abbd85f17f9a523aa668663e9047b5ab9/rmw_iceoryx_cpp/src/rmw_wait.cpp#L93).
* (extra credit) dps maps it to a `std::duration` [here](https://github.com/ros2/rmw_dps/blob/71a72c384b2a1a64101169ec8b261a6b96bd13fd/rmw_dps_cpp/src/rmw_wait.cpp#L162)
* (extra credit) gurumdds maps it to an internal [dds_Duration_t](https://github.com/ros2/rmw_gurumdds/blob/c5488cb72db66a6e33f32a1f17cdf8cd51d1571e/rmw_gurumdds_shared_cpp/include/rmw_gurumdds_shared_cpp/rmw_wait.hpp#L334).  That seems to mimic the structure of `rmw_time_t` closely.

So, in short, all of the RMWs do it differently.  No matter what we do here, we're going to have some impedance mismatch with one of them.  From that perspective, `rmw_duration_t` doesn't seem substantially better than `rmw_time_t`, just different.

However, it may make sense to look at it from the point of view of the *users* of the RMW.  Do you think it makes more sense for a caller of `rmw_wait` to use an `rmw_duration_t` instead of an `rmw_time_t`?  If so, why?

> Or more along the lines of just introducing a new name entirely (can't think what it should be called... `rmw_wait_on_set`?), then just deprecate `rmw_wait` forever in I-turtle.

Even simpler:

For Galactic:
* add in `rmw_wait_on_set(rmw_duration_t duration);`
* mark `rmw_wait(rmw_time_t *)` deprecated
* Update all internal users to use `rmw_wait_on_set`

For H-Turtle:
* Remove `rmw_wait(rmw_time_t *)`

> Oh - another note: changing rmw_qos_profile_t to use rmw_duration_t instead of rmw_time_t is also is a breaking change for anything interacting with them directly - e.g. in https://github.com/ros2/rmw_cyclonedds/pull/283/files the .sec and .nsec usages have to be fixed. How would we handle this?

Ug.  In an ideal world, we would add another structure and do the deprecation dance there as well, but that is starting to get a bit ridiculous.  I'm curious to hear your thoughts above before we discuss the finer point here.

emersonknapp: Well, at the end of the day, my _real_  goal is to solve https://github.com/ros2/rmw/issues/210, which is an actual behavioral problem that is not handled in RMW. Based on this conversation, I'm inclined to lean back on an implementation that doesn't require changing `rmw_time_t` to achieve the goal.

clalancette: > Edit: my inclination especially is based on the fact that there's no across-the-board benefit for RMW implementations.
> The original reason to change this was to make time representation more uniform across the ROS 2 core, since rcl, rclcpp, rclpy all represent times/durations as a single int64 nanoseconds. Perhaps we can gain the necessary benefit with just a few convenience functions/macros for converting between

I see.  So that would argue for making the switch to int64.  Making that more consistent does seem worthwhile, but I also think that it shouldn't block your other work on #210.  So maybe the thing to do here is to put all of this on hold, concentrate on #210, and return to this after Galactic.

emersonknapp: https://github.com/ros2/rmw/pull/301

emersonknapp:
I am closing this for now as I have no specific plan to readdress it later. I'm fairly convinced that there is not much benefit, and the simple helper functions added in #301 meet current needs.

PR #299, opened 2021-02-11
Add declaration for function to check QoS profile compatibility

jacobperron:
Currently, users who are creating a publisher or subscription can receive 'QoS incompatibility'
events from the RMW if an incompatible endpoint is discovered. While this is useful, we
currently don't have a nice way for application to generally check if two QoS profiles are
compatible. For example, it would be nice if tooling could query the communication graph and
report any detected QoS incompatibilities.

In order to reduce code duplication, I think an API for checking QoS compatibilty should live
in a common place. I've opted for `rmw` (over a place like `rcl`) since it's possible QoS
compatiblity rules may vary per RMW vendor. Since rules for all DDS implementations should be
the same, we could put that common logic in `rmw_dds_common`.

---

The motivation for this API is to add features to tooling like [rqt_graph](https://github.com/ros-visualization/rqt_graph) and [ros2doctor](https://github.com/ros2/ros2cli/tree/f3a15e4b8f71336c03e56bbbf954ea87097459a0/ros2doctor) to help diagnose QoS compatibility issues. E.g. it would be useful to change how incompatible links in the ROS graph are rendered in `rqt_graph`.

---

Connected PRs:

- rmw_dds_common (implementation): https://github.com/ros2/rmw_dds_common/pull/45
- rmw_cyclonedds: https://github.com/ros2/rmw_cyclonedds/pull/286
- rmw_fastrtps: https://github.com/ros2/rmw_fastrtps/pull/511
- rmw_connext: https://github.com/ros2/rmw_connext/pull/488
- rmw_implementation: https://github.com/ros2/rmw_implementation/pull/180
- rmw_connextdds: TODO

CI for connected PRs listed above:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13736/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/8606/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11447/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13810/)

### DIFF
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -169,6 +169,7 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_compatibility_type_t
  * \return `RMW_RET_OK` if the check was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `compatiblity` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if any of the policies have value "unknown".
+ * \return `RMW_RET_ERROR` if there is an unexpected error.
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -134,10 +134,21 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_compatibility_type_t
  * If any of the profile policies has the value "system default", then it may not be
  * possible to determine the compatibilty.
  * In this case, the output parameter `compatibility` is set to `RMW_QOS_COMPATIBILITY_WARNING`
- * and `reason` is populated
+ * and `reason` is populated.
  *
  * Profile policies must not have the value "unknown". An "unknown" value is considered an error
  * and `RMW_RET_INVALID_ARGUMENT` is returned.
+ * `reason` will be set, identifying the offending policy.
+ *
+ * If there is a compatibility warning or error, and a buffer is provided for `reason`, then an
+ * explanation of all warnings and errors will be populated into the buffer, separated by
+ * semi-colons (`;`).
+ * Errors will appear before warnings in the string buffer.
+ * If the provided buffer is not large enough, this function will still write to the buffer, up to
+ * the `reason_size` number of characters.
+ * Therefore, it is possible that not errors and warnings are communicated if the buffer size limit
+ * is reached.
+ * A buffer size of 2048 should be more than enough to capture all possible errors and warnings.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -136,8 +136,8 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_compatibility_type_t
  * In this case, the output parameter `compatibility` is set to `RMW_QOS_COMPATIBILITY_WARNING`
  * and `reason` is populated
  *
- * Profile policies must not have the value "unknown". This is considered an error, and the
- * output parameter `compatiblity` is not set.
+ * Profile policies must not have the value "unknown". An "unknown" value is considered an error
+ * and `RMW_RET_INVALID_ARGUMENT` is returned.
  *
  * <hr>
  * Attribute          | Adherence
@@ -156,7 +156,8 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_compatibility_type_t
  *   Must be pre-allocated by the caller. This parameter is optional and may be set to `NULL`.
  * \param[in] reason_size: Size of the string buffer `reason`, if one is provided.
  * \return `RMW_RET_OK` if the check was successful, or
- * \return `RMW_RET_INVALID_ARGUMENT` if any of the policies have value "unknown"
+ * \return `RMW_RET_INVALID_ARGUMENT` if `compatiblity` is NULL, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any of the policies have value "unknown".
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -113,29 +113,60 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
   false
 };
 
+typedef enum RMW_PUBLIC_TYPE rmw_qos_compatibility_type_t
+{
+  /// QoS policies are compatible
+  RMW_QOS_COMPATIBILITY_OK = 0,
+
+  /// QoS policies may not be compatible
+  RMW_QOS_COMPATIBILITY_WARNING,
+
+  /// QoS policies are not compatible
+  RMW_QOS_COMPATIBILITY_ERROR
+} rmw_qos_compatibility_type_t;
+
+
 /// Check if two QoS profiles are compatible.
 /**
- * Two QoS profiles are compatible if the a publisher and subcription
- * using their QoS policies can communicate with each other.
+ * Two QoS profiles are compatible if a publisher and subcription
+ * using the QoS policies can communicate with each other.
+ *
+ * If any of the profile policies has the value "system default", then it may not be
+ * possible to determine the compatibilty.
+ * In this case, the output parameter `compatibility` is set to `RMW_QOS_COMPATIBILITY_WARNING`
+ * and `reason` is populated
+ *
+ * Profile policies must not have the value "unknown". This is considered an error, and the
+ * output parameter `compatiblity` is not set.
  *
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | No
- * Thread-Safe        | No
+ * Thread-Safe        | Yes
  * Uses Atomics       | No
- * Lock-Free          | No
+ * Lock-Free          | Yes
  *
  * \param[in] publisher_profile: The QoS profile used for a publisher.
  * \param[in] subscription_profile: The QoS profile used for a subscription.
- * \return `true` if the publisher and subscription profiles are compatible, `false` otherwise.
+ * \param[out] compatibility: `RMW_QOS_COMPATIBILITY_OK` if the QoS profiles are compatible, or
+ *   `RMW_QOS_COMPATIBILITY_WARNING` if the QoS profiles might be compatible, or
+ *   `RMW_QOS_COMPATIBILITY_ERROR` if the QoS profiles are not compatible.
+ * \param[out] reason: A detailed reason for a QoS incompatibility.
+ *   Must be pre-allocated by the caller. This parameter is optional and may be set to `NULL`.
+ * \param[in] reason_size: Size of the string buffer `reason`, if one is provided.
+ * \return `RMW_RET_OK` if the check was successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if any of the policies have value "unknown"
  */
 RMW_PUBLIC
 RMW_WARN_UNUSED
-bool
-rmw_qos_profile_are_compatible(
+rmw_ret_t
+rmw_qos_profile_check_compatible(
   const rmw_qos_profile_t publisher_profile,
-  const rmw_qos_profile_t subscription_profile);
+  const rmw_qos_profile_t subscription_profile,
+  rmw_qos_compatibility_type_t * compatibility,
+  char * reason,
+  size_t reason_size);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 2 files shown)


rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,30 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+/// Check if two QoS profiles are compatible.
+/**
+ * Two QoS profiles are compatible if the a publisher and subcription
+ * using their QoS policies can communicate with each other.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
```
mjeronimo, line 125: This function would definitely be handy. About the attribute table: why wouldn't this function be thread-safe and lock-free? Wouldn't it just access the two structures, evaluating for compatibility?

rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,30 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] publisher_profile: The QoS profile used for a publisher.
+ * \param[in] subscription_profile: The QoS profile used for a subscription.
+ * \return `true` if the publisher and subscription profiles are compatible, `false` otherwise.
```
gbiggs, line 131: Do you think it would be possible to have this function or another function give an explanation of where the incompatibility is?

rmw/include/rmw/qos_profiles.h
jacobperron: Yeah, I think you're right. The table can likely be updated to indicate it *is* thread-safe and lock-free.

rmw/include/rmw/qos_profiles.h
jacobperron: Sure, we could add a `char **` parameter for output that can be set in the case that profiles are incompatible. This will also require we pass in an allocator for the string. It might make using the function a little more complex to use, but probably worth it for the additional information.

jacobperron: After talking with @wjwwood, I like the idea of avoiding an allocator completely and making this a fixed size string (e.g. 1024 characters). It's more amenable for real-time systems and makes the implementation easier.

jacobperron:
I've refactored the proposed API. Two major changes:

1. Now instead of a `bool`, it outputs the compatibility result as an enum type parameter. The return code is reserved for other errors not related to compatibility.
The compatibility can have one of three states:
    - "OK": the profiles are compatible
    - "WARNING": because one or more policies was set to "system default", we don't actually know if the profiles are compatible.
    - "ERROR": the profiles are incompatible.

2. Added a string buffer (and size) parameter for setting the reason for an incompatibility or warning.
The user should provide a pre-allocated array so that memory management is completely in their hands.

@mjeronimo @gbiggs PTAL

jacobperron:
Oh, and I also renamed the function on a whim.

rmw/include/rmw/qos_profiles.h
jacobperron: Updated in 219217a

mjeronimo:
I agree with the name change given the change in the return type. 

Looks good to me, although I would've preferred a definitive yes/no answer, which would, of course, require a way to query the underlying implementation for its QoS system defaults.

rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,62 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+ * \return `RMW_RET_INVALID_ARGUMENT` if any of the policies have value "unknown".
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_qos_profile_check_compatible(
+  const rmw_qos_profile_t publisher_profile,
+  const rmw_qos_profile_t subscription_profile,
+  rmw_qos_compatibility_type_t * compatibility,
+  char * reason,
```
mjeronimo, line 169: I assume that the reason string would return info only about the first incompatibility encountered, correct? That is, the incompatibility check will stop upon the first incompatibility encountered and return info about that? What if there is a warning encountered first and then a compatibility error? Will the reason string provide information about the warning or the error? I'm assuming it's about the error.

gbiggs: The first error should be given, and then the first warning if no errors. This needs to be documented.

rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,62 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+
+/// Check if two QoS profiles are compatible.
+/**
+ * Two QoS profiles are compatible if a publisher and subcription
+ * using the QoS policies can communicate with each other.
+ *
+ * If any of the profile policies has the value "system default", then it may not be
+ * possible to determine the compatibilty.
+ * In this case, the output parameter `compatibility` is set to `RMW_QOS_COMPATIBILITY_WARNING`
+ * and `reason` is populated
```
gbiggs, line 137:
```suggestion
 * and `reason` is populated.
```

gbiggs, line 137: There needs to be documentation about whether `reason` is populated in case of `ERROR` as well.

rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,62 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+ *
+ * Profile policies must not have the value "unknown". An "unknown" value is considered an error
```
gbiggs, line 139: Should this also produce a value in `reason`?

rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,62 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] publisher_profile: The QoS profile used for a publisher.
+ * \param[in] subscription_profile: The QoS profile used for a subscription.
+ * \param[out] compatibility: `RMW_QOS_COMPATIBILITY_OK` if the QoS profiles are compatible, or
+ *   `RMW_QOS_COMPATIBILITY_WARNING` if the QoS profiles might be compatible, or
+ *   `RMW_QOS_COMPATIBILITY_ERROR` if the QoS profiles are not compatible.
+ * \param[out] reason: A detailed reason for a QoS incompatibility.
+ *   Must be pre-allocated by the caller. This parameter is optional and may be set to `NULL`.
```
gbiggs, line 156:
```suggestion
 *   Must be pre-allocated by the caller. This parameter is optional and may be set to `nullptr`.
```

gbiggs (changes requested):
Some documentation improvements are needed.

rmw/include/rmw/qos_profiles.h
jacobperron: Technically, it is possible to string together all warnings/errors. E.g. each warning/error separating by a semi-colon.

If you think it's better to return a the first error or warning, it's an easy change to make.

rmw/include/rmw/qos_profiles.h
jacobperron: This is a C API; `nullptr` is not defined as it is a C++11 thing.

wjwwood:
> Added a string buffer (and size) parameter for setting the reason for an incompatibility or warning.
The user should provide a pre-allocated array so that memory management is completely in their hands.

👍 like this even more than what we talked about.

rmw/include/rmw/qos_profiles.h
gbiggs: The problem with stringing them all together is that the user might pass in a buffer that's only big enough to hold one, which might happen to be a warning. I guess that could be documented, though.

rmw/include/rmw/qos_profiles.h
gbiggs: Oh, I thought `nullptr` made it into the latest C as well. What a shame.

rmw/include/rmw/qos_profiles.h
jacobperron: Yeah, it's a problem in general that `reason` may not be large enough even for a single error/warning.
For the DDS implementations, I can make sure that errors appear before warnings.

jacobperron:
I've opened a PR with the implementation for DDS RMWs here: https://github.com/ros2/rmw_dds_common/pull/45

rmw/include/rmw/qos_profiles.h
jacobperron: Yeah, I think that makes sense. Documented in 28a9f5f.

rmw/include/rmw/qos_profiles.h
jacobperron: See 28a9f5f

jacobperron: Fixed in 28a9f5f

rmw/include/rmw/qos_profiles.h
jacobperron: Documented in 28a9f5f

rmw/include/rmw/qos_profiles.h
jacobperron: Changed my mind after doing the implementation. If there is an error code result, then no output parameters will be set: 041f02c

jacobperron:
Thanks for the feedback! I've updated this PR and opened a bunch of connected PRs that implement this for our supported RMWs (see the updated description for a list). The common implementation lives in `rmw_dds_common`, so this one could use the most scrutiny: ros2/rmw_dds_common#45

rmw/include/rmw/qos_profiles.h
```diff
@@ -163,11 +163,15 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_compatibility_type_t
  * \param[out] compatibility: `RMW_QOS_COMPATIBILITY_OK` if the QoS profiles are compatible, or
  *   `RMW_QOS_COMPATIBILITY_WARNING` if the QoS profiles might be compatible, or
  *   `RMW_QOS_COMPATIBILITY_ERROR` if the QoS profiles are not compatible.
- * \param[out] reason: A detailed reason for a QoS incompatibility.
- *   Must be pre-allocated by the caller. This parameter is optional and may be set to `NULL`.
+ * \param[out] reason: A detailed reason for a QoS incompatibility or potential incompatibility.
+ *   Must be pre-allocated by the caller.
+ *   This parameter is optional and may be set to `NULL` if the reason information is not
```
mjeronimo, line 168: Use 'nullptr's here instead of NULLs?

ivanpauno: `nullptr` is cpp only AFAIK

ivanpauno (approved):

rmw/include/rmw/qos_profiles.h
```diff
@@ -113,6 +113,78 @@ static const rmw_qos_profile_t rmw_qos_profile_unknown =
...
+ * using the QoS policies can communicate with each other.
+ *
+ * If any of the profile policies has the value "system default", then it may not be
+ * possible to determine the compatibilty.
+ * In this case, the output parameter `compatibility` is set to `RMW_QOS_COMPATIBILITY_WARNING`
+ * and `reason` is populated.
+ *
+ * Profile policies must not have the value "unknown".
+ * An "unknown" value is considered an error and `RMW_RET_INVALID_ARGUMENT` is returned.
+ * `compatibility` and `reason` will not be set.
```
ivanpauno, line 141: Pending discussion in another PR https://github.com/ros2/rmw_dds_common/pull/45#discussion_r580215422

jacobperron: Updated 93e9aa2

Unknown values now result in a compatibility warning instead of an error.

ivanpauno (approved):

fujitatomoya (approved):
lgtm

jacobperron:
Thanks for all the feedback so far!

I'm just waiting for ci.ros2.org to come back online and then I'll trigger CI and get this merged/released.

asorbini:
I noticed this new API from a new test failure in `test_rmw_implementation`, and added support for it to `rmw_connextdds` in [91753dc](https://github.com/rticommunity/rmw_connextdds/commit/91753dcc9ad7dc7c21d77eda0684ff646ad5994b).

I also noticed that these changes have not yet made it to the Rolling Debian binaries, and I had to build the tree from source in order to validate them. What is the period with which those packages are usually regenerated?

ivanpauno:
I think that releasing `rmw_implementation` is pending, but IIUC @jacobperron already released up to `rmw`.

> I also noticed that these changes have not yet made it to the Rolling Debian binaries, and I had to build the tree from source in order to validate them. What is the period with which those packages are usually regenerated?

IIUC the packages used by `build.ros2.org` for testing will be updated almost right away (delay of some hours), but the public ones (the ones you download when `apt install ...`) will only get updated after a sync.
I guess there will be a sync in about May as ROS 2 Galactic will be released for that date, though IDK.

jacobperron:
@ivanpauno is correct. If you switch to the testing repo ([see instructions](https://docs.ros.org/en/foxy/Installation/Prerelease-Testing.html#debian-testing-repository)), then you should see these changes already released there. Syncs to main happen periodically (roughly every three weeks), though I don't think there has been one for a couple months now.

@nuclearsandwich what do you think about doing a Rolling sync in the near future?

PR #301, opened 2021-02-19
Introduce RMW_DURATION_INFINITE constant and API return value promise

### LINKED ISSUE #210: QoS duration APIs need INFINITE constant and an API contract to enforce proper usage by implementations

emersonknapp:
Resolves #210 
Alternative to #255

Introduce a new constant, `RMW_DURATION_INFINITE`. Specify in `get_topic_endpoint_info` that it is expected to be returned to represent infinite durations. Add some helper functions that enable better usage of this new value.

This change is _non-breaking_, all RMW implementations will continue to work correctly as-is. However, there are two expectations that they must eventually implement:
1. If a user inputs RMW_DURATION_INFINITE for Deadline, Lifespan, or Liveliness Lease Duration QoS policy values, this should be translated into the implementation-specific infinity value.
1. When returning values in `get_topic_endpoint_info`, RMW impementations must intercept their own internall-defined "infinity" definitions and replace those values with RMW_DURATION_INFINITE, so that they can be understood correctly.

Today, I will _not_ be changing the default qos_profile values to RMW_DURATION_INFINITY, as this would break existing implementations that do not know how to handle this value correctly. I would like to create a "value deprecation" so that this default value could be switched over in time for H-Turtle, but I'm not sure how this should be done.

Dependent (but not linked, can be merged after) PRs:
* [ ] https://github.com/ros2/rmw_fastrtps/pull/515
* [ ] https://github.com/ros2/rmw_cyclonedds/pull/288
* [ ] https://github.com/ros2/rmw_connext/pull/491
* [ ] rosbag2 playback
* [ ] ros2topic info should print "Infinite" instead of some large number for readability
* Out of tree DDS implementations are recommended to follow suit. This feature will affect their ability to have rosbags played back correctly, and for the CLI to print "infinite" in the right places.


emersonknapp:
@jacobperron @clalancette @mm318 @hidmic @fujitatomoya @Barry-Xu-2018

Hey all, as participants in the #210 discussion, I would like your feedback on this. See discussion on #298 for why I'm not addressing #215 here. 

I've only performed one RMW implementation update, to show the pattern, without doing too much extra work. After going too far on other iterations of this feature, I wanted to wait until we have some consensus on the API before opening more update PRs.

rmw/src/types.c
```diff
@@ -22,3 +24,39 @@ rmw_get_zero_initialized_message_info(void)
...
+{
+  return rmw_time_total_nsec(left) == rmw_time_total_nsec(right);
+}
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_duration_t
+rmw_time_total_nsec(const rmw_time_t time)
+{
+  return RCUTILS_S_TO_NS(time.sec) + time.nsec;
```
fujitatomoya, line 41: I think it should clip into `RMW_DURATION_INFINITE` when overflow happens? because rmw_time_t possibly has bigger values tha rmw_duration_t can handle. (it would be nice to have test case on this accordingly.)

rmw/src/types.c
```diff
@@ -22,3 +24,39 @@ rmw_get_zero_initialized_message_info(void)
...
+}
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_time_t
+rmw_time_from_nsec(const rmw_duration_t nanoseconds)
```
fujitatomoya, line 47: nanoseconds could be negative, in that case should we return `RMW_DURATION_INFINITE`? (it would be nice to have test case on this accordingly.)

fujitatomoya (changes requested):

rmw/include/rmw/types.h
```diff
@@ -329,6 +329,41 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
 } rmw_time_t;
 
 typedef rcutils_time_point_value_t rmw_time_point_value_t;
+typedef rcutils_duration_value_t rmw_duration_t;
+
+/// Constant representing an infinite duration. Use rmw_time_equal for comparisons.
+/**
+  * Different RMW implementations have different representations for infinite durations.
+  * This value is reported for QoS policy durations that are left unspecified.
+  * Do not directly comparee `sec == sec && nsec == nsec`, because we don't want to be sensitive
```
jacobperron, line 338:
```suggestion
  * Do not directly compare `sec == sec && nsec == nsec`, because we don't want to be sensitive
```

rmw/include/rmw/types.h
```diff
@@ -329,6 +329,41 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
...
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_time_t
+rmw_time_from_nsec(const rmw_duration_t nanoseconds);
+
+/// Ensure that an rmw_time_t does not have nanoseconds > 1 second.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_time_t
+rmw_time_normalize(const rmw_time_t time);
```
jacobperron, line 366: minor: It might be a good time to break out time related API to it's own header, `rmw/time.h`.

jacobperron (commented):
> If a user inputs RMW_DURATION_INFINITE for Deadline, Lifespan, or Liveliness Lease Duration QoS policy values, this should be interpreted in the same way that {0, 0} (RMW_QOS_DURATION_UNSPECIFIED) has historically been interpreted.

I think we can improve the documentation around this point. The current [documentation](https://github.com/ros2/rmw/blob/ae0553adb574c031343965195ec4a73039e62389/rmw/include/rmw/types.h#L427) implies that a zero value means the policies are ignored (ie. not enforced), but digging through the various implementations it looks like a zero value causes the RMW's default value to be used instead, which happens to be "infinity" in most cases. The current documentation seems misleading, since really the policies are being enforced (albeit with very large values). Also, it isn't clear to me if the RMW default value might be changed via a vendor-specific configuration (then I think all bets are off).

Having an explicit INFINITE value in rmw makes sense to me. But I'm not sure if the behavior should be the same as a "default" value, considering the "default" value might be configured with a vendor config.

This discussion relates to another one we're having about QoS compatibility, see https://github.com/ros2/rmw_dds_common/pull/45#discussion_r580144067

rmw/test/test_types.cpp
```diff
@@ -26,3 +26,82 @@ TEST(test_types, zero_initialized_message_info) {
 
   EXPECT_FALSE(info.from_intra_process);
 }
+
+TEST(test_types, time_equal) {
+  {
+    rmw_time_t a {100, 100};
```
hidmic, line 32: @emersonknapp nit: why the extra space i.e.:
```suggestion
    rmw_time_t a{100, 100};
```
?

rmw/include/rmw/types.h
```diff
@@ -329,6 +329,41 @@ typedef struct RMW_PUBLIC_TYPE rmw_time_t
...
+bool
+rmw_time_equal(const rmw_time_t left, const rmw_time_t right);
+
+/// Return the total nanosecond representation of a time.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_duration_t
+rmw_time_total_nsec(const rmw_time_t time);
+
+/// Construct rmw_time_t from a total nanoseconds representation.
```
hidmic, line 356: @emersonknapp may be worth specifying an origin (I presume Unix'es epoch).

rmw/test/test_types.cpp
emersonknapp: I find it more readable this way but I'm happy to remove the space - the linter doesn't seem to have an opinion either way

rmw/src/types.c
emersonknapp: Thanks for calling this out. Maybe we should actually change the signature to return success? Like so:

```
rmw_ret_t rmw_time_total_nsec(const rmw_time_t time, rmw_duration_t * result);
```

(maybe this makes sense from `rmw_time_from_nsec` as well, returning an error code for negative values

rmw/include/rmw/types.h
emersonknapp: rmw_time_t doesn't encode an origin. It is only used in practice to store durations, which don't have an origin.

emersonknapp:
@jacobperron @hidmic @fujitatomoya Thanks for the feedback - I've pushed a new version that I hope addresses all concerns:

* Move time declarations and utilities into `time.h`, `time.c`, `test_time.cpp`
* Explicitly state the differences between `RMW_DURATION_UNSPECIFIED` and `RMW_DURATION_INFINITE`
* Add bounds checking to time utilities (negative durations, overflows)

fujitatomoya (approved):
lgtm

jikawa-az (approved):

rmw/include/rmw/types.h
hidmic: I see. Naming is counter-intuitive then. I see you intend to deprecate it in https://github.com/ros2/rmw/pull/298 though, so I won't block.

hidmic (approved):
LGTM pending green CI

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13807/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8674/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11516/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13895/)

rmw/test/test_time.cpp
```diff
@@ -0,0 +1,121 @@
...
+    EXPECT_EQ(rmw_time_total_nsec(time), MAXTIME);
+  }
+  {
+    // Will overflow, expect clamp
+    rmw_time_t time{1, MAXTIME};
+    EXPECT_EQ(rmw_time_total_nsec(time), MAXTIME);
+  }
+  {
+    // Overflow on seconds alone, not at the type limit
+    //
```
jacobperron, line 77:
```suggestion
```

jacobperron (approved):
LGTM

emersonknapp:
@jacobperron looks all good except the `rcl_action` tests - it looks like those failed in last night's nightly https://ci.ros2.org/view/nightly/job/nightly_win_rel/1863/#showFailuresLink so I'm assuming it's a preexisting issue - should we block on merging?

jacobperron:
@emersonknapp Looks like this is the related issue: https://github.com/ros2/rmw_connext/issues/472

I think it's okay to not block on the `rcl_action` graph test failures.

emersonknapp:
@jacobperron @hidmic friendly ping: can you merge this or direct it to the person who can?

hidmic:
Thanks for the bump @emersonknapp. Merging.

PR #302, opened 2021-02-25
Add support for content filtered topics

iuhilnehc-ynos:
Related to https://github.com/ros2/design/pull/282 to add content-filtered-topic interfaces.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -1142,6 +1142,7 @@ rmw_subscription_get_actual_qos(
  * \return `RMW_RET_INVALID_ARGUMENT` if `expression_parameters` is NULL, or
  * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `node` implementation
  *   identifier does not match this implementation, or
+ * \return `RMW_RET_UNSUPPORTED` if the implementation does not support content filter topic, or
  * \return `RMW_RET_ERROR` if an unspecified error occurs.
  */
 RMW_PUBLIC
@@ -1178,6 +1179,7 @@ rmw_subscription_set_cft_expression_parameters(
  * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `node` implementation
  *   identifier does not match this implementation, or
  * \return `RMW_RET_BAD_ALLOC` if memory allocation fails, or
+ * \return `RMW_RET_UNSUPPORTED` if the implementation does not support content filter topic, or
  * \return `RMW_RET_ERROR` if an unspecified error occurs.
  */
 RMW_PUBLIC
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -1117,6 +1117,77 @@ rmw_subscription_get_actual_qos(
   const rmw_subscription_t * subscription,
   rmw_qos_profile_t * qos);
 
+/// Set the filter expression and expression parameters for the subscription.
+/**
+ * This function will set a filter expression and an array of expression parameters
+ * for the given subscription, but not to update the original rmw_subscription_options_t
+ * of subscription.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] implementation defined, check the implementation documentation</i>
+ *
+ * \param[in] subscription subscription the subscription object to inspect.
+ * \param[in] filter_expression An filter expression to set.
+ * \param[in] expression_parameters Array of expression parameters to set,
+ *   it can be NULL if there is no placeholder in filter_expression.
+ * \return `RMW_RET_OK` if the query was successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `subscription` is NULL, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `filter_expression` is NULL, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `expression_parameters` is NULL, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `node` implementation
+ *   identifier does not match this implementation, or
+ * \return `RMW_RET_ERROR` if an unspecified error occurs.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_subscription_set_cft_expression_parameters(
+  const rmw_subscription_t * subscription,
+  const char * filter_expression,
+  const rcutils_string_array_t * expression_parameters);
+
+/// Retrieve the filter expression of the subscription.
+/**
+ * This function will return an filter expression by the given subscription.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] implementation defined, check the implementation documentation</i>
+ *
+ * \param[in] subscription subscription the subscription object to inspect.
+ * \param[out] filter_expression an filter expression, populated on success.
+ *   It is up to the caller to deallocate the filter expression later on,
+ *   using rcutils_get_default_allocator().deallocate().
+ * \param[out] expression_parameters Array of expression parameters, populated on success.
+ *   It is up to the caller to finalize this array later on, using rcutils_string_array_fini().
+ * \return `RMW_RET_OK` if the query was successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `subscription` is NULL, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `filter_expression` is NULL or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `expression_parameters` is NULL, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `node` implementation
+ *   identifier does not match this implementation, or
+ * \return `RMW_RET_BAD_ALLOC` if memory allocation fails, or
+ * \return `RMW_RET_ERROR` if an unspecified error occurs.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_subscription_get_cft_expression_parameters(
+  const rmw_subscription_t * subscription,
+  char ** filter_expression,
+  rcutils_string_array_t * expression_parameters);
+
 /// Take an incoming ROS message.
 /**
  * Take a ROS message already received by the given subscription, removing it from internal queues.
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -1133,9 +1133,15 @@ rmw_subscription_get_actual_qos(
  * <i>[1] implementation defined, check the implementation documentation</i>
  *
  * \param[in] subscription the subscription object to inspect.
- * \param[in] filter_expression An filter expression to set.
- * \param[in] expression_parameters Array of expression parameters to set,
- *   it can be NULL if there is no placeholder in filter_expression.
+ * \param[in] filter_expression A filter_expression is a string that specifies the criteria
+ *   to select the data samples of interest. It is similar to the WHERE part of an SQL clause.
+ *   Using an empty("") string can reset/clean content filtered topic for the subscription.
+ * \param[in] expression_parameters An expression_parameters is an array of strings that
+ *   give values to the ‘parameters’ (i.e., "%n" tokens begin from 0) in the filter_expression.
+ *   The number of supplied parameters must fit with the requested values in the filter_expression.
+ *   It can be NULL if there is no "%n" tokens placeholder in filter_expression.
+ *   The maximun size allowance depends on concrete DDS vendor.
+ *   (i.e., it cannot be greater than 100 on RTI_Connext.)
  * \return `RMW_RET_OK` if the query was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `subscription` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `filter_expression` is NULL, or
@@ -1149,7 +1155,7 @@ RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_ret_t
 rmw_subscription_set_cft_expression_parameters(
-  const rmw_subscription_t * subscription,
+  rmw_subscription_t * subscription,
   const char * filter_expression,
   const rcutils_string_array_t * expression_parameters);
 
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -1132,7 +1132,7 @@ rmw_subscription_get_actual_qos(
  * Lock-Free          | Maybe [1]
  * <i>[1] implementation defined, check the implementation documentation</i>
  *
- * \param[in] subscription subscription the subscription object to inspect.
+ * \param[in] subscription the subscription object to inspect.
  * \param[in] filter_expression An filter expression to set.
  * \param[in] expression_parameters Array of expression parameters to set,
  *   it can be NULL if there is no placeholder in filter_expression.
@@ -1165,7 +1165,7 @@ rmw_subscription_set_cft_expression_parameters(
  * Lock-Free          | Maybe [1]
  * <i>[1] implementation defined, check the implementation documentation</i>
  *
- * \param[in] subscription subscription the subscription object to inspect.
+ * \param[in] subscription the subscription object to inspect.
  * \param[out] filter_expression an filter expression, populated on success.
  *   It is up to the caller to deallocate the filter expression later on,
  *   using rcutils_get_default_allocator().deallocate().
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -91,6 +91,7 @@ extern "C"
 #include <stddef.h>
 #include <stdint.h>
 
+#include "rcutils/allocator.h"
 #include "rcutils/macros.h"
 #include "rcutils/types.h"
 
@@ -1117,11 +1118,10 @@ rmw_subscription_get_actual_qos(
   const rmw_subscription_t * subscription,
   rmw_qos_profile_t * qos);
 
-/// Set the filter expression and expression parameters for the subscription.
+/// Set the content filtered topic options for the subscription.
 /**
  * This function will set a filter expression and an array of expression parameters
- * for the given subscription, but not to update the original rmw_subscription_options_t
- * of subscription.
+ * for the given subscription.
  *
  * <hr>
  * Attribute          | Adherence
@@ -1132,23 +1132,15 @@ rmw_subscription_get_actual_qos(
  * Lock-Free          | Maybe [1]
  * <i>[1] implementation defined, check the implementation documentation</i>
  *
- * \param[in] subscription the subscription object to inspect.
- * \param[in] filter_expression A filter_expression is a string that specifies the criteria
- *   to select the data samples of interest. It is similar to the WHERE part of an SQL clause.
- *   Using an empty("") string can reset/clean content filtered topic for the subscription.
- * \param[in] expression_parameters An expression_parameters is an array of strings that
- *   give values to the ‘parameters’ (i.e., "%n" tokens begin from 0) in the filter_expression.
- *   The number of supplied parameters must fit with the requested values in the filter_expression.
- *   It can be NULL if there is no "%n" tokens placeholder in filter_expression.
- *   The maximun size allowance depends on concrete DDS vendor.
- *   (i.e., it cannot be greater than 100 on RTI_Connext.)
- * \return `RMW_RET_OK` if the query was successful, or
- * \return `RMW_RET_INVALID_ARGUMENT` if `subscription` is NULL, or
- * \return `RMW_RET_INVALID_ARGUMENT` if `filter_expression` is NULL, or
- * \return `RMW_RET_INVALID_ARGUMENT` if `expression_parameters` is NULL, or
- * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `node` implementation
+ * \param[in] subscription The subscription to set content filtered topic options.
+ * \param[in] options The content filtered topic options.
+ *   Use `options.filter_expression` with an empty("") string to
+ *   reset/clean content filtered topic for the subscription.
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if an argument is null, or
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `subscription` implementation
  *   identifier does not match this implementation, or
- * \return `RMW_RET_UNSUPPORTED` if the implementation does not support content filter topic, or
+ * \return `RMW_RET_UNSUPPORTED` if the implementation does not support content filtered topic, or
  * \return `RMW_RET_ERROR` if an unspecified error occurs.
  */
 RMW_PUBLIC
@@ -1156,12 +1148,11 @@ RMW_WARN_UNUSED
 rmw_ret_t
 rmw_subscription_set_cft_expression_parameters(
   rmw_subscription_t * subscription,
-  const char * filter_expression,
-  const rcutils_string_array_t * expression_parameters);
+  const rmw_subscription_content_filtered_topic_options_t * options);
 
-/// Retrieve the filter expression of the subscription.
+/// Retrieve the content filtered topic options of the subscription.
 /**
- * This function will return an filter expression by the given subscription.
+ * This function will return a content filtered topic options by the given subscription.
  *
  * <hr>
  * Attribute          | Adherence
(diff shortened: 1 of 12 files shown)


fujitatomoya (approved):

iuhilnehc-ynos:
rebase and ready to implement CFT for rmw_fastrtps

fujitatomoya:
@ivanpauno @wjwwood @clalancette could you review this? 
CC: @MiguelCompany @asorbini

rmw/include/rmw/subscription_content_filter_options.h
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * It is similar to the WHERE part of an SQL clause.
+   */
+  char * filter_expression;
+  /**
+   * give values to the tokens placeholder ‘parameters’ (i.e., "%n" tokens begin from 0) in the
+   * filter_expression. The number of supplied parameters must fit with the requested values.
+   *
+   * It can be NULL if there is no "%n" tokens placeholder in filter_expression.
+   * The maximun size allowance depends on concrete DDS vendor.
```
ivanpauno, line 43: it would be nice to have a `rmw_get_max_content_filter_expression_parameters()`, it it's possible to implement it

ivanpauno (commented):
The rmw API looks good to me.

What is going to be the behavior for DDS vendors that don't support content filtering?

rmw/include/rmw/subscription_content_filter_options.h
fujitatomoya: `NOT SUPPORT` will be returned to application for now.

fujitatomoya:
@ivanpauno 

> What is going to be the behavior for DDS vendors that don't support content filtering?

`NOT SUPPORT` will be returned to application for now.

https://github.com/ros2/design/pull/282/files#diff-0cb9ecc495b4e57ddcb33bd070916b8cc1a2c5a812c4ec2020f93b5bb858d197R221-R225

we have been discussing internally, but not implementing the rcl filtering function yet. this is something we could do as follow-up, but we are not positive on this approach.

rmw/include/rmw/subscription_content_filter_options.h
iuhilnehc-ynos: @ivanpauno 

Thank you for starting to review it. I'd appreciate that.

iuhilnehc-ynos:
@ivanpauno @fujitatomoya 

> What is going to be the behavior for DDS vendors that don't support content filtering?

There are two cases(1. create a subscriber with the content filter at the beginning; 2. set content filter after creating a subscriber that can be non-content filter) to set content filtering. 

In the first case, it just returns the subscriber as normal with an `is_cft_enabled` flag that is false.
In the second, it returns `NOT SUPPORT` as @fujitatomoya mentioned.

ivanpauno:
> we have been discussing internally, but not implementing the rcl filtering function yet. this is something we could do as follow-up, but we are not positive on this approach.

I'm not sure if it's a good idea to add a feature only supported by one DDS vendor.
In that case, we're not going to be able to take much advantage of the feature.

@wjwwood any opinion?

MiguelCompany:
> I'm not sure if it's a good idea to add a feature only supported by one DDS vendor.

@ivanpauno We've been working hard to have this feature on Fast DDS. It has just been released with v2.5.1

We are changing the Fast DDS branch for rolling in ros2/ros2#1241

@iuhilnehc-ynos Has been working on the `rmw_fastrtps` integration but we didn't want ros2/ros2#1241 to block the review of the RMW interfaces, so that's why we decided to start with the empty stub, and the content filter implementation on `rmw_fastrtps` will be done on a follow-up PR.

ivanpauno:
> @iuhilnehc-ynos Has been working on the rmw_fastrtps integration but we didn't want https://github.com/ros2/ros2/pull/1241 to block the review of the RMW interfaces, so that's why we decided to start with the empty stub, and the content filter implementation on rmw_fastrtps will be done on a follow-up PR.

Well that sounds way better!
It's still a bit strange to me to add a major feature not supported by all rmw implementations.
Ideally, we should have a way to still use content filtering in the implementations that don't support the feature, even if in that case there's no performance benefit.

fujitatomoya:
@MiguelCompany thanks for the follow-up

@ivanpauno sorry i did not mention that rmw_fastrtps status. that will be supported as well. (So rmw_fastrtps and rmw_connextdds will be supported with this PR for Humble.)

> Ideally, we should have a way to still use content filtering in the implementations that don't support the feature, even if in that case there's no performance benefit.

okay we just wanted to discuss with community how we wanna go for sure. if that is the way to go, we are expecting that we could integrate this filtering in rcl. (expecting that no interfaces need to be changed.) that is something we can do, and we will do that after this.

fujitatomoya:
Or if rmw_cyclonedds has a plan to support content filtering, we are also willing to work on rmw_cyclonedds to enable that. in this case, we do not really need filtering in rcl?

ivanpauno:
> Or if rmw_cyclonedds has a plan to support content filtering, we are also willing to work on rmw_cyclonedds to enable that. in this case, we do not really need filtering in rcl?

Maybe not in that case, but it will make implementing a fully featured rmw implementation much harder.

ivanpauno:
That sounds good to me.
@clalancette @wjwwood opinions?

fujitatomoya:
@iuhilnehc-ynos could you also share your thought?

rmw/include/rmw/subscription_content_filter_options.h
iuhilnehc-ynos: > If this is the specification in ROS 2 RMW IF, it needs to be checked and return `RMW_RET_INVALID_ARGUMENT`?

make sense. (Currently, I just return `RMW_RET_ERROR` if failed to `set_filter_expression`, I'll update it.)
```
    ReturnCode_t ret =
      filtered_topic->set_filter_expression(options->filter_expression, expression_parameters);
    if (ret != ReturnCode_t::RETCODE_OK) {
      // TODO: to check `RETCODE_BAD_PARAMETER`, then return `RMW_RET_INVALID_ARGUMENT`
      RMW_SET_ERROR_MSG(
        "failed to set_filter_expression");
      return RMW_RET_ERROR;
    }
    return RMW_RET_OK;
```

> probably we do not limit the size, but make sure that we can return the error code from implementation?

agree

rmw/include/rmw/rmw.h
```diff
@@ -1115,6 +1116,70 @@ rmw_subscription_get_actual_qos(
...
+ * \return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` if the `subscription` implementation
+ *   identifier does not match this implementation, or
+ * \return `RMW_RET_UNSUPPORTED` if the implementation does not support content filtered topic, or
+ * \return `RMW_RET_ERROR` if an unspecified error occurs.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_subscription_set_content_filter(
+  rmw_subscription_t * subscription,
```
ivanpauno, line 1148: do all DDS vendors support to set a content filter expression to a preexisting subscription, or do some of them only allow to specify one when the subscription (data reader) is created?

The other option is to have a `rmw_create_cft_subscription(rmw_subscription_t * subscription, ...)` function where the filter expression is specified when the subscription is created (being able to change the filter expression later is a more niche feature).

ivanpauno (commented):
I have done a more detailed review now, I still have to take a look at the tests

rmw/include/rmw/rmw.h
fujitatomoya: Fast-DDS and RTI Connextdds both works this interface to preexisting subscription. I do not really think that this is well described in the DDS Spec. @asorbini @MiguelCompany any thoughts?

> being able to change the filter expression later is a more niche feature

This interface is required to use CFT for ROS 2 action client id and parameter events filtering. from where i can see this, probably not really niche but more flexible. (I would use this feature for preexisting subscription.)

rmw/include/rmw/subscription_content_filter_options.h
```diff
@@ -0,0 +1,130 @@
...
+ * \param[in] expression_parameter_argv The expression parameters argv.
+ * \param[in] allocator The allocator used when copying data to the content filter options.
+ * \param[out] options The content filter options to be set.
+ * \returns RMW_RET_INVALID_ARGUMENT, or
+ * \returns RMW_RET_BAD_ALLOC, or
+ * \returns RMW_RET_OK
+ */
+RMW_PUBLIC
+rmw_ret_t
+rmw_subscription_content_filter_options_set(
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I think that the functions of `init` and `fini` should be a pair. Actually, the `set` is like a helper function calling `fini` and `init`.

rmw/include/rmw/rmw.h
iuhilnehc-ynos: `rmw_subscription_set_content_filter` has the ability to create_cft_subscription inside if the content filter of current subscription is not enabled before.  This is implemented in rmw(rmw_connext, rmw_fastrtps) layer, not DDS.

MiguelCompany: @ivanpauno As @iuhilnehc-ynos said, this is implemented in the RMW.

On the DDS spec, ContentFilteredTopic has a method to change the expression parameters. Both RTI Connext and eProsima Fast DDS have an extension that allows changing both the expression and the parameters. For other vendors this could be indirectly supported by deleting the DataReader and the ContentFilteredTopic and creating them again.

rmw/include/rmw/subscription_content_filter_options.h
```diff
@@ -0,0 +1,130 @@
...
+   */
+  char * filter_expression;
+  /**
+   * give values to the tokens placeholder ‘parameters’ (i.e., "%n" tokens begin from 0) in the
+   * filter_expression. The number of supplied parameters must fit with the requested values.
+   *
+   * It can be NULL if there is no "%n" tokens placeholder in filter_expression.
+   * The maximum index number must be smaller than 100.
+   */
+  rcutils_string_array_t * expression_parameters;
```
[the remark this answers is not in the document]
ivanpauno: That adds an extra indirection and not much value.
IMO the first choice should be inline storage, and the heap should be used when needed.

rmw/include/rmw/rmw.h
ivanpauno: > For other vendors this could be indirectly supported by deleting the DataReader and the ContentFilteredTopic and creating them again.

I think that we should try to avoid that, as removing and creating an entity is not ideal.

> This interface is required to use CFT for ROS 2 action client id and parameter events filtering. from where i can see this, probably not really niche but more flexible. (I would use this feature for preexisting subscription.)

That's a good use case.
In that case, do you only need to modify the parameters and not the filter expression?

ivanpauno, line 1148: would it be a good idea to have a different type for cft subscriptions?
Though that might not be needed in the current implementations, it might make things easier for other future ones.

See also https://github.com/ros2/rmw_fastrtps/pull/513#discussion_r823789500.

iuhilnehc-ynos: > I think that we should try to avoid that, as removing and creating an entity is not ideal.

There is a case that users do not use the content filter for a subscription totally, so in the rmw_implementation, it only needs to create a normal topic to keep the behavior as before rather than create an extra content filtered topic with an empty filter. 
Please notice that if updating the content filter after the content filtered topic is created, there is no new entity to be removed or created.

> do you only need to modify the parameters and not the filter expression

To allow the [filter_expression](https://github.com/ros2/rmw/blob/a767a7863c652fe0645286ffdba156b5085d73e5/rmw/include/rmw/subscription_content_filter_options.h#L37) to be nullptr , and then call the set_expression_paraemter of DDS.

rmw/include/rmw/subscription_content_filter_options.h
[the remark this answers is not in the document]
iuhilnehc-ynos: What if some users only need to set the filter_expression like "data='test'" without expression_parameters?

rmw/include/rmw/rmw.h
ivanpauno: > cft_subscription should only exist in concrete rmw_implement and be transparent to users.

It can always be transparent to users at a higher level, e.g. `CFTSubscription` can inherit from `SubscriptionBase` in rclcpp.
The thing is that all subscriptions now need to be able to be upgraded to a cft subscription, which is possible in DDS but might make things harder to other implementations (not if they solve the filtering subscription side, in that case the upgrade is also easy).
i.e.: At a higher level a `CFTSubscription` could be upcasted to a `Subscription`, being that transparent to the user, but you need a `CFTSubscription` if you want to modify the filter expression or the filter expression parameters.

fujitatomoya: i see that what @ivanpauno suggesting. rmw should be agnostic from DDS APIs (which btw current interfaces are NOT bound to DDS APIs), and what's more important is general interfaces to support feature in implementation. user frontend APIs are different topic, we can make it transparent as we like in client language.

IMO, CFT can be categorized into an extension of subscription. that is the original idea and reason that we have current interfaces. (@iuhilnehc-ynos i think we had internal discussion almost year ago if we should create CFT specific interfaces or not? and then we got to this interface from that reason. correct me if i am mistaken.)

iuhilnehc-ynos: The concept of content filter belongs to the topic, not a subscription.

rmw/include/rmw/subscription_content_filter_options.h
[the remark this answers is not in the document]
ivanpauno: > What if some users only need to set the filter_expression like "data='test'" without expression_parameters?

Then the array is empty. (i.e. `expression_parameters.size == 0`).
e.g.: in cpp if you need an empty `std::vector` you just store one inline, you don't need to allocate a `std::vector *` in the heap and use `nullptr` when not needed.

rmw/include/rmw/rmw.h
ivanpauno: > The concept of content filter belongs to the topic, not a subscription

This is maybe how it's specified in DDS, but you could easily ignore that and see it as a content filtered subscription (for me, it's more natural to think a topic just as a name/type, and think of content filtering as an extra requirement of the subscription on the data it wants to get).

> IMO, CFT can be categorized into an extension of subscription. that is the original idea and reason that we have current interfaces

The problem with the current interfaces is that it's not an extension, now ALL subscriptions can be content filtered.
i.e. we have added a capability to all subscriptions, not an extension of the simple subscription type with more power.

If you want to make it really an extension in rmw as well, we could have a way to get `rmw_subscription_t` from `rmw_cft_subscription_t`. But the capabality of setting/modifying a content filter expression should be of `rmw_cft_subscription_t` IMO.

Maybe my concern is not a real problem for others.
@hidmic @wjwwood for a second opinion.

fujitatomoya: I see. if we really need to divide CFT subscription from normal subscription, that is something we can do. but i am not really inclined to do so. cz every time these kind of feature we develop, we have to manage different structure and interfaces. i would keep this as subscription option.
anyway we are open to opinion!!

rmw/include/rmw/subscription_content_filter_options.h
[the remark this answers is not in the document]
iuhilnehc-ynos: I means `rcutils_string_array_t * expression_parameters` can use less memory size than `rcutils_string_array_t  expression_parameters` in the runitime if user doesn't use the expression_parameters.  (I used the `std::vector` in the rclcpp.)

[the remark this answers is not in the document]
ivanpauno: > I means rcutils_string_array_t * expression_parameters can use less memory size than rcutils_string_array_t expression_parameters in the runitime if user doesn't use the expression_parameters

The memory size of the struct doesn't really matter, it's just some more bytes inline.

My comment is just about code style, you should prefer using stack/inline storage if it's available, and only use the heap if you have a reason to do so.

rmw/include/rmw/rmw.h
ivanpauno: See also https://github.com/ros2/rmw_fastrtps/pull/513#discussion_r826208750

rmw/include/rmw/subscription_content_filter_options.h
asorbini: AFAIK the only limit in Connext to the length of a filter expression and the total number/total length of the parameters is limited by [DomainParticipantQos::resource_limits::contentfilter_property_max_length](https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/api/connext_dds/api_c/structDDS__DomainParticipantResourceLimitsQosPolicy.html#aef1ff851202f5777d31528a52d528e4f).

This property is 256 by default. `rmw_connextdds` extends it to a minimum of 1024, but the behavior can be disabled (by using variable `RMW_CONNEXT_PARTICIPANT_QOS_OVERRIDE_POLICY`), and users can customize it via XML if needed.

The length of this buffer must be long enough to accommodate several strings related a content-filtered topic (content-filtered topic name, related topic name, filter class, filter expression, and all of the parameters). See RTPS 2.5 sec. 9.6.4.1 for more information on this.

rmw/include/rmw/rmw.h
asorbini: I wouldn't use the DDS spec as a reference on how to model content-filtering. Content-filtered Topics are one of the more contentious/unfortunate decisions in the spec. It is generally agreed that filters would have probably been better modeled as part of the DataReader QoS instead of as separate Topic entities. This is because introducing a new type of Topic complicates the object model and the API, while having the filter as an option of the reader (which it ultimately is) makes it much easier to use and integrate in the API.

This is probably an argument against subclassing Subscription with a CFT-aware interface and in favor of keeping the filter as an option for the Subscription, but I'll leave that discussion/decision to others, since from an RMW implementation point of view it will not make too much of difference (but it will likely perturb the RMW API a lot more than adding it as an option :) ).

rmw/include/rmw/subscription_content_filter_options.h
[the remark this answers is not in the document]
iuhilnehc-ynos: OK, update in ae1997dfad1c3a4f44b0a723eff00fa233a68fcf . 

@asorbini , could you update it for https://github.com/ros2/rmw_connextdds/pull/68? some locations like (https://github.com/ros2/rmw_connextdds/blob/883e9cf625699e59aa9e8f19ec7fa561cc47b13f/rmw_connextdds_common/src/common/rmw_impl.cpp#L1564, https://github.com/ros2/rmw_connextdds/blob/883e9cf625699e59aa9e8f19ec7fa561cc47b13f/rmw_connextdds_common/src/common/rmw_impl.cpp#L1276, etc.)

rmw/include/rmw/subscription_content_filter_options.h
fujitatomoya: @iuhilnehc-ynos 

NOT RELATED TO INTERFACE.

i think we do need to set the maximum token length, which i think we can change it later. but to support fallback filter in ROS 2 library, this behavior needs to be consistent.

- But the DDS1.4 specification shows

```
PARAMETER - A parameter is of the form %n, where n represents a natural number (zero included) smaller than 100.
It refers to the n + 1 th argument in the given context.
```

i think this can be applied for now.

rmw/include/rmw/subscription_content_filter_options.h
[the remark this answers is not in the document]
fujitatomoya: I will go ahead resolve this, since interface is adjusted with review comment.

rmw/include/rmw/rmw.h
fujitatomoya: I believe this can be resolved since we have consensus, see https://github.com/ros2/design/pull/282#issuecomment-1071170295

rmw/include/rmw/subscription_content_filter_options.h
fujitatomoya: @iuhilnehc-ynos i think this check should be applied with `rcl` when filtering expression and parameters come in from client code?

fujitatomoya:
@ivanpauno  @wjwwood 
CC: @MiguelCompany @asorbini 

could you do another final review before this goes in.

and i think that the followings need to be merged at the same time, right?

- https://github.com/ros2/rmw/pull/302 (this PR)
- https://github.com/ros2/rmw_implementation/pull/181
- https://github.com/ros2/rmw_connextdds/pull/68
- https://github.com/ros2/rmw_fastrtps/pull/513
- https://github.com/ros2/rmw_cyclonedds/pull/289

@iuhilnehc-ynos i think https://github.com/ros2/rclcpp/pull/1561 and https://github.com/ros2/rcl/pull/894 can be merged later w/o any problem? just checking cz 1st we need to have all green for above RMW changes before merge with mainline.

fujitatomoya:
CI only with CFT RMW Changes: (https://gist.githubusercontent.com/fujitatomoya/e591118d872be65c4c3821f7c710b12f/raw/2abba2d25836f36e6d0392024d649bd04556be56/ros2.repos)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16361/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10967/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16723/)

fujitatomoya:
@iuhilnehc-ynos @asorbini CI fails with rmw_connextdds, we probably want to have stub implementation for rmw_connextdds at this moment since we need to fix the APIs soon. then we can probably address unstable issues for rmw_connextdds later to enable CFT after release?

***Mon. March 21, 2022 - Alpha + RMW freeze****

fujitatomoya:
CI: same with https://github.com/ros2/rmw/pull/302#issuecomment-1071763251 but rmw_connextdds

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16363/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10969/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16725/)

fujitatomoya:
CI still fails, i think we need to rebase https://github.com/ros2/rmw_connextdds/tree/asorbini/cft as @iuhilnehc-ynos mentioned before https://github.com/ros2/design/pull/282#issuecomment-1066287946

rmw/include/rmw/subscription_content_filter_options.h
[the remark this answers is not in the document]
asorbini: @iuhilnehc-ynos I'll look into updating the interfaces while I'm working on investigating the test failures 👍🏼

fujitatomoya:
CI: (using https://github.com/ros2/rmw_connextdds/pull/77 rmw_connextdds stub, just to make sure this goes green.)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16366/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10972/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16728/)

repo is https://gist.githubusercontent.com/fujitatomoya/312c1308d8dcd48991be734c9c7d23f1/raw/664b83b531bc1b85faab9a042edf37329a795867/ros2.repos

iuhilnehc-ynos:
Rebased the 5 repositories, then re-run CI based on https://github.com/ros2/rmw/pull/302#issuecomment-1071971744

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16367/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10973/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16729/)

fujitatomoya:
windows CI unstable, https://ci.ros2.org/job/ci_windows/16730/, this is not related to CFT RMW fix.

@ivanpauno @wjwwood 
CC: @iuhilnehc-ynos 

could you help us to merge the following PRs for RMW CFT? we do not have access permission for cyclone and connextdds.

- https://github.com/ros2/rmw/pull/302 (this PR)
- https://github.com/ros2/rmw_implementation/pull/181
- https://github.com/ros2/rmw_connextdds/pull/77 (currently stub)
- https://github.com/ros2/rmw_fastrtps/pull/513
- https://github.com/ros2/rmw_cyclonedds/pull/289

rmw_connextdds is currently stub to push the interfaces, but the following can be pushed afterwards. (CC: @asorbini )
we tried to fix the problem, but for now we ended up having stub to get the interfaces at 1st.

- https://github.com/ros2/rmw_connextdds/pull/68 (CFT enabled, but some minor errors detected.)

rmw/src/subscription_content_filter_options.c
```diff
@@ -0,0 +1,160 @@
...
+    return ret;
+  }
+
+  return rmw_subscription_content_filter_options_init(
+    src->filter_expression,
+    src->expression_parameters.size,
+    (const char **)src->expression_parameters.data,
+    allocator,
+    dst
+  );
```
ivanpauno, line 137:
```suggestion
  return rmw_subscription_content_filter_options_set(
    src->filter_expression,
    src->expression_parameters.size,
    (const char **)src->expression_parameters.data,
    allocator,
    dst
  );
```

ivanpauno (approved):
LGTM!

rmw/include/rmw/subscription_content_filter_options.h
iuhilnehc-ynos: @fujitatomoya Did you mean adding a  check in the `rcl`?

fujitatomoya:
CI with https://github.com/ros2/rmw_connextdds/pull/68
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16372/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10981/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16742/)

rmw/include/rmw/subscription_content_filter_options.h
fujitatomoya: yes, this is NOT related to RMW CFT, but we should probably consistent argument check as ROS 2 for filtering expression and parameter length. unless, some filtering expression and parameters may or may not work based on rmw implementation.

fujitatomoya:
CI again with https://github.com/ros2/rmw_connextdds/pull/68 (see https://github.com/ros2/rmw_connextdds/pull/68#issuecomment-1073074721)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16374/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10983/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16745/)

asorbini:
I started another Windows build since I should have now resolved the build issues: [Build Status](https://ci.ros2.org/job/ci_windows/16746/)

fujitatomoya:
CI is green, i will merge this into mainline.

- https://github.com/ros2/rmw/pull/302#issuecomment-1072005358
- https://github.com/ros2/rmw/pull/302#issuecomment-1072572088

ivanpauno:
@fujitatomoya are the other PRs ready as well?

e.g. https://github.com/ros2/rmw_connextdds/pull/68 has some conflicts

fujitatomoya:
@ivanpauno we still have some unstable problems for https://github.com/ros2/rmw_connextdds/pull/68, so we pushed stub version https://github.com/ros2/rmw_connextdds/pull/77 to push the interfaces in the master.

fujitatomoya:
yeah, i will check the unstable reasons but i am not sure if we can fix it in today. (CC: @asorbini )

PR #308, opened 2021-04-12
Document which QoS policies are correctly read by rmw_get_publishers/subscriptions_info_by_topic

ivanpauno:
Based on https://github.com/ros2/rclcpp/pull/1626#issuecomment-816998500.

### DIFF
--- a/rmw/include/rmw/get_topic_endpoint_info.h
+++ b/rmw/include/rmw/get_topic_endpoint_info.h
@@ -115,13 +115,7 @@ rmw_get_publishers_info_by_topic(
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
  * \par QoS that are correctly read
- *   The QoS profiles returned might have some invalid fields.
- *   Particularly, the only QoS policies that are guaranteed to be shared
- *   during discovery are the ones that participate in endpoint matching.
- *   From the current QoS settings available, the only ones not shared by DDS based
- *   implementations are `history` and `history_depth`.
- *   The rmw implementation must set the settings that were not correctly read to
- *   `RMW_QOS_POLICY_*_UNKNOWN`.
+ * Not all QoS may be read correctly, \sa rmw_get_publishers_info_by_topic() for more details.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rmw/include/rmw/get_topic_endpoint_info.h
+++ b/rmw/include/rmw/get_topic_endpoint_info.h
@@ -115,10 +115,10 @@ rmw_get_publishers_info_by_topic(
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
  * \par QoS that are correctly read
- *   The qos profiles returned might have some invalid fields.
- *   Particularly, the only qos policies that are guaranteed to be shared
+ *   The QoS profiles returned might have some invalid fields.
+ *   Particularly, the only QoS policies that are guaranteed to be shared
  *   during discovery are the ones that participate in endpoint matching.
- *   From the current qos settings available, the only ones not shared by DDS based
+ *   From the current QoS settings available, the only ones not shared by DDS based
  *   implementations are `history` and `history_depth`.
  *   The rmw implementation must set the settings that were not correctly read to
  *   `RMW_QOS_POLICY_*_UNKNOWN`.
--- a/rmw/include/rmw/get_topic_endpoint_info.h
+++ b/rmw/include/rmw/get_topic_endpoint_info.h
@@ -33,12 +33,11 @@ extern "C"
  *
  * \par QoS that are correctly read
  * The QoS profiles returned might have some invalid fields.
- * Particularly, the only QoS policies that are guaranteed to be shared
+ * The rmw implementation must set the invalid fields to `RMW_QOS_POLICY_*_UNKNOWN`.
+ * For DDS based implementations, the only QoS policies that are guaranteed to be shared
  * during discovery are the ones that participate in endpoint matching.
  * From the current QoS settings available, the only ones not shared by DDS based
  * implementations are `history` and `history_depth`.
- * The rmw implementation must set the settings that were not correctly read to
- * `RMW_QOS_POLICY_*_UNKNOWN`.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rmw/include/rmw/get_topic_endpoint_info.h
+++ b/rmw/include/rmw/get_topic_endpoint_info.h
@@ -31,6 +31,15 @@ extern "C"
  * the associated topic type, the publisher's gid, and the publisher QoS profile.
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
+ * \par QoS that are correctly read
+ *   The qos profiles returned might have some invalid fields.
+ *   Particularly, the only qos policies that are guaranteed to be shared
+ *   during discovery are the ones that participate in endpoint matching.
+ *   From the current qos settings available, the only ones not shared by DDS based
+ *   implementations are `history` and `history_depth`.
+ *   The rmw implementation must set the settings that were not correctly read to
+ *   `RMW_QOS_POLICY_*_UNKNOWN`.
+ *
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/get_topic_endpoint_info.h
```diff
@@ -31,6 +31,15 @@ extern "C"
  * the associated topic type, the publisher's gid, and the publisher QoS profile.
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
+ * \par QoS that are correctly read
+ *   The qos profiles returned might have some invalid fields.
+ *   Particularly, the only qos policies that are guaranteed to be shared
+ *   during discovery are the ones that participate in endpoint matching.
+ *   From the current qos settings available, the only ones not shared by DDS based
```
jacobperron, line 38:
```suggestion
 *   The QoS profiles returned might have some invalid fields.
 *   Particularly, the only QoS policies that are guaranteed to be shared
 *   during discovery are the ones that participate in endpoint matching.
 *   From the current QoS settings available, the only ones not shared by DDS based
```

rmw/include/rmw/get_topic_endpoint_info.h
```diff
@@ -105,6 +114,15 @@ rmw_get_publishers_info_by_topic(
  * the associated topic type, the subscription's gid, and the subscription QoS profile.
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
+ * \par QoS that are correctly read
+ *   The qos profiles returned might have some invalid fields.
+ *   Particularly, the only qos policies that are guaranteed to be shared
+ *   during discovery are the ones that participate in endpoint matching.
+ *   From the current qos settings available, the only ones not shared by DDS based
+ *   implementations are `history` and `history_depth`.
```
jacobperron, line 122:
```suggestion
 *   The QoS profiles returned might have some invalid fields.
 *   Particularly, the only QoS policies that are guaranteed to be shared
 *   during discovery are the ones that participate in endpoint matching.
 *   From the current QoS settings available, the only ones not shared by DDS based
 *   implementations are `history` and `history_depth`.
```

jacobperron (approved):
LGTM

I guess some RMW implementations should probably be updated to match this documentation. In that case, this may be considered an API change and perhaps we should hold until after we branch for Galactic.

ivanpauno:
> LGTM
> 
> I guess some RMW implementations should probably be updated to match this documentation. In that case, this may be considered an API change and perhaps we should hold until after we branch for Galactic.

The rmw implementations already match this behavior, so this is only documentation.

I don't remember if they set the policies that are not updated to `RMW_QOS_POLICY_*_UNKNOWN`, so I can delete that line of the docs.
But I think they should be doing that, if not it's hard/impossible to understand the output.

rmw/include/rmw/get_topic_endpoint_info.h
```diff
@@ -31,6 +31,15 @@ extern "C"
...
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
+ * \par QoS that are correctly read
+ *   The QoS profiles returned might have some invalid fields.
+ *   Particularly, the only QoS policies that are guaranteed to be shared
+ *   during discovery are the ones that participate in endpoint matching.
+ *   From the current QoS settings available, the only ones not shared by DDS based
+ *   implementations are `history` and `history_depth`.
+ *   The rmw implementation must set the settings that were not correctly read to
+ *   `RMW_QOS_POLICY_*_UNKNOWN`.
```
wjwwood, line 41: I don't think this needs to be indented.

Based on this:

https://www.star.bnl.gov/public/comp/sofi/doxygen/commands.html#cmdpar

Any you just want `\par` for the "heading" right?

```suggestion
 *  The QoS profiles returned might have some invalid fields.
 *  Particularly, the only QoS policies that are guaranteed to be shared
 *  during discovery are the ones that participate in endpoint matching.
 *  From the current QoS settings available, the only ones not shared by DDS based
 *  implementations are `history` and `history_depth`.
 *  The rmw implementation must set the settings that were not correctly read to
 *  `RMW_QOS_POLICY_*_UNKNOWN`.
```

wjwwood: I also wonder if the `section`/`subsection`/`subsubsection` commands are a better fit?

https://www.star.bnl.gov/public/comp/sofi/doxygen/commands.html#cmdsection

rmw/include/rmw/get_topic_endpoint_info.h
```diff
@@ -31,6 +31,15 @@ extern "C"
  * the associated topic type, the publisher's gid, and the publisher QoS profile.
  * Names of non-existent topics are allowed, in which case an empty array will be returned.
  *
+ * \par QoS that are correctly read
+ *   The QoS profiles returned might have some invalid fields.
+ *   Particularly, the only QoS policies that are guaranteed to be shared
+ *   during discovery are the ones that participate in endpoint matching.
```
wjwwood, line 37: This is specific to DDS, so I'd prefer to have this documentation read as "here's the general guarantee" and then "as an example, here's some specifics for DDS implementations". You kind of do that on the line right after this one.

rmw/include/rmw/get_topic_endpoint_info.h
wjwwood, line 41: Here's a suggestion based on my previous review comments:

```suggestion
 * \section QoS that are correctly read
 *
 * The QoS profiles returned may have some invalid fields when the
 * rmw implementation does not have the required information.
 * The rmw implementation must set these settings to
 * `RMW_QOS_POLICY_*_UNKNOWN` in order to indicate this.
 *
 * \subsection Expected behavior for DDS based rmw implementations
 *
 * In particular for DDS based rmw implementations, the only QoS policies
 * that are guaranteed to be shared during discovery are the ones that
 * participate in SEDP endpoint matching.
 * From the current QoS settings available, the only ones not shared by DDS based
 * rmw implementations are `history` and `history_depth`.
```

rmw/include/rmw/get_topic_endpoint_info.h
```diff
@@ -105,6 +114,15 @@ rmw_get_publishers_info_by_topic(
...
+ *   From the current QoS settings available, the only ones not shared by DDS based
+ *   implementations are `history` and `history_depth`.
+ *   The rmw implementation must set the settings that were not correctly read to
+ *   `RMW_QOS_POLICY_*_UNKNOWN`.
```
wjwwood, line 124: Rather than repeating ourselves, maybe we just say something like "not all QoS may be read correctly, \sa ... for more details"?

rmw/include/rmw/get_topic_endpoint_info.h
ivanpauno: I copied the style from another part of the same docblock https://github.com/ros2/rmw/blob/aa3a366a59d525ce54e1554b39f4721336601344/rmw/include/rmw/get_topic_endpoint_info.h#L52-L69

\section and \subsection seem to only work when combined with the \page command:

> Warning:
>   This command only works inside related page documentation and not in other documentation blocks!

I agree that the indentation is unnecesary though.

rmw/include/rmw/get_topic_endpoint_info.h
ivanpauno: Done in https://github.com/ros2/rmw/pull/308/commits/91c794aeb2d4d80c1b0f94453eb43be5c979b292

rmw/include/rmw/get_topic_endpoint_info.h
ivanpauno: See https://github.com/ros2/rmw/pull/308#discussion_r611952159.

rmw/include/rmw/get_topic_endpoint_info.h
ivanpauno: Clarified in https://github.com/ros2/rmw/pull/308/commits/bc07bf9cbeaeb9824493c3d7b8e1cb3be5ead379.

wjwwood (approved):

ivanpauno:
> I don't remember if they set the policies that are not updated to `RMW_QOS_POLICY_*_UNKNOWN`, so I can delete that line of the docs.
> But I think they should be doing that, if not it's hard/impossible to understand the output.

Fastrtps was already doing that https://github.com/ros2/rmw_fastrtps/blob/7bb3563bebd20c3cae03231c2e321bf132651449/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_participant_info.hpp#L149.
Cyclonedds shares through discovery all qos policies.

I send a patch to connextdds to fix this https://github.com/ros2/rmw_connextdds/pull/28.

hidmic (approved):

ivanpauno:
PR checker is happy, this a docs only PR so no need of more CI.

PR #311, opened 2021-08-10
Fix up errors in doxygen documentation.

clalancette:
While running the rosdoc2 tool against this repository, I
came across these documentation errors.  This PR fixes them
all up so that doxygen runs clean.  We still can't successfully
run rosdoc2 on this package due to one other error, but at
least this will get us ready once we figure that issue out.

### DIFF
--- a/rmw/include/rmw/event.h
+++ b/rmw/include/rmw/event.h
@@ -66,9 +66,9 @@ rmw_get_zero_initialized_event(void);
 
 /// Initialize a rmw publisher event.
 /**
- * \param[in|out] rmw_event to initialize
+ * \param[inout] rmw_event to initialize
  * \param[in] publisher to initialize with
- * \param[in|out] event_type for the event to initialize
+ * \param[inout] event_type for the event to initialize
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if invalid argument, or
  * \return `RMW_RET_UNSUPPORTED` if event_type is not supported, or
@@ -84,9 +84,9 @@ rmw_publisher_event_init(
 
 /// Initialize a rmw subscription event.
 /**
- * \param[in|out] rmw_event to initialize
+ * \param[inout] rmw_event to initialize
  * \param[in] subscription to initialize with
- * \param[in|out] event_type for the event to handle
+ * \param[inout] event_type for the event to handle
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if invalid argument, or
  * \return `RMW_RET_UNSUPPORTED` if event_type is not supported, or
@@ -103,7 +103,7 @@ rmw_subscription_event_init(
 /// Take an event from the event handle.
 /**
  * \param[in] event_handle event object to take from
- * \param[in|out] event_info event info object to write taken data into
+ * \param[inout] event_info event info object to write taken data into
  * \param[out] taken boolean flag indicating if an event was taken or not
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_BAD_ALLOC` if memory allocation failed, or
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -745,7 +745,7 @@ rmw_publisher_get_actual_qos(
  *   one registered with `publisher` on creation.
  *
  * \param[in] publisher Publisher to be used to send message.
- * \param[in] ros_message Serialized ROS message to be sent.
+ * \param[in] serialized_message Serialized ROS message to be sent.
  * \param[in] allocation Pre-allocated memory to be used. May be NULL.
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `publisher` is NULL, or
@@ -767,7 +767,7 @@ rmw_publish_serialized_message(
  * Given a message definition and bounds, compute the serialized size.
  *
  * \param[in] type_support The type support of the message to compute.
- * \param[in] bounds Artifical bounds to use on unbounded fields.
+ * \param[in] message_bounds Artifical bounds to use on unbounded fields.
  * \param[out] size The computed size of the serialized message.
  * \return `RMW_RET_OK` if successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if either argument is null, or
@@ -1001,7 +1001,7 @@ rmw_fini_subscription_allocation(
  * \param[in] topic_name Name of the topic to subscribe to, often a fully qualified
  *   topic name unless `qos_profile` is configured to avoid ROS namespace conventions
  *   i.e. to create a native topic subscription
- * \param[in] qos_profile QoS policies for this subscription
+ * \param[in] qos_policies QoS policies for this subscription
  * \param[in] subscription_options Options for configuring this subscription
  * \return rmw subscription handle, or `NULL` if there was an error
  */
@@ -1858,7 +1858,7 @@ rmw_return_loaned_message_from_subscription(
  * \param[in] service_name Name of the service to be used, often a fully qualified
  *   service name unless `qos_profile` is configured to avoid ROS namespace conventions
  *   i.e. to create a native service client.
- * \param[in] qos_profile QoS policies for this service client's connections.
+ * \param[in] qos_policies QoS policies for this service client's connections.
  * \return rmw service client handle, or `NULL` if there was an error.
  */
 RMW_PUBLIC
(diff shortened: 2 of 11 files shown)


rmw/include/rmw/event.h
```diff
@@ -66,9 +66,9 @@ rmw_get_zero_initialized_event(void);
 
 /// Initialize a rmw publisher event.
 /**
- * \param[in|out] rmw_event to initialize
+ * \param[inout] rmw_event to initialize
```
aprotyas, line 69:
```suggestion
 * \param[in,out] rmw_event to initialize
```

rmw/include/rmw/event.h
```diff
@@ -66,9 +66,9 @@ rmw_get_zero_initialized_event(void);
...
  * \param[in] publisher to initialize with
- * \param[in|out] event_type for the event to initialize
+ * \param[inout] event_type for the event to initialize
```
aprotyas, line 71:
```suggestion
 * \param[in,out] event_type for the event to initialize
```

rmw/include/rmw/event.h
```diff
@@ -84,9 +84,9 @@ rmw_publisher_event_init(
 
 /// Initialize a rmw subscription event.
 /**
- * \param[in|out] rmw_event to initialize
+ * \param[inout] rmw_event to initialize
```
aprotyas, line 87:
```suggestion
 * \param[in,out] rmw_event to initialize
```

rmw/include/rmw/event.h
```diff
@@ -84,9 +84,9 @@ rmw_publisher_event_init(
...
  * \param[in] subscription to initialize with
- * \param[in|out] event_type for the event to handle
+ * \param[inout] event_type for the event to handle
```
aprotyas, line 89:
```suggestion
 * \param[in,out] event_type for the event to handle
```

rmw/include/rmw/event.h
```diff
@@ -103,7 +103,7 @@ rmw_subscription_event_init(
 /// Take an event from the event handle.
 /**
  * \param[in] event_handle event object to take from
- * \param[in|out] event_info event info object to write taken data into
+ * \param[inout] event_info event info object to write taken data into
```
aprotyas, line 106:
```suggestion
 * \param[in,out] event_info event info object to write taken data into
```

rmw/include/rmw/security_options.h
```diff
@@ -70,7 +70,7 @@ rmw_security_options_copy(
  *
  * \param[in] security_root_path path to be set.
  * \param[in] allocator allocator used to allocate the new path.
- * \param[in|out] security_options security options to be set.
+ * \param[inout] security_options security options to be set.
```
aprotyas, line 73:
```suggestion
 * \param[in,out] security_options security options to be set.
```

aprotyas (commented):
Looks good overall. I couldn't find `\param[inout]` in the [Doxygen commands manual](https://www.doxygen.nl/manual/commands.html#cmdparam), so I suggested the batch of changes.

wjwwood (commented):
I don't feel strongly about `[inout]` vs `[in,out]`, but I think we use `[inout]` elsewhere like in rclcpp.

rmw/include/rmw/types.h
```diff
@@ -159,13 +159,6 @@ typedef struct RMW_PUBLIC_TYPE rmw_subscription_options_t
    * This setting is most often used when data should only be received from
    * remote nodes, especially to avoid "double delivery" when both intra- and
    * inter- process communication is taking place.
-   *
-   * \TODO(wjwwood): nail this down when participant mapping is sorted out.
-   *   See: https://github.com/ros2/design/pull/250
-   *
-   * The definition of local is somewhat vague at the moment.
-   * Right now it means local to the node, and that definition works best, but
-   * may become more complicated when/if participants map to a context instead.
```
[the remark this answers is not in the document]
clalancette: The original thing that caused me to look at this was the `\TODO`, which apparently doxygen doesn't understand as a tag.

That said, I believe that the situation now is different from what this comment says.  We definitely did all of the one-participant-per-context stuff back in Foxy (and finished it in Galactic), so the comment is outdated.  I didn't know what to replace it with, so I figured nothing was better than the wrong thing.

That said, if you have additional context, I'm happy to update this to something else.  Maybe @ivanpauno can weigh in as well.

clalancette:
> I don't feel strongly about `[inout]` vs `[in,out]`, but I think we use `[inout]` elsewhere like in rclcpp.

Yeah, that's true.  I was following that pattern here, though @aprotyas is correct about Doxygen suggesting `[in,out]`.  Is it better to be consistent with our style elsewhere, or correct in terms of the Doxygen documentation?  I'm not sure.  My inclination is to be consistent, and leave this as `[inout]`.  Maybe in the future we can do a pass where we update all of them (but I also don't feel strongly about this point).

rmw/include/rmw/rmw.h
```diff
@@ -2041,8 +2041,8 @@ rmw_send_request(
  *   It will also be left unchanged if this function succeeds but `taken` is false.
  *
  * \param[in] client Service client to take response from.
- * \param[out] response_header Service response header to write to.
- * \param[out] ros_request Type erased ROS service response to write to.
+ * \param[out] request_header Service response header to write to.
```
christophebedard, line 2044: I think this should be `response_header` (so the name in the signature needs to be updated). See the doc above.

wjwwood:
Yeah, I don't mind about the inout thing. I would say just leave it inout for consistency.

ivanpauno (approved):

rmw/include/rmw/rmw.h
clalancette: So I looked into this a bit.  I think I agree with you on principle.  However, all implementors (`rmw_cyclonedds`, `rmw_fastrtps_cpp`, `rmw_connextdds`) call this `request_header`, and all callers (`rcl`) also call it `request_header`.  That may all be down to the fact that this API named it that, but I'd rather not confuse the situation and change it here.

wjwwood: The `request_header` is the one passed originally when the request was made, so it's indicating which request this response that you're taking goes with. So I think the name is right.

wjwwood (approved):
lgtm

rmw/include/rmw/rmw.h
christophebedard: > The `request_header` is the one passed originally when the request was made, so it's indicating which request this response that you're taking goes with.

I don't think that's right unfortunately. It tripped me up at first when implementing `rmw_take_response` recently; I thought I could use the header's request ID/sequence number to get the corresponding response, but that's not it. By looking at the implementations, this header parameter is strictly an `out` param, so `rmw_take_response` takes any response and not a specific response for a specific request. That response is then matched with the original request elsewhere (like `rclcpp`) using the request ID from the header.

The documentation here and in `rcl` is kind of confusing/contradictory, which I why I commented when I saw this. `request_header` should be changed to `response_header`, but as @clalancette said, it's called `request_header` pretty much everywhere.

wjwwood: Sorry, I didn't mean to say the `request_header` is an input for this take function, but that it is an output, as you said.

I see why you're confused, but I don't think that `request_header` is wrong, because it really contains information about the originating request. This information is about the request, as in which request is associated with this response and as you said it is used later to associate the response with an original request by either the user or `rclcpp`.

That being said, it does contain information about the response specifically, like the source and destination timestamps. So maybe the name should be `response_header` or `response_info` (I think info is a better term, to match `rcl_take_with_info` for example, but that's not really important).

wjwwood: And `rmw_service_info_t` could more likely be named `rmw_service_response_info_t`, but for all of this the question is if changing the names is worth the churn. I'd lean towards yes (clearer is always better, also helps us feel proud in knowing what we make/maintain makes sense), but also it could be disruptive and work.

christophebedard: Yeah, renaming would make all of this confusion go away. If you think it's worth it, then I can probably do it at some point in the future. Otherwise, I'll just stop here :stuck_out_tongue_closed_eyes:

wjwwood: Changing the parameter name is not a big deal, but changing the type name for the struct is a larger change, though a valuable one I think.

I'm fine with changing the variable name, but we should do that in another pr, since this one is about fixing the documentation for what's already there. @clalancette thoughts?

clalancette: Yes, I'm fine with changing the variable name (this will have no material API/ABI impact, but will make things clearer).

If we do that, I think we should change the upstream and downstreams of this as well, so everything is consistent.

But all of that should be in a separate set of PRs.

clalancette:
I believe the conversation in https://github.com/ros2/rmw/pull/311#discussion_r687108023 is wrapped up (with further work to come, but in a different PR).

With that said, this now has approvals.  It also has green PR job.  I'm going to skip CI since this is a documentation change only.  Thanks for the reviews!

rmw/include/rmw/rmw.h
christophebedard: I opened an issue to track this: https://github.com/ros2/rmw/issues/312

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-08-19/22008/1

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

PR #313, opened 2021-09-07
Fix up documentation build for rmw when using rosdoc2

hidmic:
Precisely what the title says. This patch:

- Substitutes trailing `[]` tokens for leading `*` tokens for pointer argument types.
- Normalizes type naming to preclude duplication warnings.
- Handles `RCUTILS_DEPRECATED_WITH_MSG()` macros
- Handles `RMW_PUBLIC_TYPE` macros.
- De-duplicates some declarations.


hidmic:
Full CI!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15165/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9894/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12854/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15348/)

hidmic:
Alright, it wasn't just `rmw_connextdds_cpp`.

Re-running full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15174/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9901/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12861/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15355/)

ivanpauno:
> De-duplicates some declarations.

Is there a way to make doxygen ignore that instead of renaming *_t to *_s.
I don't mind much, but it seems a bit unnecessary to break API that way.
Though, I think we're never using `struct *_t` in code and always `*_t` directly.

rmw/include/rmw/rmw.h
```diff
@@ -297,12 +302,6 @@ rmw_ret_t
 rmw_fini_publisher_allocation(
   rmw_publisher_allocation_t * allocation);
 
-/// Return a rmw_publisher_options_t initialized with default values.
-RMW_PUBLIC
-RMW_WARN_UNUSED
-rmw_publisher_options_t
-rmw_get_default_publisher_options(void);
```
ivanpauno, line 304: was this definition duplicated?

hidmic:
> Is there a way to make doxygen ignore that instead of renaming *_t to *_s.

I followed https://github.com/ros2/rcutils/pull/333 precedent here, but perhaps `\cond` based suppression of `typedef`s is possible e.g.:

```
/// \cond flag
typedef
/// \endcond
struct type_name_t {

}
/// \cond flag
type_name_t
/// \endcond
;
```

Though I do find duplicate type definitions awkward albeit legal.

rmw/include/rmw/rmw.h
hidmic: Indeed:
https://github.com/ros2/rmw/blob/c71e97a670717cbf755627723a6d7ef5d756c4f4/rmw/include/rmw/publisher_options.h#L27-L31
Likely a leftover of a past refactor.

ivanpauno:
> I followed ros2/rcutils#333 precedent here, but perhaps \cond based suppression of typedefs is possible e.g.:

That seems to be enough precedent, we will have to bump the major number next release.

ivanpauno (approved):

rmw/include/rmw/qos_string_conversions.h
```diff
@@ -73,31 +73,31 @@ rmw_qos_policy_kind_to_str(rmw_qos_policy_kind_t kind);
  */
 RMW_PUBLIC
 const char *
-rmw_qos_durability_policy_to_str(enum rmw_qos_durability_policy_t value);
+rmw_qos_durability_policy_to_str(enum rmw_qos_durability_policy_e value);
```
clalancette, line 76: I think I'd rather use the `_t` types everywhere as a policy.  That is, this would be:

```
rmw_qos_durability_policy_t value
```

That matches more closely what we do with structs in most places.  What do you think?

hidmic: That's reasonable. See 0e9b4c282f6bc2dbfd9750348395dcdf0187672d.

clalancette (approved):
This all looks really good to me with green CI.

Thanks for iterating!

rmw/src/topic_endpoint_info.c
```diff
@@ -202,7 +202,7 @@ rmw_topic_endpoint_info_set_endpoint_type(
 rmw_ret_t
 rmw_topic_endpoint_info_set_gid(
   rmw_topic_endpoint_info_t * topic_endpoint_info,
-  const uint8_t gid[],
+  const uint8_t * gid,
```
wjwwood, line 205: Why this change? (just curious)

rmw/include/rmw/event.h
```diff
@@ -30,7 +30,7 @@ extern "C"
 #include "rmw/visibility_control.h"
 
 /// Define publisher/subscription events
-typedef enum rmw_event_type_t
+typedef enum rmw_event_type_e
```
wjwwood, line 33: Will this break rmw implementations?

clalancette: Any RMW implementation that was using a "bare" `rmw_event_type_t` before should be fine; we are still defining that as part of the typedef.  If an RMW implementation was using `enum rmw_event_type_t`, then it would now be broken.

I don't think we have any users of that in the core, but I will request a full CI run before we merge this.  There might be third-party ones that use `enum rmw_event_type_t`, so I think a release note is in order for this change.

rmw/src/topic_endpoint_info.c
hidmic: > WARNING: doxygenfunction: Unable to resolve function "rmw_topic_endpoint_info_set_gid" with arguments (rmw_topic_endpoint_info_t*, const uint8_t, size_t) in doxygen xml output for project "rmw Doxygen Project" from directory: ...
Potential matches:
> - rmw_ret_t rmw_topic_endpoint_info_set_gid(rmw_topic_endpoint_info_t *topic_endpoint_info, const uint8_t gid[], size_t size)

It appears `exhale` fails to pick up a trailing `[]` as a type modifier. I haven't dug into it deep enough to find out why.

hidmic:
Re-running full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15218/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9938/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12898/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15400/)

hidmic:
Alright, all green !

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

PR #318, opened 2022-03-03
Add sequence numbers to message info structure

ivanpauno:
* Add publication/reception sequence numbers to `rmw_message_info_t`.
* Add `rmw_feature_supported()` function, to query if the rmw implementation being used supports a feature.

Related PRs:

- https://github.com/ros2/rmw_implementation/pull/204
- https://github.com/ros2/rmw_fastrtps/pull/587
- https://github.com/ros2/rmw_connextdds/pull/74
- https://github.com/ros2/rmw_cyclonedds/pull/381

---

This information is not provided by all DDS vendors.
Both sequence numbers are an extension, and not specified by the DDS standard AFAIU.

Connext provides both numbers, see `publication_sequence_number` and `reception_sequence_number` [here](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/The_SampleInfo_Structure.htm).

FastDDS provides the `publication_sequence_number`, but not the `reception_sequence_number`.
See [here](https://fast-dds.docs.eprosima.com/en/latest/fastdds/api_reference/dds_pim/subscriber/sampleinfo.html#_CPPv4N8eprosima7fastdds3dds10SampleInfo15sample_identityE) and [here](https://github.com/eProsima/Fast-DDS/blob/e0ba220d878b5f3ba858256bb72332128cd68033/include/fastdds/rtps/common/SampleIdentity.h#L150) (docs aren't super clear, but based on some testing that's the publication sequence number).

CycloneDDS doesn't seem to provide any of both (see [here](https://github.com/eclipse-cyclonedds/cyclonedds/blob/01eafd432adc56272aa8324823f1d2fcb8a30521/src/core/ddsc/include/dds/dds.h#L152-L182)).


christophebedard:
> * Do we need to add both? Is only having the publication sequence number enough?

What's the intended use-case?

I've used `source_timestamp` as a kind of unique message ID. Obviously, I could get collisions if two publishers publish at the same exact time on the same exact topic (from different machines, or even the same process/node). Getting `source_timestamp`s does require collecting it from the DDS implementation on the publisher side, since it's not provided to `rmw`, so `publication_sequence_number`s at the `rmw` level could replace this.

Could we have collisions with (publication) sequence numbers? How are they generated, and what are the guarantees? Since it's not a standard DDS feature, and since rmw is not DDS-only(:tm:), `rmw` probably needs to have its own definition or requirements for these sequence numbers. Also, do reception sequence numbers provide anything that publication sequences numbers do not provide? Maybe I'm not thinking of the right use-case.

ivanpauno:
> Could we have collisions with (publication) sequence numbers?

Yes, each publisher will use unique sequence numbers, but sample of different publishes may use the same.
You can disambiguate with the publisher gid though (which is available in the sample info struct).
After a long long time, the sequence number may wrap around IIUC.

>  Getting source_timestamps does require collecting it from the DDS implementation on the publisher side, since it's not provided to rmw

I think that's already part of the message info struct, though IIRC it's not being filled by all the DDS vendors.

> do reception sequence numbers provide anything that publication sequences numbers do not provide?

I think they don't provide much extra info, but I'm not sure.

> What's the intended use-case?

IIUC, the idea is to use this to collect statistics (which messages were lost, etc).
@wjwwood can probably expand more.

There's also the [sample lost status](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Statuses_for_DataReaders.htm#receiving_2076951295_607362) and [sample rejected status](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Statuses_for_DataReaders.htm#receiving_2076951295_607403) in DDS, that can be used for this. We already have bindings for the first one ([this](https://github.com/ros2/rmw/blob/bceeda0a2a1d1451e1e5243fba09aaa3811f33cd/rmw/include/rmw/events_statuses/message_lost.h#L27-L33)).
The sequence numbers might be more insightful to make statistics than those statuses, but I think that having only the publication sequence number should be enough.

> How are they generated, and what are the guarantees?

This is a good question, because I actually don't know if the "publication_sequence_number" provided by FastDDS has the same guarantees that the one provided by Connext.
Based on some basic testing, they seem to be the same thing.

rmw/include/rmw/types.h
```diff
@@ -518,8 +518,21 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
 /// Information describing an rmw message
 typedef struct RMW_PUBLIC_TYPE rmw_message_info_s
 {
+  /// Time when the DDS sample was written in the publisher queue.
```
wjwwood, line 521: Let's not include the DDS specifics here and remember that it needs to be flexible in case we're not using DDS. Maybe:

```suggestion
  /// Time when the message was published by the publisher.
  /**
   * The exact point at which the timestamp is taken is not specified, but
   * it should be taken consistently at the same point in the
   * publishing process each time.
   */
```

rmw/include/rmw/types.h
```diff
@@ -518,8 +518,21 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
   rmw_time_point_value_t source_timestamp;
+  /// Time when the message was added to the subscription queue.
```
wjwwood, line 523:
```suggestion
  /// Time when the message was received by the subscription.
  /**
   * The exact point at which the timestamp is taken is not specified, but
   * it should be taken consistently at the same point in the
   * process of receiving a message each time.
```

rmw/include/rmw/types.h
```diff
@@ -518,8 +518,21 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
   rmw_time_point_value_t received_timestamp;
+  /// Sequence number of the received message for the publisher.
+  /**
+   * If the rmw implementation doesn't support sequence numbers, it's value will be UINT64_MAX.
```
wjwwood, line 527: Should be make a `#define` for this rather than `UINT64_MAX`? In case it needs to change in the future?

rmw/include/rmw/types.h
```diff
@@ -518,8 +518,21 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   */
```
wjwwood, line 528:
```suggestion
  /// Sequence number of the received message set by the publisher.
  /**
   * This sequence number is set by the publisher and therefore uniquely identifies
   * a message when combined with the publisher GID.
   *
   * If the rmw implementation doesn't support sequence numbers, it's value will be UINT64_MAX.
   */
```

rmw/include/rmw/types.h
```diff
@@ -518,8 +518,21 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+  int64_t publication_sequence_number;
+  /// Sequence number of the received message for the subscription,
+  /**
+   * If the rmw implementation doesn't support sequence numbers, it's value will be UINT64_MAX.
+   */
```
wjwwood, line 533:
```suggestion
  /// Sequence number of the received message set by the subscription.
  /**
   * This sequence number is set by the subscription regardless of which
   * publish sent the message.
   *
   * If the rmw implementation doesn't support sequence numbers, it's value will be UINT64_MAX.
   */
```

wjwwood (commented):
lgtm in general.

Do we need a method to check if this feature is supported separate from the max value? I'm just thinking that it might be useful for a tool to know if this is supported before receiving a message (or even creating a publisher/subscription).

christophebedard:
> Yes, each publisher will use unique sequence numbers, but sample of different publishes may use the same. You can disambiguate with the publisher gid though (which is available in the sample info struct)

Oh right, that's a great point.

> I think that's already part of the message info struct, though IIRC it's not being filled by all the DDS vendors.

Unless I'm mistaken, I don't think the source timestamp is made available to ROS 2 (`rmw` and up) on the publisher side (there is no "message info" when publishing a message). The source timestamp is generated inside the DDS implementation right before writing/sending a message. However, it is indeed available on the receiver side.

> IIUC, the idea is to use this to collect statistics (which messages were lost, etc). @wjwwood can probably expand more.
> 
> There's also the [sample lost status](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Statuses_for_DataReaders.htm#receiving_2076951295_607362) and [sample rejected status](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Statuses_for_DataReaders.htm#receiving_2076951295_607403) in DDS, that can be used for this. We already have bindings for the first one ([this](https://github.com/ros2/rmw/blob/bceeda0a2a1d1451e1e5243fba09aaa3811f33cd/rmw/include/rmw/events_statuses/message_lost.h#L27-L33)). The sequence numbers might be more insightful to make statistics than those statuses, but I think that having only the publication sequence number should be enough.

This makes sense! It's similar to what I'm doing.

> This is a good question, because I actually don't know if the "publication_sequence_number" provided by FastDDS has the same guarantees that the one provided by Connext. Based on some basic testing, they seem to be the same thing.

I'm wondering what downstream users of `rmw` can expect, i.e., if this is part of the interface/abstraction, or if the behaviour is always going to come down to the middleware implementation.

ivanpauno:
> Unless I'm mistaken, I don't think the source timestamp is made available to ROS 2 (rmw and up) on the publisher side

Ah yes, it's only available in the subscription side at the moment (and supported only by some rmw implementations).

ivanpauno:
> Do we need a method to check if this feature is supported separate from the max value? I'm just thinking that it might be useful for a tool to know if this is supported before receiving a message (or even creating a publisher/subscription).

That sounds okay to me.
What's the best API for this?
I think that maybe

```c
bool rmw_message_info_publication_sequence_number_supported(void);
bool rmw_message_info_subscription_sequence_number_supported(void);
```

can easily become too verbose if we start adding similar functions for other features (e.g. is `source_timestamp` supported?),

Maybe:

```c
bool rmw_feature_supported(rmw_feature_t feature_v);

typedef enum rmw_feature_e {
  RMW_FEATURE_MESSAGE_INFO_PUBLICATION_SEQUENCE_NUMBER,
  ...
} rmw_feature_t;
```

is a better idea.

rmw/include/rmw/types.h
ivanpauno: Yeah sounds good.
I thought mentioning the queue wasn't a bad idea, as we somehow acknowledge its existance with the qos settings.
But the moment other rmw implementations create the timestamp may vary.

rmw/include/rmw/types.h
ivanpauno: yes, I already changed the type to match https://github.com/ros2/rmw/blob/bceeda0a2a1d1451e1e5243fba09aaa3811f33cd/rmw/include/rmw/types.h#L355.
So that's not a bad idea.

rmw/include/rmw/types.h
ivanpauno: Should I mention wrap around in the docs as well?
Though you need to have the application running for a *really* long time for that to happen, I think technically it can.

rmw/include/rmw/types.h
ivanpauno: See https://github.com/ros2/rmw/pull/318/commits/3ad6037a79b0e14b37becc3cf1f8f2c101ae12c5

rmw/include/rmw/types.h
ivanpauno: I have added that, just in case.
See https://github.com/ros2/rmw/pull/318/commits/7810f901cf1a845814bbc6b21a3390b40a5adf39.

ivanpauno:
> I'm wondering what downstream users of rmw can expect, i.e., if this is part of the interface/abstraction, or if the behaviour is always going to come down to the middleware implementation.

I have listed some requirements of what clients can expect in https://github.com/ros2/rmw/pull/318/commits/7810f901cf1a845814bbc6b21a3390b40a5adf39.
Maybe I have specified more than needed or we want to add extra requirements.

Please feel free to suggest improvements to my wording as well!

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,66 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   *   received messages.
+   *   Those might have already been taken by other `rmw_take*()` calls that happened in between or lost.
+   *   `psn2 - psn1 - 1 = 0` if and only if the messages were sent by the publisher consecutively.
+   */
+  int64_t publication_sequence_number;
+  /// Sequence number of the received message set by the subscription.
+  /**
+   * This sequence number is set by the subscription regardless of which
+   * publish sent the message.
+   * For long running applications, the sequence number might wrap arround at some point.
```
christophebedard, line 563:
```suggestion
   * For long running applications, the sequence number might wrap around at some point.
```

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,66 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   *
+   * If the rmw implementation doesn't support sequence numbers, it's value will be
```
christophebedard, line 565:
```suggestion
   * If the rmw implementation doesn't support sequence numbers, its value will be
```

christophebedard (commented):
The requirements make sense to me.

wjwwood:
The `rmw_feature_supported()` route sounds good to me.

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,66 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   *
+   * If `rsn1` and `rsn2` are the reception sequence numbers obtained by
+   * calls to `rmw_take*()`, where `rsn1` was obtained in a call that happened before `rsn2`, then:
+   *
+   * - rsn2 > rsn1 (except in the case of a wrap around)
+   * - `rsn2 = rsn1 + 1` if and only if both `rmw_take*()` calls happened consecutively.
+   */
+  int64_t reception_sequence_number;
+  /// Global unique identifier of the publisher that sent the message.
   rmw_gid_t publisher_gid;
```
wjwwood, line 578: We might want to clarify that this gid should be unique even if the publisher is destroyed and re-created. And also verify this is true for DDS.

I'm imagining a corner case where:

- you take a message with psn = 1 pgid = abc
- publisher abc is destroyed
- publisher is created and (re)assigned gid abc
- you take a message and get psn = 1 and pgid = abc again ⚡

rmw/src/types.c
```diff
@@ -19,6 +19,9 @@ RMW_WARN_UNUSED
 rmw_message_info_t
 rmw_get_zero_initialized_message_info(void)
 {
-  rmw_message_info_t zero_initialized_message_info = {0, 0, {NULL, {0}}, false};
+  rmw_message_info_t zero_initialized_message_info = {
+    0, 0, RMW_MESSAGE_INFO_SEQUENCE_NUMBER_UNSUPPORTED,
+    RMW_MESSAGE_INFO_SEQUENCE_NUMBER_UNSUPPORTED, {NULL, {0}}, false
```
wjwwood, line 24: Maybe `0` could just be used here? I mean the values are not valid after using this function. For example, there's no way to tell between timestamp `0` and zero initialized. Just a thought.

rmw/include/rmw/types.h
christophebedard: can we also clarify the meaning of "global"? Either here or in other GID-related functions/structures of the rmw interface. Especially since it's not currently *global* global with Cyclone DDS: https://github.com/ros2/rmw_cyclonedds/issues/377

vmayoral:
Pinging @eboasson in here. This seems a rather promissing feature in `rmw` and It'd be awesome for CycloneDDS to consider adding capabilities to support it fully.

rmw/src/types.c
ivanpauno: I made this change so I don't need to update the rmw implementations that don't fill the sequence number field, they can just keep the default value unmodified.

It's true that now the name of the function is not correct, as the struct is not really "zero" initialized but "default" initialized to something else.

rmw/include/rmw/types.h
ivanpauno: > publisher is created and (re)assigned gid abc

I think that's guaranteed not to happen.
IIRC correctly, the entity id is composed by a prefix that's (pseudo-)unique for the participant (I'm not sure if they use things like MAC address to create an unique-prefix or they just a pseudorandom number), and a second part that depends on the entity type (sub, pub, ...) and an "entity key" which is generally just a counter. The counter maybe can wrap around in some implementations, but it's hard to think that will happen in practice.

ivanpauno: > can we also clarify the meaning of "global"? Either here or in other GID-related functions/structures of the rmw interface. Especially since it's not currently global global with Cyclone DDS: https://github.com/ros2/rmw_cyclonedds/issues/377

I think "global" should mean that you can get the same gid in the process that created the publisher or in the processes comunicating with that publisher.
Though it seems that nothing was depending on that in practice.

christophebedard: > I think "global" should mean that you can get the same gid in the process that created the publisher or in the processes comunicating with that publisher.

That's what I assumed as well, but it's not exactly the case in practice, and it should probably be spelled out here.

> Though it seems that nothing was depending on that in practice.

Yeah, and I would like to be able to rely on that :grin:

rmw/src/types.c
wjwwood: I see why you did this, but I really think it's abusing this function. I'd rather see a new function that is "default" rather than "zero" as you said, or update the rmw's to set unsupported if it is unsupported.

Either way you have to touch the middlewares, and we'll need to touch them anyway to support the function that tests for functionality being available.

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,66 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   * calls to `rmw_take*()`, where `psn1` was obtained in a call that happened before `psn2` and both
+   * sequence numbers are from the same publisher (i.e. also same publisher gid), then:
+   *
+   * - psn2 > psn1 (except in the case of a wrap around)
+   * - `psn2 - psn1 - 1` is the number of messages the publisher sent in the middle of both
+   *   received messages.
+   *   Those might have already been taken by other `rmw_take*()` calls that happened in between or lost.
+   *   `psn2 - psn1 - 1 = 0` if and only if the messages were sent by the publisher consecutively.
+   */
+  int64_t publication_sequence_number;
```
wjwwood, line 558: Should this not be an `uint64_t`? Why would it need to be negative?

rmw/include/rmw/types.h
ivanpauno: > That's what I assumed as well, but it's not exactly the case in practice, and it should probably be spelled out here.

Let's see if the gid can be made truly global in cyclonedds as well, if not we can modify the wording in a new PR later.

rmw/src/types.c
ivanpauno: Fair enough, see https://github.com/ros2/rmw/pull/318/commits/9677d2bfa919b73b419fe4a2aaafec4cce2d1718.

ivanpauno:
This PR and https://github.com/ros2/rmw_fastrtps/pull/587 are ready, I still need to update https://github.com/ros2/rmw_connextdds/pull/74 and create a PR in rmw_cyclonedds.

rmw/src/types.c
```diff
@@ -19,6 +19,6 @@ RMW_WARN_UNUSED
 rmw_message_info_t
 rmw_get_zero_initialized_message_info(void)
 {
-  rmw_message_info_t zero_initialized_message_info = {0, 0, {NULL, {0}}, false};
+  rmw_message_info_t zero_initialized_message_info = {0};
```
wjwwood, line 22: I think this doesn't work for some reason, IIRC. I think it works in C++ but maybe not C? I don't remember exactly, but maybe CI will tell us :)

ivanpauno: We're using that in other places, so it should be fine (e.g. [here](https://github.com/ros2/rmw/blob/bceeda0a2a1d1451e1e5243fba09aaa3811f33cd/rmw/src/network_flow_endpoint.c#L25)).
See https://stackoverflow.com/questions/11152160/initializing-a-struct-to-0.

rmw/include/rmw/types.h
wjwwood: I don't know either, but, unless we can find out why and it's a good reason, I don't think just reusing those functions a good reason to use a signed int when a unsigned makes more sense. So I think we should change these to `uint64_t`.

rmw/src/types.c
wjwwood: If it works then it works, but I'm just letting you know this kind of brace initialization for zero initializing was an issue in the past. I'm familiar with what is supposed to work. If CI passes, then I'm happy.

wjwwood: It's also why these functions exist, rather than everyone just using `= {0}`. I don't remember the details off-hand but it was necessary.

rmw/include/rmw/types.h
ivanpauno: > I don't know either, but, unless we can find out why and it's a good reason, I don't think just reusing those functions a good reason to use a signed int when a unsigned makes more sense. So I think we should change these to uint64_t.

Sounds good!
I picked `int64_t` only for consistency, but it's true that an unsigned int makes a lot more sense.

ivanpauno: See https://github.com/ros2/rmw/pull/318/commits/17804ddb09831b4414c53b935403a685e42803a0 and https://github.com/ros2/rmw/pull/318/commits/6b2b4f963de9947b2a9ab2e64907eedaed7b95ab.

rmw/include/rmw/types.h
wjwwood: This question is the only open one in my mind. If we don't resolve this then we can't really say this:

> This sequence number is set by the publisher and therefore uniquely identifies a message when combined with the publisher GID.

(from above)

wjwwood (commented):
Other than that one open discussion, it lgtm.

rmw/src/types.c
jacobperron: I think on some platforms (maybe macOS?) this was an issue.

rmw/include/rmw/types.h
ivanpauno: See https://github.com/ros2/rmw/pull/318/commits/a72d4ecd451a6843da22e4e5e1ddfbef8bd6574b.

christophebedard: I feel like we should just say it's unique globally, across machines, and fix `rmw_cyclonedds`'s behaviour.

rmw/src/types.c
ivanpauno: I guess the issue was https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119#c22 (or https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80454) and https://bugs.llvm.org/show_bug.cgi?id=21689.
It seems that the issue has been improved in gcc and completely fixed in clang.
I would keep the universal initializer if it works, and use a more verbose initialization if we see a warning in CI.

rmw/include/rmw/types.h
wjwwood: > in that case we can update the documentation

I actually don't think we can. We can't just change the expected behavior of function any more than we can just change the API or ABI, in my opinion. So instead I'd say, we can document the weaker guarantee with a TODO that we'd like it to be the stronger one in the future, and then update it to be just the strong guarantee in I-Turtle.

But I'm fine with documenting the behavior that cyclone has as the greatest common denominator (minimum requirement for this interface) and improve it later.

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,73 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+  uint64_t reception_sequence_number;
+  /// Global unique identifier of the publisher that sent the message.
+  /**
+   * In rmw_cyclonedds, the "gid" currently is not global.
+   * i.e. the id printed by the local process for its publisher and the gid printed by a remote
+   * process may be different.
+   * gid continues being unique within a process though, and a
+   * (publication_sequence_number, publisher_gid) pair should still uniquely identify a message.
+   */
   rmw_gid_t publisher_gid;
```
wjwwood, line 585: I would not mention cyclone directly. Instead I'd document the hard requirements and what we'd like in the future as a todo and then leave it at that.

My suggestion:

```suggestion
  /**
   * The identifier uniquely identifies the publisher for the local context, but
   * it will not necessarily be the same identifier given in other contexts or processes
   * for the same publisher.
   * Therefore the identifier will uniquely identify the publisher within your application
   * but may disagree about the identifier for that publisher when compared to another
   * application.
   * Even with this limitation, when combined with the publisher sequence number it can
   * uniquely identify a message within your local context.
   *
   * \todo In the future we want this to uniquely identify the publisher globally across
   *   contexts, processes, and machines.
   */
  rmw_gid_t publisher_gid;
```

I'm not sure if the limitation is per context or per process, but if it is per process then you could remove all the stuff about contexts, or I can update it again.

wjwwood: Also, I don't know if we should mention that it can also potentially roll over? Not sure that would ever happen in practice, but it would be good to say what happens even if the answer is "we don't specify what happens in that case" (i.e. undefined behavior).

ivanpauno: > I don't know if we should mention that it can also potentially roll over? Not sure that would ever happen in practice

Yes, I think it could roll over (I'm not sure what exactly DDS vendors do), and some rmw implementations could use pseudo-random numbers as gid that could potentially collide (IIRC DDS doesn't do that).

ivanpauno:
```suggestion
  /**
   * The identifier uniquely identifies the publisher for the local context, but
   * it will not necessarily be the same identifier given in other contexts or processes
   * for the same publisher.
   * Therefore the identifier will uniquely identify the publisher within your application
   * but may disagree about the identifier for that publisher when compared to another
   * application.
   * Even with this limitation, when combined with the publisher sequence number it can
   * uniquely identify a message within your local context.
   * Publishers gid generated by rmw implementations might at some point collide, in which
   * case it's not posible to distinguish from which message was the publisher.
   * The details of how GIDs are generated are rmw implementation dependent.
   *
   * \todo In the future we want this to uniquely identify the publisher globally across
   *   contexts, processes, and machines.
   */
  rmw_gid_t publisher_gid;
```

ivanpauno: Fixed in https://github.com/ros2/rmw/pull/318/commits/dcd3ea3ee6e958b6f575e7686c82b8ccfdbf1f07

jacobperron (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16351/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10962/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16717/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16356/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10964/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16719/)

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,82 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   * The identifier uniquely identifies the publisher for the local context, but
+   * it will not necessarily be the same identifier given in other contexts or processes
+   * for the same publisher.
+   * Therefore the identifier will uniquely identify the publisher within your application
+   * but may disagree about the identifier for that publisher when compared to another
+   * application.
+   * Even with this limitation, when combined with the publisher sequence number it can
+   * uniquely identify a message within your local context.
+   * Publishers gid generated by rmw implementations might at some point collide, in which
+   * case it's not posible to distinguish from which message was the publisher.
```
wjwwood, line 588:
```suggestion
   * Publisher GIDs generated by the rmw implementation collide at some point, in which
   * case it is not possible to distinguish which publisher sent the message.
```

rmw/include/rmw/types.h
```diff
@@ -515,11 +515,82 @@ typedef struct RMW_PUBLIC_TYPE rmw_gid_s
...
+   * The details of how GIDs are generated are rmw implementation dependent.
```
wjwwood, line 589: Now that I see it written, I don't think this ok. I mean the user has to be able to assume that a gid is unique right? I think the middleware should probably raise an error or at least define what happens when this occurs.

```suggestion
   * The details of how GIDs are generated are rmw implementation dependent.
   *
   * It is possible the the rmw implementation needs to reuse a publisher GID,
   * due to running out of unique identifiers or some other constraint, in which case
   * the rmw implementation may document what happens in that case, but that
   * behavior is not defined here.
   * However, this should be avoided, if at all possible, by the rmw implementation,
   * and should be unlikely to happen in practice.
```

wjwwood (approved):
Other than a doc update, lgtm with CI.

I offered a revision of the docs again, and it brings to my attention that we need to nail down what we mean by `gid` and "global(ly)", but that shouldn't block this I think.

We could use:

- GID (Global Identifier): what we use now, where global means per process, but could mean in the ROS graph
- GUID (Globally Unique Identifier): means the same as GID except each is unique and should never be reused withing the scope of "global"
- UUID (universally unique identifier): unique in the literal universe, within statistical bounds, which is what the general purpose uuid's are used for CS and usually require as "good" as is possible random numbers

Anyway, we need to figure out what we need/want and be clearer about it, again not blocking this pr.

wjwwood:
@ivanpauno you might want to update the content of the issue description (now that we've iterated a bit) and link to all the related prs, rather than relying on cross-references from github.

ivanpauno:
> GID (Global Identifier): what we use now, where global means per process, but could mean in the ROS graph
> GUID (Globally Unique Identifier): means the same as GID except each is unique and should never be reused withing the scope of "global"
> UUID (universally unique identifier): unique in the literal universe, within statistical bounds, which is what the general purpose uuid's are used for CS and usually require as "good" as is possible random numbers

I think that the idea of the gid was always for it to be global and unique.
And with global, I mean all the processes involved would see the same ids.
AFAIU, UUID/GUID are the same thing (though the term UUID is typically used to refer to a particular standard).

rmw/include/rmw/types.h
ivanpauno: Minimal variant of the proposed wording (only added a `could`):

```suggestion
   * Publisher GIDs generated by the rmw implementation could collide at some point, in which
   * case it is not possible to distinguish which publisher sent the message.
```

ivanpauno: See https://github.com/ros2/rmw/pull/318/commits/52d902f50d5b1b92f0118d7e0681a9798c8e578c

rmw/include/rmw/types.h
ivanpauno: Added in https://github.com/ros2/rmw/pull/318/commits/96b9fd2be174d9c3ebbb17752328dac965fc4654

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16722/)

ivanpauno:
> @ivanpauno you might want to update the content of the issue description (now that we've iterated a bit) and link to all the related prs, rather than relying on cross-references from github.

Done

Issue #304, opened 2021-03-09
Get most compatibility QoS policies given QoS for multiple endpoints


jacobperron:
#### Feature description

In the presence of multiple publishers, with potentially different QoS settings, it is useful to automatically setup a subscription with a QoS that is compatible with the majority (if not all) of the publishers. Presumably, what it means for two QoS profiles to be compatible depends on the RMW implementation. Therefore, it makes sense to offer an API that, given one or more endpoints, returns a QoS profile that is compatible with the most endpoints.

Immediately useful in the following tools:

- ros2topic for choosing QoS based on existing publishers/subscriptions (https://github.com/ros2/ros2cli/issues/594)
- rosbag2 for choosing QoS for recording and playback
- ros1_bridge (for example, https://github.com/ros2/ros1_bridge/pull/304)
- [domain_bridge](http://github.com/ros2/domain_bridge) for bridging multiple publishers on the same topic

#### Implementation considerations

Given the non-symmetric nature of publisher and subscription QoS policies, it may make sense to add two API's: one for getting a publisher QoS and one for getting a subscription QoS.

There already exists an implementation for DDS RMWs in rosbag2, see related issue about moving logic to a common place: https://github.com/ros2/rosbag2/issues/601

I'm proposing we actually move the logic to the RMW layer, since it may vary depending on the RMW implementation.
Perhaps moving this logic to RMW is overkill, as the existing QoS policies are fairly specific to DDS. Alternatively, we could hard-code the compatibility rules in `rcl`.

emersonknapp:
I can see this being an RMW function - with an implementation in rmw_dds_common that dds-based RMWs can all use, to avoid duplication. It does make sense that other implementations may have other rules (such as ignoring all qos so all is compatible)

ivanpauno:
The feature sounds good to me!

> Given the non-symmetric nature of publisher and subscription QoS policies, it may make sense to add two API's: one for getting a publisher QoS and one for getting a subscription QoS.

:+1: you actually need the two functions, only one won't work.

> I'm proposing we actually move the logic to the RMW layer, since it may vary depending on the RMW implementation.

IMO when we added the concept of QoS to ROS 2 we kind of accepted the DDS endpoint matching logic.
i.e.: it feels ok to me to just have a method in `rcl`.

I'm not against a `rmw` function though.

mjeronimo:
[...]

PR #319, opened 2022-04-08
Add functions to get a compatible QoS profiles

jacobperron:
Connects to https://github.com/ros2/rmw/issues/304

Given one or more publisher (subscription) QoS profiles, modify the given subscription (publisher) QoS profile such that it is compatible with most of the publisher (subscription) profiles.
Do this while not sacrificing the level of service when possible.

This API is ultimately intended to be leveraged by ROS tools that typically want to be flexible with respect to the QoS when subscribing or publishing to a topic.

---

Opening for early feedback about the API. There is a reference implementation for one of the functions in https://github.com/ros2/rmw_dds_common/pull/59

My next step is to add `rcl` functions that query the ROS graph to collect QoS info and then use this API to return the most compatible QoS profile, e.g.

```c
rcl_ret_t rcl_get_compatible_subscription_qos_from_topic(
  const rcl_node_t * node,
  const char * topic_name,
  rmw_qos_profile_t * subscription_qos_profile
)
```

I'm leaning towards putting the endpoint query logic in `rcl`, since it can be implemented in an RMW-agnostic way (using the API proposed in this PR). However, we could also consider doing all of that logic in RMW in case the RMW implementation can do something more efficient :shrug:

### DIFF
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -220,7 +220,8 @@ rmw_qos_profile_check_compatible(
  * \return `RMW_RET_INVALID_ARGUMENT` if `subscription_profile` is `nullptr`, or
  * \return `RMW_RET_ERROR` if there is an unexpected error.
  */
-RMW_DDS_COMMON_PUBLIC
+RMW_PUBLIC
+RMW_WARN_UNUSED
 rmw_ret_t
 rmw_qos_profile_get_most_compatible_for_subscription(
   const rmw_qos_profile_t * publisher_profiles,
@@ -267,7 +268,8 @@ rmw_qos_profile_get_most_compatible_for_subscription(
  * \return `RMW_RET_INVALID_ARGUMENT` if `subscription_profile` is `nullptr`, or
  * \return `RMW_RET_ERROR` if there is an unexpected error.
  */
-RMW_DDS_COMMON_PUBLIC
+RMW_PUBLIC
+RMW_WARN_UNUSED
 rmw_ret_t
 rmw_qos_profile_get_most_compatible_for_publisher(
   const rmw_qos_profile_t * subscription_profiles,
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -180,6 +180,101 @@ rmw_qos_profile_check_compatible(
   char * reason,
   size_t reason_size);
 
+/// Get compatible QoS policies for a subscription.
+/**
+ * Given one or more publisher QoS profiles, return a QoS profile for a subscirption
+ * that is compatible with the majority of the publisher profiles while maintaining the highest
+ * level of service possible.
+ *
+ * This function modifies the provided subscription QoS profile in order to make it compatible
+ * with the publisher profiles.
+ * It should only modify those members of the QoS profile that it needs to improve compatibility.
+ * For example, for DDS implementations this function should not modify history depth, which
+ * has no role in determining compatibility.
+ *
+ * It's possible not all input QoS profiles are compatible with the output subscription QoS
+ * profile.
+ * You can optionally provide a boolean array that will by set by this function to indicate
+ * which publisher QoS profiles are compatible with the result.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] publisher_profiles: An array of QoS profiles for publishers.
+ * \param[in] publisher_profiles_length: The number of elements in `publisher_profiles`.
+ * \param[out] subscription_profile: QoS policies that are compatible with the majorty of
+ *   the input publisher profiles.
+ * \param[out] compatible_publisher_profiles: An array of boolean values indicating if the
+ *   corresponding Qos profile at the same index in the publisher profiles array is compatible
+ *   with the resultant subscription QoS profile or not.
+ *   This parameter is optional and may be `nullptr`.
+ *   If provided, it must be the same length as the publisher profiles array.
+ * \return `RMW_RET_OK` if the operation was successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `publisher_profiles` is `nullptr`, or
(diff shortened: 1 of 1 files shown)


jacobperron:
Here's a WIP PR for `rcl`: https://github.com/ros2/rcl/pull/976

rmw/include/rmw/qos_profiles.h
```diff
@@ -180,6 +180,103 @@ rmw_qos_profile_check_compatible(
...
+ *   the input subscription profiles.
+ * \param[out] compatible_subscription_profiles: An array of boolean values indicating if the
+ *   corresponding QoS profile at the same index in the subscription profiles array is compatible
+ *   with the resultant publisher QoS profile or not.
+ *   This parameter is optional and may be `nullptr`.
+ *   If provided, it must be the same length as the subscription profiles array.
+ * \return `RMW_RET_OK` if the operation was successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `publisher_profiles` is `nullptr`, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `publisher_profiles_length` is 0, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `subscription_profile` is `nullptr`, or
```
wjwwood, line 268:
```suggestion
 * \return `RMW_RET_OK` if the operation was successful, or
 * \return `RMW_RET_INVALID_ARGUMENT` if `subscription_profiles` is `nullptr`, or
 * \return `RMW_RET_INVALID_ARGUMENT` if `subscription_profiles_length` is 0, or
 * \return `RMW_RET_INVALID_ARGUMENT` if `publisher_profile` is `nullptr`, or
```

wjwwood (approved):
lgtm

rmw/include/rmw/qos_profiles.h
```diff
@@ -180,6 +180,103 @@ rmw_qos_profile_check_compatible(
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] publisher_profiles: An array of QoS profiles for publishers.
+ * \param[in] publisher_profiles_length: The number of elements in `publisher_profiles`.
+ * \param[out] subscription_profile: QoS policies that are compatible with the majorty of
```
j-rivero, line 210:
```suggestion
 * \param[out] subscription_profile: QoS policies that are compatible with the majority of
```
A side from the typo, could it be that "majority" is somehow ambiguous here? I mean, we probably want to know if the policy is "compatible enough" to be used and that implies that is compatible with all input publisher profiles. Is my understanding correct?

jacobperron: For DDS implementations, it will be compatible with all input publisher profiles, but I'm purposefully using the word "majority" here to leave room for non-DDS implementations that may not be able to match all input publishers (their QoS compatibility rules may be different).

j-rivero (approved):

rmw/include/rmw/qos_profiles.h
```diff
@@ -180,6 +180,103 @@ rmw_qos_profile_check_compatible(
...
+ *   the input publisher profiles.
+ * \param[out] compatible_publisher_profiles: An array of boolean values indicating if the
+ *   corresponding QoS profile at the same index in the publisher profiles array is compatible
+ *   with the resultant subscription QoS profile or not.
+ *   This parameter is optional and may be `nullptr`.
+ *   If provided, it must be the same length as the publisher profiles array.
+ * \return `RMW_RET_OK` if the operation was successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if `publisher_profiles` is `nullptr`, or
```
ivanpauno, line 218: Could we have a `RMW_RET_MOST_COMAPATIBLE_NOT_ALL`?
Or sth like that, to indicate that the returned profile is not compatible with all the provided profiles, without the need of providing a `compatible_publisher_profiles` array.

ivanpauno (approved):
I have one suggestion.
I'm not sure if `compatible_subscription_profiles`/`compatible_publisher_profiles` will be useful in practice, but having them is okay.

rmw/include/rmw/qos_profiles.h
jacobperron: I'm open to that idea. My original thinking was that if there are one or more incompatible profiles, then the user could invoke this function again with just those profiles in an effort to match them. But, as you point out, I'm not sure if this would be used in practice :shrug:

wjwwood: We could do that, but is that better than the array? It seems like the array just provides strictly more information. If the function tells me it matches some but not all, my immediate next question is "which ones"? At least that's my thinking.

wjwwood: I think `RMW_RET_MOST_COMAPATIBLE_NOT_ALL` and the array could be useful (the array is optional btw), just because it is easier for the caller to check, rather than always requesting the array and then iterating over it to know if it's a 100% fit or only a partial fit.

jacobperron: I like the additional return code, I can add it.

I considered the array of bools a compromise; because we could alternatively return an array of structs describing precisely what QoS policies are compatible, incompatible, or unknown for each input QoS profile. This seems like it would add a lot of detail and implementation overhead for something I'm not sure anyone would use.

wjwwood: > I considered the array of bools a compromise; because we could alternatively return an array of structs describing precisely what QoS policies are compatible, incompatible, or unknown for each input QoS profile. This seems like it would add a lot of detail and implementation overhead for something I'm not sure anyone would use.

That's fair, I guess knowing which it is not compatible with doesn't tell you much more, but I guess then there are direct comparison functions, where you could compare the one it didn't match with the one it proposed and get more details on why they don't match that way.

fujitatomoya (approved):
looks good to me.

jacobperron:
See https://github.com/ros2/rmw/pull/320 (and connected PRs) for an alternative to this PR. IMO, it makes the interface a lot nicer for the user for a couple reasons:

- They only have to change the QoS policy value in order to take advantage of the feature; instead of calling extra functions and doing error handling.
- They can choose which policies they was to be adaptive; it's not all-or-nothing like this PR.

jacobperron:
Closing in favor of #320

PR #320, opened 2022-04-15
Add 'best available' QoS policies

### LINKED ISSUE #304: Get most compatibility QoS policies given QoS for multiple endpoints

jacobperron:
Resolves #304 
Alternative to https://github.com/ros2/rmw/pull/319

The best available policy should select the highest level of service for the QoS setting while matching with the majority of endpoints.
For example, in the case of a DDS middleware subscription, this means:

- Prefer reliable reliability if all existing publishers on the same topic are reliable, otherwise use best effort.
- Prefer transient local durability if all existing publishers on the same topic are transient local, otherwise use volatile.
- Prefer manual by topic liveliness if all existing publishers on the same topic are manual by topic, otherwise use automatic.
- Use a deadline that is equal to the largest deadline of existing publishers on the same topic.
- Use a liveliness lease duration that is equal to the largest lease duration of existing publishers on the same topic.

---

Connected PRs:

- https://github.com/ros2/rmw_dds_common/pull/60
- https://github.com/ros2/rmw_fastrtps/pull/598
- https://github.com/ros2/rmw_connextdds/pull/85
- https://github.com/ros2/rmw_cyclonedds/pull/389
- https://github.com/ros2/rclcpp/pull/1920
- https://github.com/ros2/rclpy/pull/928
- https://github.com/ros2/system_tests/pull/501

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16633/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11215/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/17034/)

I've created branches named `jacob/qos_best_available` from PRs which are not using that name so CI will pick up the relevant changes.


jacobperron:
The implementation for DDS middlewares is in https://github.com/ros2/rmw_dds_common/pull/60
I've integrated the feature into rmw_fastrtps in https://github.com/ros2/rmw_fastrtps/pull/598

I still need to:
- [ ] integrate feature into rmw_cyclonedds and rmw_connextdds
- [ ] add some integration tests (e.g. in [system_tests](https://github.com/ros2/system_tests)).

rmw/include/rmw/types.h
```diff
@@ -381,6 +381,9 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_reliability_policy_e
   /// Attempt to deliver samples, but some may be lost if the network is not robust
   RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT,
 
+  /// Will match the majority of endpoints and use a reliable policy if possible
```
wjwwood, line 384: You might want to put some more details here about how this works, specifically that it will only consider the existing endpoints and that it will be able to use reliable if all the opposite endpoints use reliable and that you can get the actual QoS after creation.

Also worth documenting, what happens for publishers (always reliable)?

rmw/include/rmw/types.h
```diff
@@ -413,6 +416,9 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_durability_policy_e
   /// Samples are not persistent
   RMW_QOS_POLICY_DURABILITY_VOLATILE,
 
+  /// Will match the majority of endpoints and use a transient local policy if possible
```
wjwwood, line 419: Worth documenting that volatile will match publishers that are transient local, but that durable messages that were sent before creating the subscription will not be received.

rmw/include/rmw/types.h
```diff
@@ -452,18 +458,27 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_e
...
   /// Liveliness policy has not yet been set
-  RMW_QOS_POLICY_LIVELINESS_UNKNOWN = 4
+  RMW_QOS_POLICY_LIVELINESS_UNKNOWN = 5
 } rmw_qos_liveliness_policy_t;
 
 /// QoS Deadline default.
 #define RMW_QOS_DEADLINE_DEFAULT RMW_DURATION_UNSPECIFIED
+/// Will match the majority of endpoints while maintaining as strict a policy as possible
+/// Value is RMW_DURATION_INFINITE - 1
+#define RMW_QOS_DEADLINE_BEST_AVAILABLE {9223372036LL, 854775806LL}
```
wjwwood, line 472:
```suggestion
/// Will match the majority of endpoints while maintaining as strict a policy as possible
/**
 * Value is RMW_DURATION_INFINITE - 1
 */
#define RMW_QOS_DEADLINE_BEST_AVAILABLE {9223372036LL, 854775806LL}
```

wjwwood, line 472: Can you explain, briefly for my benefit, how this define works? Why not the naive `RMW_DURATION_INFINITE - 1` or something like that?

wjwwood (commented):
lgtm

One comment about the naming, "best" seems natural, but technically there's no reason to consider some of these better than the alternative.

For example, we could say `RMW_QOS_POLICY_RELIABILITY_RELIABLE_IF_POSSIBLE` instead of `RMW_QOS_POLICY_RELIABILITY_BEST_AVAILABLE` and it would be more accurate I think. I'm not sold on that name really, but "best" is definitely ambiguous to me, especially for things like duration and lifespace, etc. Not a blocker, just wanted to bring it up.

rmw/include/rmw/types.h
ivanpauno: IIUC, it could also be `RMW_DURATION_INFINITE + 1`, which is outside of the currently valid range but can still be represented correctly (?). Or just {MAX, MAX} (?)

ivanpauno (approved):
The proposed interface sounds pretty reasonable.

We should really document that this options should be used with care.
Abusing it can create issues of qos profiles changing randomly when launching a bunch of nodes together (because of discovery timing).

jacobperron:
>  "best" seems natural, but technically there's no reason to consider some of these better than the alternative.

True. It could be misleading to users who may think this is the "best" option. While a name like `RMW_QOS_POLICY_RELIABILITY_RELIABLE_IF_POSSIBLE` makes sense for reliability, durability, and liveliness, what would we call the new deadline and lease duration sentinel values? Maybe something like  `RMW_QOS_DEADLINE_MATCH_ENDPOINTS`?

rmw/include/rmw/types.h
jacobperron: We can't naively do `RMW_DURATION_INFINITE - 1` because it evaluates to this expression that can't be compiled: `{9223372036LL, 854775807LL} - 1`.

I don't know if adding one, or using `INT64_MAX` is a good idea either, since [`rmw_time_equal`](https://github.com/ros2/rmw/blob/01d0db60c5bac242fe6f8a6d49494512f1e20c61/rmw/src/time.c#L22-L44) is typically used to compare values. `rmw_time_equal` converts the time to nanoseconds, which is clamped to `INT64_MAX`, but `RMW_DURATION_INFINITE` is exactly `INT64_MAX` nanoseconds. So, the following is true:

```c++
#define RMW_DURATION_MAX = {INT64_MAX, INT64_MAX}
assert(rmw_time_equal(RMW_DURATION_INFINITE, RMW_DURATION_MAX));
```

jacobperron:
I think there is something nice about using the same term for all policy types, e.g. how should we name the functions introduced in https://github.com/ros2/rmw_dds_common/pull/60?

jacobperron:
Maybe instead of "best available" we call it "highest available" (as in highest level of service, while matching the most endpoints)?

rmw/include/rmw/types.h
jacobperron: Updated documentation in latest commits. I didn't explicitly mention that publishers are always reliable, since that seems like a middleware implementation detail. Instead I use more general language that 'reliable' will be used if it matches all discovered endpoints.

rmw/include/rmw/types.h
jacobperron: Done in [9988e82](https://github.com/ros2/rmw/pull/320/commits/9988e8224be2411600de0a7e1fbc08c7410c4b12)[](https://github.com/jacobperron)

rmw/include/rmw/types.h
jacobperron: [f6d2357](https://github.com/ros2/rmw/pull/320/commits/f6d235776de66a3ff1b9388ed5021fa0c1eb2288)[](https://github.com/jacobperron)

jacobperron:
> We should really document that this options should be used with care.

PTAL at the latest changes to the documentation.

rmw/include/rmw/types.h
ivanpauno: > I don't know if adding one, or using INT64_MAX is a good idea either, since [rmw_time_equal](https://github.com/ros2/rmw/blob/01d0db60c5bac242fe6f8a6d49494512f1e20c61/rmw/src/time.c#L22-L44) is typically used to compare values. rmw_time_equal converts the time to nanoseconds, which is clamped to INT64_MAX, but [RMW_DURATION_INFINITE is exactly INT64_MAX nanoseconds](https://github.com/ros2/rmw/blob/01d0db60c5bac242fe6f8a6d49494512f1e20c61/rmw/include/rmw/time.h#L49). So, the following is true:

Yeah, I think that holds for any rmw time with seconds >= `9223372037LL` (and also for seconds == 9223372036LL and nanoseconds >= 854775807LL).
Using an "in range" value as an special marker is a bit confusing IMO.
Maybe we could add an exception to `rmw_time_equal()`, or maybe ignore the feature for `deadline`.
Maybe the second is a better idea, I'm not sure.

ivanpauno (approved):
The only open discussion is https://github.com/ros2/rmw/pull/320#discussion_r851565365, otherwise it looks good to me

rmw/include/rmw/types.h
wjwwood: > Using an "in range" value as an special marker is a bit confusing IMO.

Why? Just because `rmw_time_equal()` isn't aware of it?

ivanpauno: I don't think it will cause any real issue, I just think it's weird to have:

{9223372036LL, 854775806LL} -> means that the middleware will pick a number for you
{9223372036LL, 854775807LL} -> means infinite (i.e. "let middleware pick deadline" + 1 = infinite?)
{9223372036LL, 854775808LL} -> means infinite
{0LL, 9223372036854775807LL} -> also means infinite

And thinking about the purpose of the "publishers deadline QoS", which seems to only be generating QoS compatibility errors, I'm not sure if adding this is a good idea.
Though it seems to be handy for tools like `ros2 topic pub`, so maybe it's not bad ...

wjwwood: Yeah, I mean so long as it is documented, I don't think it's that big of an issue personally. I think it unlikely that someone accidentally reaches one of these values in practice.

rmw/include/rmw/types.h
```diff
@@ -452,18 +487,72 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_liveliness_policy_e
...
+   *
+   * The actual QoS policy can be retrieved after the endpoint is created with
+   * `rmw_get_subscriptions_info_by_topic` or `rmw_get_publishers_info_by_topic`.
+   *
+   * The middleware is not expected to update the policy after creating a subscription or
+   * publisher, even if the chosen policy is incompatible with newly discovered endpoints.
+   * Therefore, this policy should be used with care since non-deterministic behavior
+   * can occur due to races with discovery.
+   */
+  RMW_QOS_POLICY_LIVELINESS_BEST_AVAILABLE = 4,
```
wjwwood, line 504: Should we be concerned about changing the enum value?

Maybe we should consider making this one 5, and leaving the unknown enum as 4?

I mean, it's a breaking change no matter what, and maybe it isn't a problem, just thinking out loud.

wjwwood (approved):
A few comments, but generally lgtm, nothing blocking.

rmw/include/rmw/types.h
```diff
@@ -381,6 +381,22 @@ typedef enum RMW_PUBLIC_TYPE rmw_qos_reliability_policy_e
...
   RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT,
 
+  /// Will match the majority of endpoints and use a reliable policy if possible
+  /**
+   * A policy will be chosen at the time of creating a subscription or publisher.
+   * A reliable policy will by chosen if it matches with all discovered endpoints,
+   * otherwise a best effort policy will be chosen.
+   *
+   * The actual QoS policy can be retrieved after the endpoint is created with
+   * `rmw_get_subscriptions_info_by_topic` or `rmw_get_publishers_info_by_topic`.
```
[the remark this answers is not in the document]
jacobperron: See [069042b](https://github.com/ros2/rmw/pull/320/commits/069042bb6c0bbabfa7908a1fad4ab41fa9b6089c)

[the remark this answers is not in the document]
jacobperron: [069042b](https://github.com/ros2/rmw/pull/320/commits/069042bb6c0bbabfa7908a1fad4ab41fa9b6089c)

rmw/include/rmw/types.h
jacobperron: No longer changing existing enum values: [069042b](https://github.com/ros2/rmw/pull/320/commits/069042bb6c0bbabfa7908a1fad4ab41fa9b6089c)

jacobperron: I did the same for the other enums [95241ad](https://github.com/ros2/rmw/pull/320/commits/95241adf07ddccef32b5ef5862a329c048640341)

wjwwood (approved):
lgtm

After thinking about it, I can't come up with a better name than "best available", we'll just want to document it well.

If the other discussions are resolved, then this is g2g for me.

jacobperron:
I added a note about services: [492c48f](https://github.com/ros2/rmw/pull/320/commits/492c48fdf9a94e4c490eeb5a534eaa76ab59034e)

rmw/include/rmw/types.h
jacobperron: I think this is sufficiently documented; let me know if you think something is lacking.

jacobperron:
CI is passing for all connected PRs :tada: https://github.com/ros2/rmw/pull/320#issue-1205954246
I think the only thing left to resolve is the comment on this ticket: https://github.com/ros2/rmw_connextdds/pull/84#discussion_r859116214

j-rivero:
Run an smoke CI on ci_linux with `--packages-up-to test_communication test_quality_of_service` to check how things are in all the open PRs after latest changes on https://github.com/ros2/rmw_connextdds/pull/85. I've used this[ ros2.repos](https://gist.githubusercontent.com/j-rivero/ae76708b04089543cd9f190966d1203f/raw/08eb37dc1beef1844be7d7b73a8034d47703c770/ros2.repos) file.
CI is passing there :tada:  @jacobperron a final look to that rmw_connextdds PR and we should be in a good moment to run the full CI again I think.

rmw/include/rmw/types.h
wjwwood: I think so too.

j-rivero (approved):

Issue #330, opened 2022-09-13
Connect/Disconnect notification/callback support


fujitatomoya:
#### Feature description

- RMW can notify the event that any subscriptions connect / disconnect on the topic to each publisher.
- Based on this capability, in `rclcpp` and `rclpy`, user application can register user callback that will be called when subscription connects or disconnects on the topic for publisher.

ROS 1 Similar function: https://docs.ros.org/en/noetic/api/roscpp/html/classros_1_1NodeHandle.html#ae4711ef282892176ba145d02f8f45f8d

#### Use case

On visualization node, to suppress unnecessary CPU utilization, enable its visualization process only when RViz is connected.

![image](https://user-images.githubusercontent.com/43395114/189972176-536b8108-6b39-413d-aaca-6e4475e1ca5e.png)

#### Related Information / Reference

- [image_pipeline comment out subscription match for temporary](https://github.com/ros-perception/image_pipeline/blob/ea216eba6d2c611adbd78bb9ef9db25816d1405c/depth_image_proc/src/point_cloud_xyz.cpp#L59-L60)
- [ROS2 publisher callback on subscription match](https://answers.ros.org/question/351827/ros2-publisher-callback-on-subscription-match/)
- [pointcloud_to_laserscan does polling to check on subscription match](https://github.com/ros-perception/pointcloud_to_laserscan/blob/rolling/src/pointcloud_to_laserscan_node.cpp#L108)

#### Implementation considerations

Since all tier-1 rmw implementation is DDS implementation, we can support this feature with `on_publication_matched` of `DataWriterListener Interface`. (we confirmed the code of Fast-DDS.)

- [x] rmw https://github.com/ros2/rmw/pull/331
- [x] rmw_implementation https://github.com/ros2/rmw_implementation/pull/216
- [x] rmw_fastrtps https://github.com/ros2/rmw_fastrtps/pull/645
- [x] rmw_connextdds https://github.com/ros2/rmw_connextdds/pull/101
- [x] rmw_cyclonedds https://github.com/ros2/rmw_cyclonedds/pull/435
- [x] rcl https://github.com/ros2/rcl/pull/1033
- [x] rclcpp https://github.com/ros2/rclcpp/pull/2105
- [x] rclpy https://github.com/ros2/rclpy/pull/1083
- [x] https://github.com/ros2/demos/pull/607
- [x] https://github.com/ros2/ros2_documentation/pull/3406
- [x] https://github.com/ros2/ros2_documentation/pull/3420

[...]

PR #331, opened 2022-09-22
Add connect & disconnect event

Barry-Xu-2018:
Related to #330


fujitatomoya:
@Barry-Xu-2018 thanks for the effort, i will take a look.

CC: @ivanpauno @wjwwood @MiguelCompany @asorbini

Barry-Xu-2018:
In order to make it easier to understand the definition in PR, I modify rmw_fastrtps to explain how to use.  
https://github.com/Barry-Xu-2018/rmw_fastrtps/commit/9923376c18cb19e6f2cb0bddffbb5c93f9db9d2d

rmw/include/rmw/events_statuses/matched.h
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * For subscriber, total cumulative count the concerned subscriber discovered a 'match' with
+   * a publisher.
+   */
+  size_t total_count;
+
+  /**
+   * The change in total_count since the last time the status was read.
+   */
+  size_t total_count_change;
```
fujitatomoya, line 41: I see that this is come from DDS spec, but not sure what these for. does anyone have idea? CC: @MiguelCompany @asorbini

fujitatomoya (commented):
overall, this is what i was expecting to extend based on `rmw_event` interface. this is really straight forward to extend for each Tier-I RMW implementation.

rmw/include/rmw/events_statuses/matched.h
MiguelCompany: The only difference between `total_count` and `current_count` is that `total_count` is not decreased when there is an unmatch. So I don't know if it makes sense to expose it. Perhaps `current_count` is the only meaningful field for RCL.

AndreasR30:
This would be an interesting feature. If this gets merged this additionally needs to be exposed to rcl/rclcpp/rclpy API, right?

fujitatomoya:
@AndreasR30 yes, that is the idea.

rmw/include/rmw/event.h
```diff
@@ -37,11 +37,15 @@ typedef enum rmw_event_type_e
   RMW_EVENT_REQUESTED_DEADLINE_MISSED,
   RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE,
   RMW_EVENT_MESSAGE_LOST,
+  RMW_EVENT_SUBSCRIPTION_CONNECT,
+  RMW_EVENT_SUBSCRIPTION_DISCONNECT,
```
ivanpauno, line 41: I would be consistent and use matched/unmatched here, though I don't mind much

rmw/include/rmw/events_statuses/matched.h
ivanpauno: Based on @MiguelCompany comment, `current_count` seems the most useful.

ivanpauno: We could also keep `current_count_change`, so it's easy to know when the callback was triggered because two endpoints matched

ivanpauno (commented):
LGTM!
But based on the comments we probably don't need to expose `total_count`/`total_count_change`

Barry-Xu-2018:
Thanks for your comments.  
I will remove "total_count/total_count_change".

fujitatomoya (approved):

fujitatomoya:
@Barry-Xu-2018 can you resolve the comments?

and i just add the list https://github.com/ros2/rmw/issues/330#issue-1371797574 which repositories to be updated.

Barry-Xu-2018:
One point needs to be discussed.   

Now use the same status info for subscription matched/unmatched event and publication matched/unmatched event
User can get rmw_matched_status_t via `rmw_take_event`.

```c
typedef struct RMW_PUBLIC_TYPE rmw_matched_status_s
{
  /**
   * For publisher, the number of subscribers currently matched to the concerned publisher.
   *
   * For subscriber, the number of publishers currently matched to the concerned subscriber.
   */
  size_t current_count;

  /**
   * The change in current_count since the last time the status was read.
   */
  size_t current_count_change;
} rmw_matched_status_t;
```
About current_count_change, there are 2 ways to set. 
1. current_count_change = matched_count - unmatched_count (from last read status)
2.  for matched event, current_count_chagne = matched_count from last read status (not care unmatched event)
     for unmatched event, current_count_chagne = unmatched_count from last read status (not care matched event)

Which way is better ? I should hear real requirement.  
If 2 is selected, I think the name of `rmw_matched_status_t` maybe unsuitable.   
@fujitatomoya @ivanpauno

ivanpauno:
@Barry-Xu-2018 I don't understand the question.
aren't we just copying what DDS provides? e.g. [for FastDDS](https://fast-dds.docs.eprosima.com/en/latest/fastdds/api_reference/dds_pim/core/status/matchedstatus.html#_CPPv4N8eprosima7fastdds3dds13MatchedStatus20current_count_changeE)

Barry-Xu-2018:
> aren't we just setting what DDS provides? e.g. [for FastDDS](https://fast-dds.docs.eprosima.com/en/latest/fastdds/api_reference/dds_pim/core/status/matchedstatus.html#_CPPv4N8eprosima7fastdds3dds13MatchedStatus20current_count_changeE)

Yes. 
Here, `current_count_change` means matched_count - unmatched_count from last read status.   
If matched/unmatched appears in pairs, the current_count_change is 0.  
Now we have matched and unmatched event separately.  For example, user calls rmw_take_event() for matched event and current_count_change is 0. But it doesn't mean matched event doesn't occur from last read status.  

So I provide a second way.  It can get matched or unmatched count from last read status (Of cause, I am not sure if this is useful for user.) from [onPublicationMatched](https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_publisher_info.hpp#L69-L81) and [on_subscription_matched](https://github.com/ros2/rmw_fastrtps/blob/ccea7a356900a16a2872670b67952c99a72ab65d/rmw_fastrtps_shared_cpp/include/rmw_fastrtps_shared_cpp/custom_subscriber_info.hpp#L97-L110)

ivanpauno:
Ah okay, I see now.
Maybe we could have:

```
current_count = <DDS current_count>
matched_events_change = total_count_change
unmatched_events_change = total_count_change - current_count_change
```

We could also keep the DDS style, and provide `current_count`/`current_count_change`/`total_count_change`.
Docs should be clear about their meaning though.

fujitatomoya:
notifying underlying RMW Tier I implementation maintainers, @MiguelCompany @asorbini @eboasson 

if you guys have any thoughts, please share here 👍

fujitatomoya:
> If matched/unmatched appears in pairs, the current_count_change is 0.

I thought this is one of the condition, and event itself means there were matched or unmatched event.
I mean `current_count_change` could be 0, if someone comes and someone else goes, but event comes up.

but I believe https://github.com/ros2/rmw/pull/331#issuecomment-1311778905 would be better and clear understanding for user application.

Barry-Xu-2018:
@ivanpauno @fujitatomoya 

Thanks for your comments.  

> ```
> current_count = <DDS current_count>
> matched_events_change = total_count_change
> unmatched_events_change = total_count_change - current_count_change
> ```
> We could also keep the DDS style, and provide current_count/current_count_change/total_count_change.
> Docs should be clear about their meaning though.

Totally agree.  
I will update code.

Barry-Xu-2018:
After updating below repository, I will add test at rmw_implementation to do full test. And then make PR for these repositories at the same time. 
- rmw_fastrtps (Done. Not make a PR)
- rmw_cyclonedds (Ongoing)
- rmw_connextdds (ToDo)

Barry-Xu-2018:
During developing test case (Refer to [test_event.cpp](https://github.com/Barry-Xu-2018/rmw_implementation/blob/topic-matched-unmatched-event-test/test_rmw_implementation/test/test_event.cpp)), I think current design (rmw_matched_status_t) makes user confused since matched & unmatched event all use the same structure.   

Current design refers to DDS definition. But for user, he maybe doesn't know it.  
So we can use more clear definition.
e.g.
For matched event, it returns
```c++
typedef struct RMW_PUBLIC_TYPE rmw_matched_status_s
{
  /**
   * For publisher, the number of subscribers currently matched to the concerned publisher.
   *
   * For subscriber, the number of publishers currently matched to the concerned subscriber.
   */
  int32_t current_matched_count;

  /**
   * Matched count since the last time the status was read.
   */
  int32_t current_matched_count_change;
} rmw_matched_status_t;
```

For unmatched event, it returns
```c++
typedef struct RMW_PUBLIC_TYPE rmw_unmatched_status_s
{
  /**
   * For publisher, the number of subscribers currently matched to the concerned publisher.
   *
   * For subscriber, the number of publishers currently matched to the concerned subscriber.
   */
  int32_t current_matched_count;

  /**
   * Unmatched count since the last time the status was read.
   */
  int32_t current_unmached_count_change;
} rmw_unmatched_status_t;
```

What do you think ? @ivanpauno  @fujitatomoya

Barry-Xu-2018:
Now, below changes has been done
- rmw_fastrtps (https://github.com/Barry-Xu-2018/rmw_fastrtps/tree/topic-matched-unmatched-event-support)
- rmw_cyclonedds (https://github.com/Barry-Xu-2018/rmw_cyclonedds/tree/topic-add-match-unmatch-event-support)
- rmw_connextdds (https://github.com/Barry-Xu-2018/rmw_connextdds/tree/topic-matched-unmatched-event-support)
- rmw_implementation (https://github.com/Barry-Xu-2018/rmw_implementation/tree/topic-matched-unmatched-event-test)

Fastdds and Cyclonedds can pass all tests.  
Connextdds can pass some tests since it doesn't `rmw_event_set_callback`.  

Wait for the opinions for https://github.com/ros2/rmw/pull/331#issuecomment-1321135801.   
If interface isn't changed, I will start to implement rcl, rclcpp and rclpy.

eboasson:
> During developing test case (Refer to [test_event.cpp](https://github.com/Barry-Xu-2018/rmw_implementation/blob/topic-matched-unmatched-event-test/test_rmw_implementation/test/test_event.cpp)), I think current design (rmw_matched_status_t) makes user confused since matched & unmatched event all use the same structure.
> 
> Current design refers to DDS definition. But for user, he maybe doesn't know it. So we can use more clear definition. e.g. For matched event, it returns
> [...]
> What do you think ? @ivanpauno @fujitatomoya

What I suppose I would like to be able to find out easily is: how many new matches showed up, and how many did I lose. DDS doesn't really help with that, and I think this definition makes more sense.

I had a quick look at your code for the Cyclone RMW layer. It looks fine but it did make me wonder whether it really is computing the desired data in all cases. I'd say one can model the total_count and current_count with a count of the cumulative number of matches and of the cumulative number of unmatches:
```
total_count = M+U
current_count = M-U
```
so you can recover those two that the DDS spec doesn't expose directly:
```
M = (total_count+current_count)/2
U = (total_count-current_count)
```
and then you can easily compute the changes locally. I'm just not quite sure that that is what the code does, but that may be because I haven't looked very carefully 😆

Barry-Xu-2018:
@eboasson  
Thanks for sharing your thoughts.  

According to my understanding, you mean rmw_take_event() takes the same event information regardless of the match or mismatch event.

```c++
typedef struct RMW_PUBLIC_TYPE rmw_matched_status_s
{
  int32_t total_count;
  int32_t current_count;
} rmw_matched_status_t;
```  
User can get match/unmatch count according to these 2 values.  
match/unmatched count means count from application launching.  

This will lose 'change' information.  
I mean that it loses match/unmatch change information from last call rmw_take_event() to get match/unmatched.  
DDS provides this information.  
Whether is this information necessary to be exposed to user ?

eboasson:
> @eboasson Thanks for sharing your thoughts.
> 
> According to my understanding, you mean rmw_take_event() takes the same event information regardless of the match or mismatch event.
> 
> ```c++
> typedef struct RMW_PUBLIC_TYPE rmw_matched_status_s
> {
>   int32_t total_count;
>   int32_t current_count;
> } rmw_matched_status_t;
> ```
> 
> User can get match/unmatch count according to these 2 values. match/unmatched count means count from application launching.
> 
> This will lose 'change' information. I mean that it loses match/unmatch change information from last call rmw_take_event() to get match/unmatched. DDS provides this information. Whether is this information necessary to be exposed to user ?

I am sorry I created confusion: I like what you proposed to provide to the user, providing number of new matches separately from number of new unmatches.

What I think is missing is being able to poll the event and detect cases where a match appeared and a match disappeared in between reading the event twice. The "current count" will then not change, and you would expect that accumulating all returned "current_count_change"s would equal the most recent "current count".

So just by monitoring the current count, there are some cases where something interesting happened that you can't detect from looking at the counters. If you also provide a cumulative version of the number of matches (so incremented each time a new match is created) and a cumulative version of the number of unmatches (incremented when a match is removed), then that becomes visible.

The matching part is easy, that you can get straight from DDS (my previous comment was in error, I edited it). For the unmatching part, you have to do something more, but it is available, and so you can provide the data to the user. All I was going on about is that I thought you made a mistake in that calculation.

Barry-Xu-2018:
> So just by monitoring the current count, there are some cases where something interesting happened that you can't detect from looking at the counters. If you also provide a cumulative version of the number of matches (so incremented each time a new match is created) and a cumulative version of the number of unmatches (incremented when a match is removed), then that becomes visible.

Yes. User may want to get the count of match/unmatch change between read status twice.

> The matching part is easy, that you can get straight from DDS (my previous comment was in error, I edited it). For the unmatching part, you have to do something more, but it is available, and so you can provide the data to the user. 

Yes. About unmatch change, we can get by `total_count_change - current_count_change`.  This formula may confuse users since the user doesn't understand detailed information about `total_count_change` and `current_count_change`.  So I want to directly provide `current_unmached_count_change` to user.  

> All I was going on about is that I thought you made a mistake in that calculation.  

Could you tell me where is a mistake?

eboasson:
> Could you tell me where is a mistake?

Only if there actually is a mistake 😜 Of this I'm not sure. What it saw looks fine in principle, I was hunting for my `total_count_change - current_count_change` and that I couldn't find. But whether you need it depends on what guarantee you want to provide:
* If you're content with seeing the *net* change reflected in the event counts, i.e., we went from 3 to 5 matched readers, then I think it is fine
* If in the case you went from 3-to-8-to-5 readers, you want to inform the publisher that there are currently 5, but you gained 8 and lost 3, then I'm either overlooking something or you're not quite guaranteeing that.

I'm not sure how it interacts with listener callbacks, but I think that doesn't affect it. In any case, in DDS there is no guarantee that you will see each event separately (even if, currently, you will in Cyclone).

One other point of concern that I have, probably entirely unnecessary, is that the mechanism is not re-entrant. I don't know if that is a concern for the wider community, it is just that I make an effort to keep Cyclone thread-safe and so I am always checking that sort of thing 🤓

Barry-Xu-2018:
> What it saw looks fine in principle, I was hunting for my total_count_change - current_count_change and that I couldn't find.

Current codes only return `total_count_change` and `current_count_change`. If user want to know unmatched count, he has to calculate it by himself.  

> But whether you need it depends on what guarantee you want to provide:
>
> - If you're content with seeing the net change reflected in the event counts, i.e., we went from 3 to 5 matched readers, then I think it is fine
> - If in the case you went from 3-to-8-to-5 readers, you want to inform the publisher that there are currently 5, but you gained 8 and lost 3, then I'm either overlooking something or you're not quite guaranteeing that.

For unmatched case, 3-to-8-to-5 leads to  (assume status has been read while reader count is 3. `total_count_change` and  `current_count_change` will be set to 0 once reading status.)
```
total_count_change = 5  
count_change = 5  
current_count_change = 2  
```
So, matched count is total_count_change (5) and unmatched count is `total_count_change - current_count_change` (5-3 = 2)
These 3 items are from DDS.  

> In any case, in DDS there is no guarantee that you will see each event separately (even if, currently, you will in Cyclone).

Actually, we don't need to care about each event. Instead, we care about `total_count_change`, `count_change` and `current_count_change` from DDS.  It can show what change (matched or unmatched) occurs from last read status.  

> One other point of concern that I have, probably entirely unnecessary, is that the mechanism is not re-entrant. I don't know if that is a concern for the wider community, it is just that I make an effort to keep Cyclone thread-safe and so I am always checking that sort of thing 

Thanks for your kind reminder. I should check code to keep thread-safe.

Barry-Xu-2018:
https://github.com/ros2/rmw/pull/331#issuecomment-1321135801   
About this comments, I'd like to hear your opinions.   
@ivanpauno @fujitatomoya

fujitatomoya:
@Barry-Xu-2018 sorry i am pretty occupied, but i will try to catch up in this week.

fujitatomoya:
@Barry-Xu-2018 i might be mistaken...

`total_count_change` is required to user application, what is the use case?  this is just an incremental number every time detects the corresponding endpoints, i am not really sure what this is for...

simplified interface suggested https://github.com/ros2/rmw/pull/331#issuecomment-1321135801 makes sense to me. but can we use the same structure for match and unmatch event?

@eboasson appreciate your comments, that helps!

Barry-Xu-2018:
> total_count_change is required to user application, what is the use case? this is just an incremental number every time detects the corresponding endpoints, i am not really sure what this is for...

I think total_count_change is only used for calculating unmatched count. User should don't care this value.

> simplified interface suggested https://github.com/ros2/rmw/pull/331#issuecomment-1321135801 makes sense to me. but can we use the same structure for match and unmatch event?

Yes.  We can use the same structure as the below.

```
struct RMW_PUBLIC_TYPE rmw_matched_unmatched_status_s
{
  /**
   * For publisher, the number of subscribers currently matched to the concerned publisher.
   *
   * For subscriber, the number of publishers currently matched to the concerned subscriber.
   */
  int32_t current_matched_count;

  /**
   * Matched  or unmatched count since the last time the status was read.
   */
  int32_t current_count_change;
};
typedef  struct rmw_matched_unmatched_status_s rmw_matched_status_t;
typedef  struct rmw_matched_unmatched_status_s rmw_unmatched_status_t;
```

fujitatomoya:
> Yes. We can use the same structure as the below.

IMO, this makes sense to me.

Barry-Xu-2018:
@fujitatomoya @ivanpauno   

I update codes to simplify status structure.   
Not use definition used in DDS to this status structure.    
If you accept this change, I will update other repositories.

fujitatomoya (approved):
lgtm

fujitatomoya:
@ivanpauno @wjwwood can you guys take a look?

Barry-Xu-2018:
Discuss this interface at https://github.com/ros2/rmw_fastrtps/pull/645#discussion_r1048473839

rmw/include/rmw/event.h
```diff
@@ -37,11 +37,15 @@ typedef enum rmw_event_type_e
   RMW_EVENT_REQUESTED_DEADLINE_MISSED,
   RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE,
   RMW_EVENT_MESSAGE_LOST,
+  RMW_EVENT_SUBSCRIPTION_MATCHED,
+  RMW_EVENT_SUBSCRIPTION_UNMATCHED,
```
fujitatomoya, line 41: I believe current interface works and easy to understand for user. but those have to be defined for publisher and subscription separately? if we can make it common events, that would be better. e.g) `RMW_EVENT_MATCHED/UNMATCHED` will be generated for both subscription and publisher?

Barry-Xu-2018: We cannot use common event.
For example, user can call [rcl_take_event()](https://github.com/ros2/rcl/blob/6c0cc99804f9e0e2032fca7ac759f933a88af31c/rcl/include/rcl/event.h#L131) to get event info. If only comment event pass to lower layer, we don't know which matched event for publisher or subscriber should be returned.

Barry-Xu-2018:
Do rebase.

methylDragon:
Does this require the changes for the middleware implementations to be merged together?

fujitatomoya:
@methylDragon yes, all dependent PRs are listed on here, https://github.com/ros2/rmw/issues/330

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18076/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12603/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18720/)

Barry-Xu-2018:
@iuhilnehc-ynos 

I fixed errors.   
Please re-run Linux and Linux-aarch64.  For window, I want to run it after Linux and Linux-arrch64 all pass.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18078/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12606/)

Barry-Xu-2018:
@iuhilnehc-ynos 

Now Linux all passed.   
Please help to run CI on Windows.

iuhilnehc-ynos:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18722/)

eboasson (approved):

Barry-Xu-2018:
@iuhilnehc-ynos 

Please help to run CI again like https://github.com/ros2/rmw/pull/331#issuecomment-1436568818

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18163/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12702/)

Barry-Xu-2018:
The failure Linux/Linux-aarch64  is related to 
```
10:16:34 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/demos/demo_nodes_cpp/src/services/introspection_service.cpp:19:[Kfatal erro[Krcl/service_introspection.h: No such file or directory
```
I will rebase related codes.

Barry-Xu-2018:
@iuhilnehc-ynos 

Rebase was done. Please help to run CI again.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18165/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12704/)

Barry-Xu-2018:
> some test failures and uncrustify warnings, could you address those?

I am investigating the cause of failure for test case [test_events__rmw_cyclonedds_cpp](https://ci.ros2.org/job/ci_linux/18165/testReport/junit/projectroot/test/test_events__rmw_cyclonedds_cpp/)

Barry-Xu-2018:
I fixed found problems.  

@iuhilnehc-ynos Please help to run CI again.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18179/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12715/)

iuhilnehc-ynos:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18846/)

Barry-Xu-2018:
The cause for unstable on window is related to this warning
```
10:30:40 C:\ci\ws\src\ros2\rclcpp\rclcpp_lifecycle\src\lifecycle_node.cpp(170,1): warning C4273: 'rclcpp_lifecycle::LifecycleNode::get_fully_qualified_name': inconsistent dll linkage [C:\ci\ws\build\rclcpp_lifecycle\rclcpp_lifecycle.vcxproj]

10:30:40 C:\ci\ws\src\ros2\rclcpp\rclcpp_lifecycle\include\rclcpp_lifecycle/lifecycle_node.hpp(182,3): message : see previous definition of 'get_fully_qualified_name' (compiling source file C:\ci\ws\src\ros2\rclcpp\rclcpp_lifecycle\src\lifecycle_node.cpp) [C:\ci\ws\build\rclcpp_lifecycle\rclcpp_lifecycle.vcxproj]
```
But PRs didn't modify this file.

clalancette:
> The cause for unstable on window is related to this warning

That was fixed in https://github.com/ros2/rclcpp/pull/2117 , and can be ignored in this PR.

rmw/include/rmw/events_statuses/matched.h
```diff
@@ -0,0 +1,62 @@
...
+   * If you get matched status, this count means the number of publishers matched to the
+   * concerned subscriber.
+   * If you get unmatched status, this count means the number of publishers unmatched to
+   * the concerned subscriber.
+   */
+  int32_t current_count_change;
+};
+
+typedef struct rmw_matched_unmatched_status_s rmw_matched_status_t;
+typedef struct rmw_matched_unmatched_status_s rmw_unmatched_status_t;
```
wjwwood, line 56: After a discussion in the middleware working group, there is an important decision to make here about whether or not this should be two events or one (matched and unmatched events vs a single event).

On the one hand, it's perhaps more intuitive for users to see them as separate events, but at the DDS level the information comes at us in a single event, which could be a coalescence of multiple match and unmatch events. It is possible to represent any single DDS event with multiple of the rmw events as proposed here, without missing anything, but there is one corner case which is that you may produce the events in the wrong order.

Consider:

- DDS event with current change = 0, but total change = 4 (previous was 0)
  - you can infer four separate events, two match and two unmatch, and
  - you can split these into two separate rmw events, but you don't know the order, but you must present the rmw events in some order (arbitrarily perhaps), leading the user to believe they happened separately
- Send rmw matched event, with matched count of 2 and current count change of 2
- Send rmw unmatched event, with matched count of 0 and current count change of -2 (or 2 if we use it absolutely)

This might mistakenly imply to the user that we got two matched _and then_ two unmatched, but in reality we may have gotten "match, unmatch, match, unmatch", or something else.

The single DDS event also does not have this information, but because it is a single event it also does not imply there was an order. Which seems more, not sure the right word, maybe "honest"? "correct"?

I see a few things we could do about this:

- use a single event to encapsulate both matched and unmatched at the same time, like DDS
- document this issue and pick a strategy, e.g. always do matched then unmatched
- always return this event as a sequence of events, so multiple can "occur at the same time", kind of similar to having a single event that encapsulates both

Personally, I like the idea of having a structure like DDS (containing both kinds of events) and breaking it into single events further up the stack, so the user can have the whole story with this interface, or get them one at a time (with the caveat about arbitrarily serializing the events), as they choose. But I don't feel strongly about it.

rmw/include/rmw/events_statuses/matched.h
```diff
@@ -0,0 +1,62 @@
...
+   * If you get unmatched status, this count means the number of subscribers unmatched to
+   * the concerned publisher.
+   *
+   * For subscriber
+   * If you get matched status, this count means the number of publishers matched to the
+   * concerned subscriber.
+   * If you get unmatched status, this count means the number of publishers unmatched to
+   * the concerned subscriber.
+   */
+  int32_t current_count_change;
```
wjwwood, line 52: As documented, this should be unsigned, in my opinion.

If we leave it signed, then maybe it should be negative for unmatched? That doesn't actually sound that good to me, but making it unsigned and leaving it documented as is does sound ok to me.

Also use `size_t`.

rmw/include/rmw/events_statuses/matched.h
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+struct RMW_PUBLIC_TYPE rmw_matched_unmatched_status_s
+{
+  /**
+   * For publisher, the number of subscribers currently matched to the concerned publisher.
+   *
+   * For subscriber, the number of publishers currently matched to the concerned subscriber.
+   */
+  int32_t current_matched_count;
```
wjwwood, line 35: I would recommend making this unsigned and using `size_t`.

rmw/include/rmw/events_statuses/matched.h
Barry-Xu-2018: Yes.

rmw/include/rmw/events_statuses/matched.h
Barry-Xu-2018: Yes.

rmw/include/rmw/events_statuses/matched.h
Barry-Xu-2018: @wjwwood thanks for your information.   

So the decision is to use one structure like DDS for matched and unmatched event and use one event.   

About event order, user can depend on rmw_event_set_callback().  If set callback for matched/unmatched event, user can get the nonfiction while the event occurs.   
If only one event, user cannot know which matched or unmatched event happen.  User must use rmw_take_event() to get status structure to further check.   

Matched status structure returned rmw_take_event() means status changes from last calling rmw_take_event. As you said, it cannot know the order of event.    

Whether we keep the same items in matched event structure like DDS event ?  (All information from DDS are all provided  to user.)
```c
struct matched_status_t {
  uint32_t total_count;
  int32_t total_count_change;
  uint32_t current_count;
  int32_t current_count_change;
};
```

MiguelCompany: > After a discussion in the middleware working group, there is an important decision to make here about whether or not this should be two events or one (matched and unmatched events vs a single event).

I missed the working group because I had another meeting that overlapped. In my opinion, the only useful information for an application (even more if we take into account the motivation in #330) is the value of `current_count`. So I think it would be enough with a single `matched_count_change` event with the `current_count` value.

A value of '0' means the topic is "disconnected". A possitive value means the topic is "connected".
I think this matches the expectations mentioned in #330, is easier to implement, and the meaning will always be correct.

fujitatomoya: > DDS event with current change = 0, but total change = 4 (previous was 0)

(not really the point, but just checking my understanding)

I think in this case, there has been 4 increase, and 4 decrease, since current change is 0.

> The single DDS event also does not have this information, but because it is a single event it also does not imply there was an order. 

true. if we have 2 events, we have to create the order which we do not really know.

> So the decision is to use one structure like DDS for matched and unmatched event and use one event.

yes, i believe so. i think what we can do is to create the event tells `difference`.

> About event order, user can depend on rmw_event_set_callback(). If set callback for matched/unmatched event, user can get the nonfiction while the event occurs.

the problem is this order generated in rmw, right? as mentioned before, if there is a single event mixed with matched and unmatched, there should be no order.

> If only one event, user cannot know which matched or unmatched event happen. User must use rmw_take_event() to get status structure to further check.

this is true, callbacks are mapped to event type, so interface does not allow us to determine event type inside the callback.
user callback can know the either events happened, then taking data via rmw_take_event() to figure out which event happened.

> So I think it would be enough with a single matched_count_change event with the current_count value.
> A value of '0' means the topic is "disconnected". A possitive value means the topic is "connected".
> I think this matches the expectations mentioned in #330, is easier to implement, and the meaning will always be correct.

probably we can take this as `is_connected` not `matched` event, i think this proposal almost covers requirement.

but i think we need to consider the following cases?

1. DDS current_count: 0 -> 1 or greater , `connected` event.
2. DDS current_count: 1 -> 2 or greater , no event? cz already `connected`.
3. DDS current_count: 2 or greater -> 1 , no event? cz still `connected`.
4. DDS current_count: 1 or greater -> 0 , `disconnected` event.
5. DDS current_count: X -> X , no event? cz still nothing seems to change?. (connectivity map could be changed.)

to cover https://github.com/ros2/rmw/issues/330#issue-1371797574, obviously we do not need 2,3 and 5 cases.
on the other hand, we can pass all 1-5 cases then application can figure out what happened with this event.

if we take the path with the former, the following pseudo code makes sense?

```c
if (current_count_change == 0) {
  // skip generating event, since there is no change for number of connection.
}
if (current_count_change > 0) {
  if (current_count - current_count_change == 0) {
    // CONNECT event, transition 0 to current_count_change
  } else {
    // skip generating event, since it bas been already connected.
  }
} else {
  if (current_count == 0) {
    // DISCONNECT event, transition X to 0
  } else {
    // skip generating event, since it is being still connected.
  }
}
```

Barry-Xu-2018: > I missed the working group because I had another meeting that overlapped. In my opinion, the only useful information for an application (even more if we take into account the motivation in https://github.com/ros2/rmw/issues/330) is the value of current_count. So I think it would be enough with a single matched_count_change event with the current_count value.
>
> A value of '0' means the topic is "disconnected". A possitive value means the topic is "connected".
I think this matches the expectations mentioned in https://github.com/ros2/rmw/issues/330, is easier to implement, and the meaning will always be correct.

About `matched_count_change`, how to get this value ? Do you means matched_count_change is current_count_change ?   
If yes, this will lose some information. I think `total_count_change` is important.   
User can get how many matched event happened by total_count_change.  Similarly, the number of unmatched event is gotten by `total_count_change - current_count`.

Barry-Xu-2018: > > The single DDS event also does not have this information, but because it is a single event it also does not imply there was an order.
> 
> true. if we have 2 events, we have to create the order which we do not really know.

rmw_event_set_callback() can set callback to DDS layer. If we use 2 events, I think we can make sure the sequence of calling callback is the order of matched/unmatched event.

Barry-Xu-2018:
We should decide what should be in matched_status_t.   

Now I had updated code based on the below contents for one matched event. If any change, I can easy to change it based on this updated code.  
```
struct matched_status_s {
  uint32_t total_count;
  int32_t total_count_change;
  uint32_t current_count;
  int32_t current_count_change;
};
```
- rmw
    https://github.com/Barry-Xu-2018/rmw/tree/topic-add-matched-event-new
- rmw_fastrtps
    https://github.com/Barry-Xu-2018/rmw_fastrtps/tree/topic-add-matched-event-support-new
- rmw_cyclonedds
    https://github.com/Barry-Xu-2018/rmw_cyclonedds/tree/topic-add-match-event-support-new
- rmw_connextdds
    https://github.com/Barry-Xu-2018/rmw_connextdds/tree/topic-matched-event-support-new
- rmw_implementation
    https://github.com/Barry-Xu-2018/rmw_implementation/tree/topic-add-matched-event-support-new
- rcl
    https://github.com/Barry-Xu-2018/rcl/tree/topic-matched-event-support-new
- rclcpp
    https://github.com/Barry-Xu-2018/rclcpp/tree/topic-matched-event-support-new
- rclpy
    https://github.com/Barry-Xu-2018/rclpy/tree/topic-matched-event-support-new

fujitatomoya:
as far as i can see, we got 3 options (all single event.).

1. Expose the same structure comes from DDS, including total count. (https://github.com/ros2/rmw/pull/331#issuecomment-1465650625)
2. without `total_count` from option 1, i am not really thinking of the use case for `total_count`.
3. only `connect` or `disconnect` event, described in https://github.com/ros2/rmw/pull/331#discussion_r1132997635

IMO, option 2 looks good to me.

@Barry-Xu-2018 @iuhilnehc-ynos @clalancette @wjwwood @MiguelCompany @asorbini @eboasson what do you think?

Barry-Xu-2018:
> IMO, option 2 looks good to me.

Yeah, I also incline this option. For `total_count`, maybe some analysis depends on this. Such as, get the number of connections within a specified time. Of course, we can use other way to get this.

iuhilnehc-ynos:
I vote for option 1 :smiley:  

e.g. a case using `total_count` for a publisher: If the subscribe count it serviced reached a milestone value, please do something.

fujitatomoya:
> e.g. a case using total_count for a publisher: If the subscribe count it serviced reached a milestone value, please do something.

yeah, with `total_count` user can also see how many connected and disconnected since last time against `currnet_count`. that would be information for user application too.
what i was thinking between option 1 and 2, is option 1 could be complicated for user, but we can add some examples and add well-explained doc for this case. since number of events for the user application are the same for option 1 and 2, probably option 1 could be appropriate choice. i am okay to go with option 1.

on the other hand, option 3 can reduce the number of events for user application significantly, user application does not need to be notified with the event for `already connected`(new endpoint is online, but there has been already someone) and `still connected`(some endpoints are gone offline but still some connected) cases. actually this is enough to support this feature.

any thoughts?

Barry-Xu-2018:
> on the other hand, option 3 can reduce the number of events for user application significantly, user application does not need to be notified with the event for already connected(new endpoint is online, but there has been already someone) and still connected(some endpoints are gone offline but still some connected) cases. actually this is enough to support this feature.  

Yeah. First connection event and last disconnection event is real use case from user.

Barry-Xu-2018:
Do rebase

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18248/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12781/)

Barry-Xu-2018:
@iuhilnehc-ynos 

I modify codes in rmw_fastrtps and rcl.  
Please help run Linux & Linux-aarch64 again.

iuhilnehc-ynos:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18248/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12782/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18923/)

Jenkins Job for Linux is not run [before](https://github.com/ros2/rmw/pull/331#issuecomment-1473125693), we can use the old job id which can fetch the latest commit id for all PRs.

Barry-Xu-2018:
Failed cases on Windows are

[projectroot.test_n_nodes__rmw_cyclonedds_cpp](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/(root)/projectroot/test_n_nodes__rmw_cyclonedds_cpp/)

[rcl.TestLogRosoutFixtureGeneral__rmw_connextdds.test_multi_add_remove_sublogger_message](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/rcl/TestLogRosoutFixtureGeneral__rmw_connextdds/test_multi_add_remove_sublogger_message/)

[rcl.TestLoggingRosoutWithDifferentSettings/TestLoggingRosoutFixture__rmw_connextdds.test_logging_rosout/test_enable_explicit_global_rosout_enable_node_option (test_enable_explicit_global_rosout_enable_node_option)](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/rcl/TestLoggingRosoutWithDifferentSettings_TestLoggingRosoutFixture__rmw_connextdds/test_logging_rosout_test_enable_explicit_global_rosout_enable_node_option__test_enable_explicit_global_rosout_enable_node_option______________________/)

[test_rclcpp.TestNNodes.test_10_nodes](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/test_rclcpp/TestNNodes/test_10_nodes/)

[test_rclcpp.TestNNodesAfterShutdown.test_10_nodes](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/test_rclcpp/TestNNodesAfterShutdown/test_10_nodes/)

[projectroot.test.test_logging_rosout__rmw_connextdds](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/projectroot/test/test_logging_rosout__rmw_connextdds/)

[rcl.TestLoggingRosoutWithDifferentSettings/TestLoggingRosoutFixture__rmw_connextdds.test_logging_rosout/test_enable_implicit_global_rosout_enable_node_option (test_enable_implicit_global_rosout_enable_node_option)](https://ci.ros2.org/job/ci_windows/18923/testReport/junit/rcl/TestLoggingRosoutWithDifferentSettings_TestLoggingRosoutFixture__rmw_connextdds/test_logging_rosout_test_enable_implicit_global_rosout_enable_node_option__test_enable_implicit_global_rosout_enable_node_option______________________/)

These test cases aren't related to PRs' change.

fujitatomoya:
@Barry-Xu-2018 thanks for working on this.

@clalancette the following PRs need to be merged together.

-  rmw https://github.com/ros2/rmw/pull/331
-  rmw_implementation https://github.com/ros2/rmw_implementation/pull/216
-  rmw_fastrtps https://github.com/ros2/rmw_fastrtps/pull/645
-  rmw_connextdds https://github.com/ros2/rmw_connextdds/pull/101
-  rmw_cyclonedds https://github.com/ros2/rmw_cyclonedds/pull/435
-  rcl https://github.com/ros2/rcl/pull/1033
-  rclcpp https://github.com/ros2/rclcpp/pull/2105
-  rclpy https://github.com/ros2/rclpy/pull/1083

But i do not have merge permission for https://github.com/ros2/rmw_cyclonedds, could you help us to merge those?

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18259/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12796/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18936/)

PR #334, opened 2022-10-08
Add rmw count clients, services

leeminju531:
this PR is for service counts using it's name like topic.
Related to https://github.com/ros2/ros2cli/pull/771

Signed-off-by: leeminju531 <dlalswn531@naver.com>

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -3020,7 +3020,7 @@ rmw_count_clients(
  *
  * \param[in] node Handle to node to use to query the ROS graph.
  * \param[in] service_name Fully qualified ROS topic name.
- * \param[out] count Number of publishers matching the given topic name.
+ * \param[out] count Number of services matching the given topic name.
  * \return `RMW_RET_OK` if the query was successful, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `node` is NULL, or
  * \return `RMW_RET_INVALID_ARGUMENT` if `service_name` is NULL, or
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -2937,6 +2937,108 @@ rmw_count_subscribers(
   const char * topic_name,
   size_t * count);
 
+/// Count the number of known clients matching a service name.
+/**
+ * This function returns the numbers of clients of a given service in the ROS graph,
+ * as discovered so far by the given node.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | Maybe [1]
+ * Lock-Free          | Maybe [1]
+ * <i>[1] implementation defined, check the implementation documentation</i>
+ *
+ * \par Runtime behavior
+ *   To query the ROS graph is a synchronous operation.
+ *   It is also non-blocking, but it is not guaranteed to be lock-free.
+ *   Generally speaking, implementations may synchronize access to internal resources using
+ *   locks but are not allowed to wait for events with no guaranteed time bound (barring
+ *   the effects of starvation due to OS scheduling).
+ *
+ * \par Thread-safety
+ *   Nodes are thread-safe objects, and so are all operations on them except for finalization.
+ *   Therefore, it is safe to query the ROS graph using the same node concurrently.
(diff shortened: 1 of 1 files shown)


leeminju531:
F.Y.I
- rmw_cyclonedds (https://github.com/ros2/rmw_cyclonedds/pull/427)
- rmw_fastrtps (https://github.com/ros2/rmw_fastrtps/pull/641)
- rmw_connextdds (https://github.com/ros2/rmw_connextdds/pull/93)

There is yet no reviewers in rmw_connextdds (https://github.com/ros2/rmw_connextdds/pull/93)
I would be grateful if someone could confirm.

methylDragon (approved):

methylDragon:
Wait out for CI

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17732/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12274/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18376/)

rmw/include/rmw/rmw.h
```diff
@@ -2937,6 +2937,108 @@ rmw_count_subscribers(
...
+rmw_ret_t
+rmw_count_clients(
+  const rmw_node_t * node,
+  const char * service_name,
+  size_t * count);
+
+/// Count the number of known servers matching a service name.
+/**
+ * This function returns the numbers of servers of a given service in the ROS graph,
+ * as discovered so far by the given node.
```
fujitatomoya, line 2994: Do we actually support multiple servers on single service name? i believe that we can create the services on the same name, but how it works is really dependent on implementation. If this is supported, i would expect that it can also support load balancing.

IMO, multiple service servers on single topic is not officially supported. in that case, I am not sure what this is for.

rmw/include/rmw/rmw.h
```diff
@@ -2937,6 +2937,108 @@ rmw_count_subscribers(
...
+ *   Therefore, it is safe to query the ROS graph using the same node concurrently.
+ *   However, access to primitive data-type arguments is not synchronized.
+ *   It is not safe to read or write `service_name` or `count` while rmw_count_services()
+ *   uses them.
+ *
+ * \pre Given `node` must be a valid node handle, as returned by rmw_create_node().
+ *
+ * \param[in] node Handle to node to use to query the ROS graph.
+ * \param[in] service_name Fully qualified ROS topic name.
+ * \param[out] count Number of publishers matching the given topic name.
```
fujitatomoya, line 3023:
```suggestion
 * \param[out] count Number of services matching the given topic name.
```

rmw/include/rmw/rmw.h
fujitatomoya: @leeminju531 @methylDragon what do you think?

leeminju531: The ROS2 design says
>DDS currently does not have a ratified or implemented standard for request-response style RPC which could be used to implement the concept of services in ROS. There is currently an RPC specification being considered for ratification in the OMG DDS working group, and several of the DDS vendors have a draft implementation of the RPC API.

I think It's seem like none of the specification in this situation. I don't know if some supports like load balancing will be added in the future, As long as most vendors allow multiple servers on the same name, it seems like a necessary function even now.

ivanpauno: > Do we actually support multiple servers on single service name? i believe that we can create the services on the same name, but how it works is really dependent on implementation. If this is supported, i would expect that it can also support load balancing.

It's not supported, but you can still create more than one.
I think that the default behavior is that all servers reply, no load balancing.
I'm not completely sure though, and I'm not sure if something breaks in that case.

Being able to get the number of service servers still seems okay.

fujitatomoya: > It's not supported, but you can still create more than one.
> I think that the default behavior is that all servers reply, no load balancing.

it seems that it depends on implementation, i just checked with Fast-DDS, it sometimes both servers reply, sometimes only one server replies. for doing this, i guess we need to consider more details, like how we can match the service type, how we can choose the server, how to load balance and so on. this could be new feature.

> Being able to get the number of service servers still seems okay.

yeah, at least we can create multiple servers on the same service path.

methylDragon:
Posting this in here for reference too: https://github.com/ros2/rmw_implementation/pull/208

PR #329, opened 2022-08-23
make writer_guid uint8_t[] instead of int8_t[] for consistency with rmw_gid_t

ihasdapie:
This PR address another inconsistency I noticed with with gids in rmw, namely that `rmw_gid_t` is an `uint8_t` array and `rmw_request_id_t.writer_guid` is a `int8_t` array. Changing this would silence some `rcl` compilation warnings introduced by the service introspection feature (https://github.com/ros2/ros2/issues/1285). 

Reference:

https://github.com/ros2/rmw/blob/2259c3f6f3de8c479a9d5c74cdcd03d0010413cd/rmw/include/rmw/types.h#L356-L363

https://github.com/ros2/rmw/blob/2259c3f6f3de8c479a9d5c74cdcd03d0010413cd/rmw/include/rmw/types.h#L620-L627

I'm going to keep this PR as draft until I review the rmw implementations to check if any notable assumptions are made about the signedness of `writer_guid`, but I don't expect it to be an issue.

Signed-off-by: Brian Chen <brian.chen@openrobotics.org>

PR #335, opened 2022-10-13
Update documentation about expectations for GIDs

jacobperron:
Replaces https://github.com/ros2/rmw/pull/329
Follow-up to https://github.com/ros2/rmw/pull/328

---

This PR contains two changes:

- Use gid struct for request id
Everywhere else in rmw, we use the type rmw_gid_t for global identifiers.
For example, rmw_get_gid_for_publisher and rmw_get_gid_for_client.
Furthermore, other global identifiers have a maximum size of 24 bytes, not 16 bytes.
Changing the type and name of the client id in the rmw_request_id_t struct makes it consistent with other parts of the rmw API.

- Update documented expectations for GIDs
  * Clarify that GIDs are globally unique within a ROS domain and consistent between processes/hosts.
  * Specify that any unused bytes in the GID should be set to zero (e.g. in cases where not all 24 bytes are used).
   This allows application code to directly compare GIDs without having to ask the middleware.

The latter change is especially useful for the proposed [service introspection](https://github.com/ros-infrastructure/rep/pull/360) feature. E.g. if tooling wants to compare 24 byte GIDs, we want to ensure that any unused bytes are not set to garbage values and that they are consistent between processes and hosts.

If these changes are acceptable, I'll proceed in making the necessary changes to rmw implementations to:

- Update any references to "writer_guid" to use the new "client_gid.data" member.
- Ensure that they set unused bytes in the GID to zeros

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -3039,7 +3039,7 @@ rmw_count_services(
   const char * service_name,
   size_t * count);
 
-/// Get the unique identifier (gid) of a publisher.
+/// Get the globally unique identifier (GID) of a publisher.
 /**
  * <hr>
  * Attribute          | Adherence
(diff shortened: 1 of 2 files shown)


jacobperron:
For context, for DDS implementation the [DDS-RTPS spec](https://www.omg.org/spec/DDSI-RTPS/2.5/PDF) describes a GUID as being 16 octets and globally unique within a domain.

From section _8.2.1.2_:
> Type used to hold globally-unique RTPS-entity identifiers. These are identifiers used
to uniquely refer to each RTPS Entity in the system.
Must be possible to represent using 16 octets.
The following values are reserved by the protocol: GUID_UNKNOWN

Specifically, for a DDS-Entity, from _8.2.4 The RTPS Entity_:
> Globally and uniquely identifies the
RTPS Entity within the DDS
domain

I think these align well with documentation changes proposed in this PR. But, please let me know if you think these expectations may be problematic.

christophebedard:
> But, please let me know if you think these expectations may be problematic.

This means that https://github.com/ros2/rmw_cyclonedds/issues/377 will have to be fixed (which I am totally in favour of, of course).

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18181/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18842/)

clalancette:
One of the failing tests on Windows is a known flake.  The other one is new to me, but I honestly don't see how this PR could possibly have caused it, given that these are documentation updates only.  So I'm going to go ahead and merge this anyway.

PR #338, opened 2022-12-14
[rmw] Improve handling of dynamic discovery

gbiggs:
This PR adds to `rmw` functionality necessary to support the [improved handling of dynamic discovery](https://github.com/ros2/ros2/issues/1359).

It adds handling of the new discovery-related configuration parameters to `rmw` so they can be used by `rcl` to control the behaviour of RMW implementations.


rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,78 @@
...
+  /// Forces multicast discovery off
+  RMW_AUTOMATIC_DISCOVERY_RANGE_OFF = 1,
+  /// Allows multicast but only on the localhost
+  RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST = 2,
+  /// Allows multicast on the reachable network
+  RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET = 3,
+} rmw_automatic_discovery_range_t;
+
+/// Maximum number of peers that can be manually specified
+#define RMW_DISCOVERY_PARAMS_MAX_PEERS 32
```
ivanpauno, line 40: is there a limit in DDS vendors?

ivanpauno (commented):
LGTM!

I have some minor comments/suggestions.

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,78 @@
...
+{
+  /// How far to allow multicast to be used
+  rmw_automatic_discovery_range_t automatic_discovery_range;
+
+  /// The list of manually-specified peers to perform static discovery with
+  /**
+   * Each peer is specified as a hostname or an IP address (IPv4 and IPv6 are both acceptable), or
+   * a subnet, e.g. 192.168.0.0/24.
+   */
+  char static_peers[RMW_DISCOVERY_PARAMS_MAX_PEERS][RMW_DISCOVERY_PARAMS_PEER_MAX_LENGTH];
```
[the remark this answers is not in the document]
methylDragon: If there's no reason that the `RMW_DISCOVERY_PARAMS_MAX_PEERS` should be capped at 32, we should either have that "32" be a passed in cmake arg, an environment variable, or otherwise go with a dynamic array:

```
peer_address_t * static_peers;
```

And then take in a size argument in the `rmw_get_zero_initialized_discovery_params` below.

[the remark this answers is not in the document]
ivanpauno: > There is the open question about what to do for people who want to pre-allocate everything, but this isn't the only place they'll have to deal with that.

IIUC, this would only be allocated once from the environment variables when you create the node, so it shouldn't be an issue.

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,78 @@
...
+typedef enum RMW_PUBLIC_TYPE rmw_automatic_discovery_range_e
+{
+  /// Uses ROS_LOCALHOST_ONLY environment variable.
+  RMW_AUTOMATIC_DISCOVERY_RANGE_DEFAULT = 0,
+  /// Forces multicast discovery off
+  RMW_AUTOMATIC_DISCOVERY_RANGE_OFF = 1,
+  /// Allows multicast but only on the localhost
+  RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST = 2,
+  /// Allows multicast on the reachable network
+  RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET = 3,
```
fujitatomoya, line 36: What if system has multiple network interface? that is also rmw implementation dependent as it is now?

rmw/src/discovery_params.c
```diff
@@ -0,0 +1,55 @@
...
+
+  if (left->static_peers_count != right->static_peers_count) {
+    return false;
+  }
+
+  for (size_t ii = 0; ii < left->static_peers_count; ++ii) {
+    if (strncmp(
+        left->static_peers[ii],
+        right->static_peers[ii],
+        RMW_DISCOVERY_PARAMS_PEER_MAX_LENGTH) != 0)
```
fujitatomoya, line 48: What about the one is with hostname which can be DNSed to IP, and the other with the same IP address? that is not supported here?

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-02-16/29927/1

sloretz:
@arjo129 can this be closed now that #349 merged?

rmw/include/rmw/discovery_params.h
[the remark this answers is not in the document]
wjwwood: `static_peers` is dynamic now.

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,139 @@
...
+#include "rmw/macros.h"
+#include "rmw/ret_types.h"
+#include "rmw/visibility_control.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// Used to enable or disable the use of multicast for node discovery
```
wjwwood, line 29: I think we should reword the docstrings to avoid mentioning multicast. This should affect discovery range even if multicast is not used by the underlying middleware, in my opinion.

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,139 @@
...
+/// Maximum length of a peer hostname or IP address
+#define RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH 256
+
+/// Struct to typedef some of the peer addresses
+typedef struct rmw_peer_address_s
+{
+  char peer_address[RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH];
+} rmw_peer_address_t;
+
+/// Used to specify the parameters that control how discovery is performed
```
wjwwood, line 51:
```suggestion
/// Used to specify the options that control how discovery is performed
```

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,139 @@
...
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_discovery_options_equal(
+  const rmw_discovery_options_t * const left,
+  const rmw_discovery_options_t * const right,
+  bool * result);
+
+/// Perform a deep copy of the discovery parameters from src into dst using the
```
wjwwood, line 100:
```suggestion
/// Perform a deep copy of the discovery options from src into dst using the
```

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * The dst will be left with an owned copy of the static peers
+ * array whose string values match the src. If successful, src and dst will
+ * evaluate as equal using rmw_discovery_options_equal.
+ * \param[in] src discovery option parameters to be copied.
+ * \param[in] allocator to use.
+ * \param[out] dst Destination parameters to use.
+ * \return RMW_RET_OK if success.
+ * \return RMW_RET_INVALID_ARGUMENT if either the src, allocator or dst is null.
+ * \return RMW_RET_BAD_ALLOC if allocation fails.
```
wjwwood, line 111:
```suggestion
 * The dst will be left with an owned copy of the static peers
 * array whose string values match the src.
 * If successful, src and dst will evaluate as equal using
 * rmw_discovery_options_equal.
 *
 * \param[in] src discovery option parameters to be copied.
 * \param[in] allocator to use.
 * \param[out] dst Destination parameters to use.
 * \return RMW_RET_OK if success.
 * \return RMW_RET_INVALID_ARGUMENT if either the src, allocator or dst is null.
 * \return RMW_RET_BAD_ALLOC if allocation fails.
```

wjwwood (commented):
API and functions that are implemented here lgtm, but we should have some tests for those new functions that live here. I think @sloretz and I can help with that tomorrow.

wjwwood:
I also think we need to address @fujitatomoya's comments, as I think they're still relevant even after the changes.

rmw/include/rmw/discovery_params.h
arjo129: Yes. That is out of scope of this PR.

arjo129:
I don't have push access to this repo. Would you like me to open yet another PR?

clalancette:
> I don't have push access to this repo. Would you like me to open yet another PR?

@arjo129 I invited you to have push access.  Once you accept, you should be able to continue here.

rmw/include/rmw/discovery_options.h
sloretz: Reworded to avoid mentioning multicast in 9e72cec

rmw/include/rmw/discovery_options.h
sloretz: `parameters` -> `options` in 91beb13

rmw/include/rmw/discovery_options.h
sloretz: `parameters` -> `options` in 91beb13

rmw/include/rmw/discovery_options.h
sloretz: one sentence per line in 5fc6ce9

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,78 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// Used to enable or disable the use of multicast for node discovery
+typedef enum RMW_PUBLIC_TYPE rmw_automatic_discovery_range_e
+{
+  /// Uses ROS_LOCALHOST_ONLY environment variable.
+  RMW_AUTOMATIC_DISCOVERY_RANGE_DEFAULT = 0,
```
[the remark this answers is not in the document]
sloretz: The QoS enums have a [*_SYSTEM_DEFAULT value](https://github.com/ros2/rmw/blob/866bc299f5278c91c107353a907d5e971d2e8e24/rmw/include/rmw/qos_profiles.h#L33). I forget what this was for. Was it to allow setting QoS via a middleware specific configuration?

If so,  `RMW_AUTOMATIC_DISCOVERY_RANGE_SYSTEM_DEFAULT` might be useful to say "don't mess with discovery settings, they're already configured with middleware specific APIs".

sloretz:
@wjwwood FYI 2cf62e35a3536bfa5bed74f54ca829511010c657 is a sizeable change. It:

* adds `rmw_discovery_options_init()` to allocation options with a set number of static peers
* adds an allocator member to the `rmw_discovery_options_t` object
* uses a pointer to an allocator as an argument [because I saw that used elsewhere](https://github.com/ros2/rmw/blob/866bc299f5278c91c107353a907d5e971d2e8e24/rmw/include/rmw/get_topic_names_and_types.h#L95)
* errors if `copy` is called with same object for src and dst
* adds a whole bunch of tests

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,169 @@
...
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_discovery_options_copy(
+  const rmw_discovery_options_t * src,
+  const rcutils_allocator_t * allocator,
```
wjwwood, line 149: We're inconsistent on this, but I think this should be non-const. This doesn't actually do anything, because it's like saying `const int ...`, since the `const` here only says we're not going to change the value of the pointer, not that the thing being pointed to is `const`.

wjwwood: Also, the allocator's state (conceptually) is changing when this is called. So I think it's also more correct to say it should be non-const.

rmw/src/discovery_options.c
```diff
@@ -0,0 +1,165 @@
...
+  const rmw_discovery_options_t * src,
+  const rcutils_allocator_t * allocator,
+  rmw_discovery_options_t * dst)
+{
+  RMW_CHECK_ARGUMENT_FOR_NULL(src, RMW_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR(allocator, return RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(dst, RMW_RET_INVALID_ARGUMENT);
+  if (src == dst) {
+    RMW_SET_ERROR_MSG("src and dst must be different");
+    return RMW_RET_INVALID_ARGUMENT;
```
wjwwood, line 128: Is this case documented in the docstring?

wjwwood (changes requested):
one small necessary (imo) change, otherwise lgtm

rmw/include/rmw/discovery_options.h
sloretz: That makes sense. Changed back to non-const in 8258369b13058e7dc59993eec655cfa76b9c1d44

rmw/src/discovery_options.c
sloretz: Documented this case in 6086e681e33637580f13cb37589de187bc5e8d1a

wjwwood (approved):

rmw/include/rmw/discovery_options.h
mxgrey: > This doesn't actually do anything

I apologize for the bikeshedding I'm about to do, but I don't think it's accurate to say it's the same as `const int ...`. The `const` in this case says we're not going to change **the values of any fields in the struct that you are giving us a reference to**. E.g.

```
typedef struct MyData_s {
  int my_num;
} MyData_t;

void can_change_field_value(MyData_t * data) {
  data->my_num += 1;
}

void cannot_change_field_value(const MyData_t * data) {
  data->my_num += 1; // fails to compile
}
```

> says we're not going to change the value of the pointer

The syntax for that would be `rcutils_allocator_t * const allocator`. That indeed would give an equivalent function signature to `rcutils_allocator_t * allocator`.

> allocator's state (conceptually) is changing when this is called

I already shared some thoughts on logical vs physical const and the principle of least mutability [back in this comment](https://github.com/ros2/rmw/pull/349#discussion_r1130398836) so I won't rehash that. It's mostly inconsequential here because I suppose no one downstream is going to care much about having an immutable allocator, but more broadly I think C/C++'s relationship with mutability is a perilous one, being mutable by default, so developers have a responsibility to make generous use of `const` wherever applicable. This discussion isn't worth holding up this PR, but I think it's an issue worth being mindful of more generally.

rmw/include/rmw/discovery_params.h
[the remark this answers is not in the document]
wjwwood: That's right @sloretz.

I think @mxgrey is also right, it's weird if we're never sending the value down to rmw, so we should either not have it as an rmw option or we should handle it in the rmw layer. It might make sense to have ROS/RCL version that has system default but not have one in RMW if that one doesn't make sense. I know we only have RMW enums atm, however.

However, it might just be that we shouldn't be automatically be converting the DEFAULT into LOCALHOST as we are now.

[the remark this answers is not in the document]
sloretz: > However, it might just be that we shouldn't be automatically be converting the `DEFAULT` into `LOCALHOST` as we are now.

If we are just converting `DEFAULT` to `LOCALHOST` ( I see that happening in the rmw_cyclonedds_cpp PR) then I agree it doesn't seem like an option at the RMW layer is necessary.

If someone wants to use an XML config, and doesn't want the `rmw_*` layer to mess with discovery settings, and the default discovery range is localhost,  what option should they provide for the range? `SUBNET` since [the default behavior seems to be to do nothing, at least for cyclonedds](https://github.com/ros2/rmw_cyclonedds/pull/429/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R1075-R1078)?

wjwwood (approved):

PR #339, opened 2022-12-20
Add in inconsistent topic defines and data structures.

clalancette:

Part of https://github.com/ros2/ros2/issues/1361


ivanpauno:
what's the equivalent event in DDS?

ivanpauno:
Found it: https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/INCONSISTENT_TOPIC_Status.htm

rmw/include/rmw/events_statuses/inconsistent_topic.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// Inconsistent topic information provided by a topic.
+typedef struct RMW_PUBLIC_TYPE rmw_inconsistent_topic_status_s
+{
+  /// Lifetime cumulative number of inconsistent topics detected.
```
ivanpauno, line 30: the definition for RTI connext is 

```
Total cumulative count of (DataReader, DataWriter) pairs whose topic names match the Topic to which this status is attached, but whose DDS types are inconsistent.
```
 
which is a bit confusing ...
I guess the other DDS vendors use the same (or similar) definition.

We would have to be more clear in the docs here

ivanpauno (approved):
LGTM, I only have one comment

rmw/include/rmw/events_statuses/inconsistent_topic.h
fujitatomoya: I am not sure how application uses this information? do we have use case???

fujitatomoya (approved):

rmw/include/rmw/events_statuses/inconsistent_topic.h
fujitatomoya: https://www.omg.org/spec/DDS/1.4/PDF says,

- total_count
  - Total cumulative count of the Topics discovered whose name matches the Topic to which this status is attached and whose type is inconsistent with the Topic.

although i understand how this works, not really sure how i can use it in the application.

ivanpauno: "count of topics" I guess it means "data writers and data readers".
e.g. if you have
1. data_reader topic_name="my_topic"; type="typeA"
2. data_writer topic_name="my_topic"; type="typeB"
3. data_reader topic_name="my_topic"; type="typeC"
4. data_writer topic_name="my_topic"; type="typeB"

`1.` and `3.` have an incompatible total count of 3, `2.` and `4.` of 2.
i.e. it counts both data writers and readers.

I'm also not sure if it counts incompatibilites of entities that are still alive, or if it counts all historical incompatibilities.
e.g. if you have a data reader that discovers an incompatible data writer. Then the data writer is killed, is the incompatibility count still one or zero?
It would be more useful if it's 0 IMO, but I'm not sure what DDS provides.

> although i understand how this works, not really sure how i can use it in the application.

Honestly, greater than 0 is bad, 0 is okay.
I think that the exact matter doesn't matter, it can only be useful for later debugging.

fujitatomoya: thanks for bringing up the example, that helps me to understand. i might be mistaken.

CC: @asorbini @MiguelCompany @eboasson 

> e.g. if you have
> 1. data_reader topic_name="my_topic"; type="typeA"
> 2. data_writer topic_name="my_topic"; type="typeB"
> 3. data_reader topic_name="my_topic"; type="typeC"
> 4. data_writer topic_name="my_topic"; type="typeB"

according to the following definitions,

- RTI connext

```
Total cumulative count of (DataReader, DataWriter) pairs whose topic names match the Topic to which this status is attached, but whose DDS types are inconsistent.
```

- DDS Spec 1.4

```bash
Total cumulative count of the Topics discovered whose name matches the Topic to which this status is attached and whose type is inconsistent with the Topic.
```

> 1. data_reader topic_name="my_topic"; type="typeA"

count `2` notified on this reader, since the topic is inconsistent with 2 writers "typeB" as `pair`. (2 topics are inconsistent)

> 2. data_writer topic_name="my_topic"; type="typeB"

count `2` notified on this writer, since the topic is inconsistent with readers "typeA" and "typeC" as `pair`.  (2 topics are inconsistent)

> 3. data_reader topic_name="my_topic"; type="typeC"

count `2` notified on this reader, since the topic is inconsistent with 2 writers "typeB" as `pair`. (2 topics are inconsistent)

> 4. data_writer topic_name="my_topic"; type="typeB"

count `2` notified on this writer, since the topic is inconsistent with readers "typeA" and "typeC" as `pair`.  (2 topics are inconsistent)

>I'm also not sure if it counts incompatibilites of entities that are still alive, or if it counts all historical incompatibilities.

i believe `total_count` is just incremental value when it detects inconsistent topics. so it keeps all history.

ivanpauno: That may be how it works.
I don't think that any of both definitions are crystal clear, so it would be nice to test and confirm.

fujitatomoya: @clalancette thanks for checking on this. for me, this makes sense to me.

methylDragon (approved):

PR #342, opened 2023-01-12
Change the system default depth to 1.

### LINKED ISSUE #341: Warning about history depth of 0 even while using system default QoS
## Bug report

**Required Info:**

- Operating System: Ubuntu 22.04
- Installation type: docker image from binaries
- DDS implementation: CycloneDDS
- Client library (if applicable): rclcpp

#### Steps to reproduce issue

```
pub_ = this->create_publisher<geometry_msgs::msg::Pose>("/pose", rclcpp::SystemDefaultsQoS());
```

#### Expected behavior
No QoS warning

#### Actual behavior
```
[node-7] [WARN] [1673487873.053699528] [rclcpp]: A zero depth with KEEP_LAST doesn't make sense; no data could be stored.This will be interpreted as SYSTEM_DEFAULT
```

#### Additional information
This warning was [recently](https://github.com/ros2/rclcpp/pull/2048) added. I would expect that if I use rclcpp::SystemDefaultsQoS() that I would not see any warning or error for QoS. I realize that the depth was probably always set to zero for the default and it was corrected by the rmw to 1 and only now the warning has been posted. But regardless, the expectation for a user is that the system default QoS would not generate a warning.

Secondarily, the warning itself is not very useful when my application has several publishers and subscriptions. It's probably not practical given the architecture of the code but I would find the warning message a lot more useful if the topic name was included.

clalancette:
This is what the RMWs were doing anyway, and setting it to a number here (and not an unknown of 0) means that we won't get warnings from the rclcpp and rclpy layers.

Fixes #341

### DIFF
--- a/rmw/include/rmw/qos_profiles.h
+++ b/rmw/include/rmw/qos_profiles.h
@@ -90,7 +90,7 @@ static const rmw_qos_profile_t rmw_qos_profile_parameter_events =
 static const rmw_qos_profile_t rmw_qos_profile_system_default =
 {
   RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT,
-  RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT,
+  1,
   RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT,
   RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT,
   RMW_QOS_DEADLINE_DEFAULT,


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17905/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12473/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18577/)

fujitatomoya (approved):
@clalancette it looks good for now, but according to https://docs.ros.org/en/rolling/Concepts/About-Quality-of-Service-Settings.html

```
Different RMW implementations may have different defaults.
```

that means, we should ask RMW implementation what the default value is?

clalancette:
> that means, we should ask RMW implementation what the default value is?

Hm, that's a good point.  Maybe this isn't the correct thing to do, then.  Maybe the correct thing to do is to special-case `SystemDefaultsQoS` to avoid the warning message, since in that case the user very specifically asked to use whatever the underlying middleware defaults are.

What do you think?

fujitatomoya:
> Maybe the correct thing to do is to special-case SystemDefaultsQoS to avoid the warning message, since in that case the user very specifically asked to use whatever the underlying middleware defaults are.

this sounds right to me. and then rmw implementation sets the implementation specific QoS depth accordingly. (e.g `rmw_fastftps` case, it will set `1` for depth.)

PR #348, opened 2023-02-14
Add type hash to rmw_topic_endpoint_info_t (rep2011)

emersonknapp:
Part of ros2/ros2#1159
Depends on ros2/rosidl#722

Adds `topic_type_hash` field to `rmw_topic_endpoint_info_t` so that it can be filled by RMW implementations.

### DIFF
--- a/rmw/include/rmw/topic_endpoint_info.h
+++ b/rmw/include/rmw/topic_endpoint_info.h
@@ -21,7 +21,6 @@ extern "C"
 #endif
 
 #include "rcutils/allocator.h"
-#include "rcutils/sha256.h"
 #include "rosidl_runtime_c/type_hash.h"
 #include "rmw/types.h"
 #include "rmw/visibility_control.h"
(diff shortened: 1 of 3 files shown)


emersonknapp:
@wjwwood @methylDragon @ivanpauno this is ready for review, the types are finalized in the dependency PR, we're only hashing out final details

rmw/include/rmw/topic_endpoint_info.h
```diff
@@ -135,6 +138,40 @@ rmw_topic_endpoint_info_set_topic_type(
   const char * topic_type,
   rcutils_allocator_t * allocator);
 
+/// Set the topic type hash in the given topic endpoint info data structure.
+/**
+ * This functions copies value `topic_type_hash` into data structure `topic_type_hash` member.
```
clalancette, line 143:
```suggestion
 * This function copies value `topic_type_hash` into data structure `topic_type_hash` member.
```

But actually, I don't actually understand what this is trying to tell me.  I guess it copies the data from the given `rosidl_type_hash_t` structure into the `rmw_topic_endpoint_info_t` structure member, correct?  It might be better to rewrite this sentence to say that.

clalancette (changes requested):
One minor thing to fix here, then I think this is good to go.

rmw/include/rmw/topic_endpoint_info.h
emersonknapp: Yeah sorry, copypasta result, I hadn't read it too closely, just changed out the names. I've done a pass to clean up the language on all the docstrings in this file.

clalancette (approved):
Looks good to me with green CI.

clalancette:
I had to rebase this to get the incompatible type stuff in.  I've done that now, here is another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18237/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12768/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18909/)

clalancette:
The test failures on Windows are known flakes.  So going ahead and merging this one in.

PR #349, opened 2023-03-02
Improve handling of dynamic discovery - rmw changes (Extends #338)

arjo129:
This PR extends #338 to support dynamic memory allocation.

### DIFF
--- a/rmw/include/rmw/discovery_options.h
+++ b/rmw/include/rmw/discovery_options.h
@@ -43,7 +43,8 @@ typedef enum RMW_PUBLIC_TYPE rmw_automatic_discovery_range_e
 #define RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH 256
 
 /// Struct to typedef some of the peer addresses
-typedef struct rmw_peer_address_s {
+typedef struct rmw_peer_address_s
+{
   char peer_address[RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH];
 } rmw_peer_address_t;
 
@@ -58,7 +59,7 @@ typedef struct RMW_PUBLIC_TYPE rmw_discovery_options_s
    * Each peer is specified as a hostname or an IP address (IPv4 and IPv6 are both acceptable), or
    * a subnet, e.g. 192.168.0.0/24.
    */
-  rmw_peer_address_t* static_peers;
+  rmw_peer_address_t * static_peers;
 
   /// The number of manually-specified peers
   size_t static_peers_count;
--- a/rmw/include/rmw/discovery_options.h
+++ b/rmw/include/rmw/discovery_options.h
@@ -71,18 +71,24 @@ RMW_WARN_UNUSED
 rmw_discovery_options_t
 rmw_get_zero_initialized_discovery_options(void);
 
-/// Compare two discovery parameter instances for equality. Equality means the
-/// automatic_discovery_range values are equal, they have the same
-/// static_peers_count value, and each entry in static_peers is evaluated as
-/// equal using strncmp.
-///
-/// NOTE: If the two parameter structs list the static peers in different orders
-/// then this will evaulate as NOT equal.
-///
-/// RMW_RET_OK will be returned when the input arguments are valid.
-/// RMW_RET_INVALID_ARGUMENT will be returned when any input is a nullptr,
-/// or if something in either struct was malformed, such as static_peers being
-/// a nullptr while static_peers_count is non-zero.
+/// Compare two discovery parameter instances for equality.
+/**
+ * Equality means the automatic_discovery_range values are equal, they have the same
+ * static_peers_count value, and each entry in static_peers is evaluated as
+ * equal using strncmp.
+ *
+ * NOTE: If the two parameter structs list the static peers in different orders
+ * then this will evaulate as NOT equal.
+ *
+ * \param[in] left - The first set of parameters to compare
+ * \param[in] right - The second set of parameters to compare
+ * \param[out] result - The result of the calculation.
+ *
+ * \return RMW_RET_OK when the input arguments are valid.
+ * \return RMW_RET_INVALID_ARGUMENT will be returned when any input is a nullptr,
+ * or if something in either struct was malformed, such as static_peers being
+ * a nullptr while static_peers_count is non-zero.
+ */
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_ret_t
@@ -92,9 +98,18 @@ rmw_discovery_options_equal(
   bool * result);
 
 /// Perform a deep copy of the discovery parameters from src into dst using the
-/// given allocator. The dst will be left with an owned copy of the static peers
-/// array whose string values match the src. If successful, src and dst will
-/// evaluate as equal using rmw_discovery_options_equal.
+/// given allocator.
+/**
+ * The dst will be left with an owned copy of the static peers
+ * array whose string values match the src. If successful, src and dst will
+ * evaluate as equal using rmw_discovery_options_equal.
+ * \param[in] src discovery option parameters to be copied.
+ * \param[in] allocator to use.
+ * \param[out] dst Destination parameters to use.
+ * \return RMW_RET_OK if success.
+ * \return RMW_RET_INVALID_ARGUMENT if either the src, allocator or dst is null.
+ * \return RMW_RET_BAD_ALLOC if allocation fails.
+ */
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_ret_t
@@ -103,6 +118,14 @@ rmw_discovery_options_copy(
   rcutils_allocator_t allocator,
   rmw_discovery_options_t * dst);
 
+/// Destructor for rmw_discovery_options_t
+/**
+ * \param[in] discovery_options to destroy
+ * \param[in] allocator to be used for destruction.
+ * \return RMW_RET_OK if success.
+ * \return RMW_RET_INVALID_ARGUMENT if allocator is invalid
+ * or discovery_options is null.
+ */
 RMW_PUBLIC
 RMW_WARN_UNUSED
 rmw_ret_t
--- a/rmw/include/rmw/discovery_options.h
+++ b/rmw/include/rmw/discovery_options.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef RMW__DISCOVERY_PARAMETERS_H_
-#define RMW__DISCOVERY_PARAMETERS_H_
+#ifndef RMW__DISCOVERY_OPTIONS_H_
+#define RMW__DISCOVERY_OPTIONS_H_
 
 #include "rcutils/allocator.h"
 
@@ -40,15 +40,15 @@ typedef enum RMW_PUBLIC_TYPE rmw_automatic_discovery_range_e
 } rmw_automatic_discovery_range_t;
 
 /// Maximum length of a peer hostname or IP address
-#define RMW_DISCOVERY_PARAMETERS_PEER_MAX_LENGTH 256
+#define RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH 256
 
 /// Struct to typedef some of the peer addresses
 typedef struct rmw_peer_address_s {
-  char peer_address[RMW_DISCOVERY_PARAMETERS_PEER_MAX_LENGTH];
+  char peer_address[RMW_DISCOVERY_OPTIONS_STATIC_PEERS_MAX_LENGTH];
 } rmw_peer_address_t;
 
 /// Used to specify the parameters that control how discovery is performed
-typedef struct RMW_PUBLIC_TYPE rmw_discovery_parameters_s
+typedef struct RMW_PUBLIC_TYPE rmw_discovery_options_s
 {
   /// How far to allow multicast to be used
   rmw_automatic_discovery_range_t automatic_discovery_range;
@@ -62,13 +62,13 @@ typedef struct RMW_PUBLIC_TYPE rmw_discovery_parameters_s
 
   /// The number of manually-specified peers
   size_t static_peers_count;
-} rmw_discovery_parameters_t;
+} rmw_discovery_options_t;
 
 /// Return a zero-initialized discovery parameters structure.
 RMW_PUBLIC
 RMW_WARN_UNUSED
-rmw_discovery_parameters_t
-rmw_get_zero_initialized_discovery_parameters(void);
+rmw_discovery_options_t
+rmw_get_zero_initialized_discovery_options(void);
 
 /// Compare two discovery parameter instances for equality. Equality means the
 /// automatic_discovery_range values are equal, they have the same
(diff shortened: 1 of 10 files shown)


rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+{
+  /// How far to allow multicast to be used
+  rmw_automatic_discovery_range_t automatic_discovery_range;
+
+  /// The list of manually-specified peers to perform static discovery with
+  /**
+   * Each peer is specified as a hostname or an IP address (IPv4 and IPv6 are both acceptable), or
+   * a subnet, e.g. 192.168.0.0/24.
+   */
+  peer_address_t* static_peers;
```
wjwwood, line 61: I think the style checks will catch this, but I think it will be:

```suggestion
  peer_address_t * static_peers;
```

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_discovery_params_t
+rmw_get_zero_initialized_discovery_params(void);
+
+/// Compare two discovery parameter instances for equality.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+bool
+rmw_discovery_params_equal(rmw_discovery_params_t * left, rmw_discovery_params_t * right);
```
wjwwood, line 77: What is the behavior of this function if either of the pointers are nullptr? I think it would be useful to document the expectations, and potentially we may need a return code with an "out" parameter for the comparison result. Similarly since the struct contains an array, what is the behavior when one of the two have invalid arrays (nullptr to array head)?

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+  /// The list of manually-specified peers to perform static discovery with
+  /**
+   * Each peer is specified as a hostname or an IP address (IPv4 and IPv6 are both acceptable), or
+   * a subnet, e.g. 192.168.0.0/24.
+   */
+  peer_address_t* static_peers;
+
+  /// The number of manually-specified peers
+  size_t static_peers_count;
+} rmw_discovery_params_t;
```
wjwwood, line 65: nitpick: we should use full names and avoid abbreviations for the benefit of folks who are translating this from English to something else, e.g. `rmw_discovery_parameters_t`.

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+rmw_discovery_params_equal(rmw_discovery_params_t * left, rmw_discovery_params_t * right);
+
+/// Copy a discovery parameter
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_discovery_params_copy(
+  const rmw_discovery_params_t * src,
+  const rcutils_allocator_t * allocator,
+  rmw_discovery_params_t * dst);
```
wjwwood, line 86: What is the behavior of this function? I assume it copies the array of static peers, using the allocator, but we should have it detail the things it will do during the copy.

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_discovery_params_fini(
+  rmw_discovery_params_t * discovery_params,
+  const rcutils_allocator_t * allocator);
```
wjwwood, line 93: In other places in the API we prefer to store the allocator in the data structure so that you don't need to pass it in the fini function, making it more or less impossible to accidentally pass the wrong allocator here. This has the disadvantage of making the data structure larger (by the size  of the allocator) but the advantage that the above issue I described and so that you don't have to check the allocator pointer here (might be nullptr).

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+bool
+rmw_discovery_params_equal(rmw_discovery_params_t * left, rmw_discovery_params_t * right);
+
+/// Copy a discovery parameter
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_ret_t
+rmw_discovery_params_copy(
+  const rmw_discovery_params_t * src,
+  const rcutils_allocator_t * allocator,
```
wjwwood, line 85: I think we sometimes pass the allocator as a pointer and other times as a value, which is unfortunately inconsistent of us, but my guiding rule is that if you will definitely use the allocator, e.g. ensured allocations will happen or you'll store the allocator in the data structure as part of init, then passing as a value is nice, so you don't have to validate the pointer to the allocator.

rmw/include/rmw/discovery_params.h
wjwwood: Also, not sure the allocator being const here makes sense. There's a philosophical debate about deep const-ness and expectation signaling, but in this case my expectation is that the conceptual "state" of the allocator will change when something is fini'ed (or may change) and therefore probably shouldn't be const, even though we're technically not changing any shallow values in the allocator struct itself.

rmw/include/rmw/impl/cpp/key_value.hpp
```diff
@@ -29,7 +29,7 @@ namespace cpp
 // TODO(karsten1987): Implement based on
 // https://github.com/PrismTech/opensplice/blob/master/docs/pdf/OpenSplice_refman_CPP.pdf
 static std::map<std::string, std::vector<uint8_t>>
-parse_key_value(std::vector<uint8_t> kv)
+parse_key_value(const std::vector<uint8_t>& kv)
```
wjwwood, line 32:
```suggestion
parse_key_value(const std::vector<uint8_t> & kv)
```

rmw/include/rmw/discovery_params.h
wjwwood: Same goes for the file name, `discovery_parameters.h` and related files too.

rmw/include/rmw/discovery_params.h
```diff
@@ -0,0 +1,98 @@
...
+  /// Forces multicast discovery off
+  RMW_AUTOMATIC_DISCOVERY_RANGE_OFF = 1,
+  /// Allows multicast but only on the localhost
+  RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST = 2,
+  /// Allows multicast on the reachable network
+  RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET = 3,
+} rmw_automatic_discovery_range_t;
+
+/// Maximum length of a peer hostname or IP address
+#define RMW_DISCOVERY_PARAMS_PEER_MAX_LENGTH 256
```
wjwwood, line 43: nitpick: use full name, e.g. `RMW_DISCOVERY_PARAMETERS_STATIC_PEER_MAX_LENGTH`, including "static" and the full "parameters" name.

rmw/include/rmw/init_options.h
```diff
@@ -54,6 +55,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_init_options_s
   rmw_security_options_t security_options;
   /// Enable localhost only
   rmw_localhost_only_t localhost_only;
+  /// Configure discovery
+  rmw_discovery_params_t discovery_params;
```
wjwwood, line 59: Kind of meta, but seeing it here, I'm wondering why "parameters" rather than "options"? Seems like `rmw_discovery_options_t` would be a better name. "parameters" is a kind of overloaded term in the ROS world, so avoiding it in favor of a more conventional and less overloaded term like "options" seems attractive to me.

rmw/src/discovery_params.c
```diff
@@ -0,0 +1,98 @@
...
+
+  dst->automatic_discovery_range = src->automatic_discovery_range;
+  dst->static_peers =
+      allocator->zero_allocate(
+        src->static_peers_count, 
+        sizeof(peer_address_t),
+        allocator->state);
+  for (size_t i = 0; i < src->static_peers_count; i++)
+  {
+    strncpy(
```
wjwwood, line 73: This is likely to produce warnings about not using the `_s` "safe" version of the function. Especially on Windows. Just be ready for that in the CI testing.

rmw/src/discovery_params.c
```diff
@@ -0,0 +1,98 @@
...
+rmw_ret_t
+rmw_discovery_params_fini(
+  rmw_discovery_params_t * discovery_params,
+  const rcutils_allocator_t * allocator)
+{
+  RMW_CHECK_ARGUMENT_FOR_NULL(discovery_params, RMW_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR(allocator, return RMW_RET_INVALID_ARGUMENT);
+
+  if (discovery_params->static_peers_count > 0)
+    allocator->deallocate(discovery_params->static_peers, allocator->state);
```
wjwwood, line 94: Use `{}`'s:

```suggestion
  if (discovery_params->static_peers_count > 0) {
    allocator->deallocate(discovery_params->static_peers, allocator->state);
  }
```

rmw/src/discovery_params.c
```diff
@@ -0,0 +1,98 @@
...
+  RMW_CHECK_ARGUMENT_FOR_NULL(dst, RMW_RET_INVALID_ARGUMENT);
+
+  dst->automatic_discovery_range = src->automatic_discovery_range;
+  dst->static_peers =
+      allocator->zero_allocate(
+        src->static_peers_count, 
+        sizeof(peer_address_t),
+        allocator->state);
+  for (size_t i = 0; i < src->static_peers_count; i++)
+  {
```
wjwwood, line 72: uncrustify should catch this (maybe it hasn't been run yet), but I think it should be:

```suggestion
  for (size_t i = 0; i < src->static_peers_count; i++) {
```

wjwwood (changes requested):
A few style and implementation questions, as well as some naming considerations, but otherwise looks reasonable to me.

One thing I'm unsure about still, and I'm sure I'll see this elsewhere, is how the static peers are managed? Seems like there should be functions to help modify that array within the discovery params, in order to help avoid mistakes. Similar to how we have helpers for using the array types in rcutils:

https://github.com/ros2/rcutils/blob/rolling/include/rcutils/types/char_array.h#L106-L240

I don't think we need all of those, but some functions to help with resizing and adding/removing would be good I think.

wjwwood:
Also we should consolidate this with https://github.com/ros2/rmw/pull/338 and address the feedback there as well.

rmw/include/rmw/discovery_params.h
mxgrey: These changes should address your concerns: [d63987b](https://github.com/ros2/rmw/pull/349/commits/d63987bb9501f8c119d146690ccc58801adcd092)

It's now using `rmw_ret_t` as a return value with a `bool *` output parameter. Passing in a nullptr for any argument gets a `RMW_RET_INVALID_ARGUMENT` return value. If either parameter struct has a nullptr array with a non-zero static peers count, that also triggers a `RMW_RET_INVALID_ARGUMENT`.

rmw/include/rmw/discovery_params.h
mxgrey: Names have been updated, including changing `peer_address` to `rmw_peer_address`: [62d0aba](https://github.com/ros2/rmw/pull/349/commits/62d0aba8b438e59e7e690d54403198c6ec6859f3)

rmw/include/rmw/discovery_params.h
mxgrey: Documentation updated: [3850bc3](https://github.com/ros2/rmw/pull/349/commits/3850bc3b3893ddd04778e278ae685c341db2a384)

rmw/include/rmw/discovery_params.h
mxgrey: > In other places in the API we prefer to store the allocator in the data structure

Based on downstream usage of this API in the various RMW implementations, I don't think there's any value in storing the allocator in the struct. All the different RMW implementations which are handling the `rmw_discovery_parameters` struct already have an allocator stored which they can use for this.

rmw/include/rmw/discovery_params.h
mxgrey: Also some more `params` -> `parameters`: [d96be10](https://github.com/ros2/rmw/pull/349/commits/d96be103a67cfeead6fbadec02a6c7b02ac90b3c)

rmw/include/rmw/discovery_params.h
mxgrey: Changed to value parameter: [245d6bc](https://github.com/ros2/rmw/pull/349/commits/245d6bc8b69b0097734d5ec90c156014af6825b9)

rmw/include/rmw/discovery_params.h
mxgrey: This concern is irrelevant now because it's being passed by value: [245d6bc](https://github.com/ros2/rmw/pull/349/commits/245d6bc8b69b0097734d5ec90c156014af6825b9)

rmw/include/rmw/init_options.h
mxgrey: I wish I had seen this before my earlier renaming iterations, but I agree.

[1d16081](https://github.com/ros2/rmw/pull/349/commits/1d16081bd4319b467b998460ae2d4201fb51c6d4) changes `discovery_parameters` to `discovery_options`.

rmw/include/rmw/discovery_params.h
mxgrey: Taken care of in [1d16081](https://github.com/ros2/rmw/pull/349/commits/1d16081bd4319b467b998460ae2d4201fb51c6d4), plus `parameters` -> `options`.

rmw/include/rmw/impl/cpp/key_value.hpp
mxgrey: Buried inside of [1d16081](https://github.com/ros2/rmw/pull/349/commits/1d16081bd4319b467b998460ae2d4201fb51c6d4)

rmw/src/discovery_params.c
mxgrey: It seems `strncpy_s` [was introduced in C11](https://en.cppreference.com/w/c/string/byte/strncpy) but we're currently using `C99` according to the `CMAKE_C_STANDARD` setting in the `CMakeLists.txt`.

Are we okay with bumping that to `C11`?

rmw/src/discovery_params.c
mxgrey: Fixed: [1152b9b](https://github.com/ros2/rmw/pull/349/commits/1152b9b1392488e0b7c74ec9213d30df82f023d4)

rmw/src/discovery_params.c
mxgrey: Fixed: [1152b9b](https://github.com/ros2/rmw/pull/349/commits/1152b9b1392488e0b7c74ec9213d30df82f023d4)

rmw/src/discovery_params.c
wjwwood: Either way, you'll have to do some conditional logic to support Windows. This is why we sometimes use the rctuils format string functions, rather than using `strncpy`, but either way is fine.

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,116 @@
...
+/// static_peers_count value, and each entry in static_peers is evaluated as
+/// equal using strncmp.
+///
+/// NOTE: If the two parameter structs list the static peers in different orders
+/// then this will evaulate as NOT equal.
+///
+/// RMW_RET_OK will be returned when the input arguments are valid.
+/// RMW_RET_INVALID_ARGUMENT will be returned when any input is a nullptr,
+/// or if something in either struct was malformed, such as static_peers being
+/// a nullptr while static_peers_count is non-zero.
```
wjwwood, line 85: nitpick: this doesn't match our doxygen style for the rest of the code in this package, specifically:

- new sentences start on new lines
- use of `///` only for the brief, `/**` blocks otherwise
- use of directives like `\return`

Example of a standard block:

https://github.com/ros2/rmw/blob/114e1e7ba9174b7dff8d10b82fdcb0b54ae5aff8/rmw/include/rmw/init_options.h#L74-L106

re: use one sentence per line

> - [any] Each sentence must start on a new line.
>   - Rationale: For longer paragraphs a single change in the beginning makes the diff unreadable since it carries forward through the whole paragraph.

https://docs.ros.org/en/rolling/Contributing/Code-Style-Language-Versions.html#markdown-restructured-text-docblocks

I'd reformat this one as:

```suggestion
/// Compare two discovery parameter instances for equality.
/**
 * Equality means the automatic_discovery_range values are equal,
 * they have the same static_peers_count value, and each entry in
 * static_peers is evaluated as equal using strncmp.
 *
 * NOTE: If the two parameter structs list the static peers in different orders
 * then this will evaulate as NOT equal.
 *
 * \return `RMW_RET_OK` when the input arguments are valid, or
 * \return `RMW_RET_INVALID_ARGUMENT` when any input is a nullptr, or
 *   if something in either struct was malformed, such as static_peers being
 *   a nullptr while static_peers_count is non-zero.
 */
```

rmw/include/rmw/discovery_options.h
```diff
@@ -0,0 +1,116 @@
...
+rmw_ret_t
+rmw_discovery_options_equal(
+  const rmw_discovery_options_t * const left,
+  const rmw_discovery_options_t * const right,
+  bool * result);
+
+/// Perform a deep copy of the discovery parameters from src into dst using the
+/// given allocator. The dst will be left with an owned copy of the static peers
+/// array whose string values match the src. If successful, src and dst will
+/// evaluate as equal using rmw_discovery_options_equal.
```
wjwwood, line 97: Same nitpick about style as the other, but also (not a nitpick) we should have documentation about the return codes that are possible and when they will occur.

wjwwood (commented):
small style comments, but otherwise lgtm

We should really close #338 or merge this one into it.

wjwwood:
Aside from a conversation about strncpy, I think all my review comments were addressed. I'm going to merge this with geoff's prs and continue there.

PR #358, opened 2023-10-20
Add `rmw_publisher_count_non_local_matched_subscriptions`

MiguelCompany:
Adds function that returns the number of matched subscriptions that are not in the same context as the publisher.

Part of https://github.com/ros2/rclcpp/issues/2202

full repos file [here](https://gist.github.com/MiguelCompany/c92eb293f97ad82ced4aa1f4b0c60d5e/raw/fe4f53f423377908cf2cebfd7bc666c3d67a356f/count_non_local_subscriptions_rolling.repos)

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -661,6 +661,35 @@ rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count);
 
+/// Retrieve the number of non local matched subscriptions to a publisher.
+/**
+ * Query the underlying middleware to determine how many non local subscriptions are
+ * matched to a given publisher.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | Maybe [1]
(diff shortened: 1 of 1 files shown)


rmw/include/rmw/rmw.h
```diff
@@ -661,6 +661,35 @@ rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count);
 
+/// Retrieve the number of non local matched subscriptions to a publisher.
+/**
+ * Query the underlying middleware to determine how many non local subscriptions are
+ * matched to a given publisher.
+ *
```
fujitatomoya, line 668: I think it would be better to define `what is local` means by here.

```suggestion
 * The definition of local means that in the same context.
 *
```

fujitatomoya (changes requested):

fujitatomoya:
maybe we can discuss this interface for next MW WG.

sloretz:
@MiguelCompany Friendly ping. Would you be willing to address @fujitatomoya's feedback?

fujitatomoya:
@MiguelCompany @wjwwood i would like to have a quick chat in next MW WG which is scheduled on 15th.Nov.2023.

PR #359, opened 2023-10-24
definition of local means being in the same context.

fujitatomoya:
Just a comment fix after https://github.com/ros2/design/pull/250.

Now `local` means by here is being in the same context. precisely with DDS-RTPS, GUID prefix matches.


fujitatomoya:
CC: @clalancette @wjwwood

sloretz:
@clalancette friendly waffle ping

clalancette (approved):

fujitatomoya:
@clalancette thanks for your review, since this is just a doc fix and Rpr is green, i will go ahead to merge this.

PR #364, opened 2024-03-19
Message Count API Creation

CursedRock17:
This PR is meant in preparation for [issue #1109](https://github.com/ros2/rcl/issues/1109) in the `rcl` repository by creating a `rmw` API for counting messages a `Subscription` is ready to receive.

### DIFF
--- a/rmw/include/rmw/rmw.h
+++ b/rmw/include/rmw/rmw.h
@@ -3041,6 +3041,43 @@ rmw_count_services(
   const char * service_name,
   size_t * count);
 
+/// Get the number of messages ready for a subscription
+/**
+ * This function returns the number of messages ready for a given subscriber
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@CursedRock17 thanks for creating PR.

IMO, this requires RMW change, i would suggest that before moving forward, probably it would be nice to have discussion on basic design describes how we could implement this underlying rmw implementation, including client interfaces? maybe [REP](https://github.com/ros-infrastructure/rep) required, i am not sure... @clalancette @wjwwood WDYT?

clalancette:
> IMO, this requires RMW change, i would suggest that before moving forward, probably it would be nice to have discussion on basic design describes how we could implement this underlying rmw implementation, including client interfaces? maybe [REP](https://github.com/ros-infrastructure/rep) required, i am not sure... @clalancette @wjwwood WDYT?

Yeah, agreed.  I'd like to see draft PRs, or at least documentation, on how we would implement this for `rmw_fastrtps`, `rmw_cyclonedds`, `rmw_connextdds`, and `rmw_zenoh`.

PR #369, opened 2024-06-02
add mingw support

felixf4xu:
this PR changes the `if` condition from system (`_WIN32`) to compiler (`_MSC_VER`), so that the **mingw** environment (which also defines `_WIN32`) can compile successfully.


fujitatomoya (commented):
i think this is visual studio compiler condition, so it seems okay to change.

@ros2/team what do you think?

mjcarroll (approved):

mjcarroll:
> what do you think?

This looks safe to me, and consistent with other applications that need to compile on Windows without using MSVC.  I wouldn't expect any impact, so LGTM with green CI.

clalancette:
I'm fine with the change, but I think we should target this to `rolling` first, and make sure things are fine there before backporting to `jazzy`.

felixf4xu:
@clalancette Thanks for the review. I'm fine if it's merged into `rolling` branch first.

Should I change the merge target or you can do it during the merge? (I'm not sure if the whole review process will be triggered again if I edit the PR at this time)

felixf4xu:
create a new PR #370 and close this one.

Issue #377, opened 2024-09-08
rmw_get_zero_initialized_names_and_types has data race in multi thread environment


comicfans:
## Bug report

**Required Info:**

- Operating System:
ubuntu 22.04
- Installation type:
source
- Version or commit hash:
  6.1.1
- DDS implementation:
  don't know
- Client library (if applicable):
 rclcpp

#### Steps to reproduce issue

Thread Sanitizer report 
https://github.com/ros2/rmw/blob/ac4f9afccca3dc305f2b519fdb0fb9131250f2a5/rmw/src/names_and_types.c#L31
```c
rmw_names_and_types_t
rmw_get_zero_initialized_names_and_types(void)
{
  static rmw_names_and_types_t zero = {
    .names = {0, NULL, {NULL, NULL, NULL, NULL, NULL}},
    .types = NULL,
  };
  zero.names = rcutils_get_zero_initialized_string_array();
  return zero;
}

```
zero.names already inited, but later being re-assigned during runtime,while accessing from different thread, this is a data race. similar problem here https://github.com/ros2/rcutils/pull/478

#### Additional information

----

PR #378, opened 2024-09-09
Initialize the NULL strucutre with static value.

fujitatomoya:
closes https://github.com/ros2/rmw/issues/377

### DIFF
--- a/rmw/src/discovery_options.c
+++ b/rmw/src/discovery_options.c
@@ -21,10 +21,8 @@
 rmw_discovery_options_t
 rmw_get_zero_initialized_discovery_options(void)
 {
-  static const rmw_discovery_options_t result = {
-    .automatic_discovery_range = RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET,
-    .static_peers_count = 0,
-  };
+  // All members are initialized to 0 or NULL by C99 6.7.8/10.
+  static const rmw_discovery_options_t result;
   return result;
 }
 
--- a/rmw/src/discovery_options.c
+++ b/rmw/src/discovery_options.c
@@ -21,10 +21,10 @@
 rmw_discovery_options_t
 rmw_get_zero_initialized_discovery_options(void)
 {
-  rmw_discovery_options_t result = (rmw_discovery_options_t) {
+  static const rmw_discovery_options_t result = {
     .automatic_discovery_range = RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET,
     .static_peers_count = 0,
-  };  // NOLINT(readability/braces): false positive
+  };
   return result;
 }
 
--- a/rmw/src/event.c
+++ b/rmw/src/event.c
@@ -24,11 +24,13 @@ extern "C" {
 rmw_event_t
 rmw_get_zero_initialized_event(void)
 {
-  const rmw_event_t event = {
+  // TODO(@fujitatomoya): This is not exatly zero initialized structure.
+  /// We should introduce xxx_get_default_event to return the default values.
+  static const rmw_event_t event = {
     .implementation_identifier = NULL,
     .data = NULL,
     .event_type = RMW_EVENT_INVALID
-  };  // NOLINT(readability/braces): false positive
+  };
   return event;
 }
 
--- a/rmw/src/names_and_types.c
+++ b/rmw/src/names_and_types.c
@@ -24,10 +24,8 @@
 rmw_names_and_types_t
 rmw_get_zero_initialized_names_and_types(void)
 {
-  static rmw_names_and_types_t zero = {
-    .names = {0, NULL, {NULL, NULL, NULL, NULL, NULL}},
-    .types = NULL,
-  };
+  // All members are initialized to 0 or NULL by C99 6.7.8/10.
+  static const rmw_names_and_types_t zero;
   return zero;
 }
 
--- a/rmw/src/names_and_types.c
+++ b/rmw/src/names_and_types.c
@@ -28,7 +28,6 @@ rmw_get_zero_initialized_names_and_types(void)
     .names = {0, NULL, {NULL, NULL, NULL, NULL, NULL}},
     .types = NULL,
   };
-  zero.names = rcutils_get_zero_initialized_string_array();
   return zero;
 }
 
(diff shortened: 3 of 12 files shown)


fujitatomoya:
@comicfans could you try this patch to see if it fixes your asan warning?

methylDragon:
Makes sense to me, since the the struct is nested.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21820/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16090/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1322/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22520/)

methylDragon (approved):

clalancette (changes requested):
So looking through `rmw` here, it looks like quite a few of the `get_zero_initialized` functions don't follow this pattern.  I'm thinking we should update all of them in this PR, what do you think @fujitatomoya ?

fujitatomoya:
@clalancette sounds good to me, let me try to replace them all.

rmw/src/names_and_types.c
```diff
@@ -28,7 +28,6 @@ rmw_get_zero_initialized_names_and_types(void)
     .names = {0, NULL, {NULL, NULL, NULL, NULL, NULL}},
     .types = NULL,
   };
-  zero.names = rcutils_get_zero_initialized_string_array();
```
fujitatomoya, line 31: Now i think opposite against this change.

maybe we should call `rcutils_get_zero_initialized_string_array` for both `names` and `types` to initialize those fields with allocating memory in stack? unless, once `rcutils_string_array_t` is modified, we need to update this code too.
IMO, correct way is to use stack memory and return the initialized values?

fujitatomoya: @methylDragon what do you think above?

comicfans:
> @comicfans could you try this patch to see if it fixes your asan warning?

confirm this fix the problem

rmw/src/discovery_options.c
```diff
@@ -21,10 +21,10 @@
 rmw_discovery_options_t
 rmw_get_zero_initialized_discovery_options(void)
 {
-  rmw_discovery_options_t result = (rmw_discovery_options_t) {
+  static const rmw_discovery_options_t result = {
```
fujitatomoya, line 24: use `static const` to have this in the code read-only segment so that we can use code cache and avoid runtime memory allocation.

rmw/src/event.c
```diff
@@ -24,11 +24,13 @@ extern "C" {
 rmw_event_t
 rmw_get_zero_initialized_event(void)
 {
-  const rmw_event_t event = {
+  // TODO(@fujitatomoya): This is not exatly zero initialized structure.
+  /// We should introduce xxx_get_default_event to return the default values.
```
fujitatomoya, line 28: some xxx_get_zero_initialized_yyy functions have default values which are not zero initialized. i think what we need to do here is to call zero initialized function 1st and then get the default values for the structure. i would like to process one thing at a time, so this `TODO` comment is added as follow-up item.

fujitatomoya (commented):
@methylDragon @clalancette what do you think? could you take a look when you have time?

fujitatomoya:
@clalancette can you take a look at this?

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21831/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16101/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1333/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22537/)

rmw/src/discovery_options.c
```diff
@@ -21,10 +21,10 @@
...
     .automatic_discovery_range = RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET,
     .static_peers_count = 0,
```
clalancette, line 26: So actually, RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET is 0, so we can use the universal specifier here if we want:

```suggestion
  // All members are initialized to 0 or NULL by C99 6.7.8/10.
  static const rmw_discovery_options_t result;
```

rmw/src/discovery_options.c
```diff
@@ -21,10 +21,10 @@
...
-  };  // NOLINT(readability/braces): false positive
+  };
```
clalancette, line 27:
```suggestion
```

clalancette (changes requested):
One minor change, then I'm happy with this.

rmw/src/discovery_options.c
fujitatomoya: i was thinking that once `RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET` enum is updated to another value in the future, this breaks easily. (that i think we should do that but guaranteed.) So that was my intention to leave this here. (i guess original author though about this in the same way)

clalancette: While that is true, that shouldn't matter here.  This is the `get_zero_initialized` function, so it should always return 0 (in my opinion).

fujitatomoya: yeah `RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET ` should have been used with `xxx_get_default_option`. that is what i will do with the another patch, i am okay with 0 init then.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21834/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16104/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1336/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22543/)

fujitatomoya:
@clalancette all comments are resolved, lgtm?

clalancette (approved):

fujitatomoya:
RHEL is unstable, https://ci.ros2.org/job/ci_linux-rhel/1336/testReport/.

* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1341/)

PR #379, opened 2024-09-18
add rmw_get_default_xxx to return the structure with default values.

fujitatomoya:
follow up for https://github.com/ros2/rmw/pull/378

- rmw_get_zero_initialized_xxx should return zero initialized data structure without any default non-zero values. zero initialized data is mostly used for copying the data object (allocating destination data structure) and finalization.
- rmw_get_default_xxx should return the data structure with default values, so that application can use those default values to process the initialization.

### DIFF
--- a/rmw/include/rmw/event.h
+++ b/rmw/include/rmw/event.h
@@ -68,6 +68,12 @@ RMW_WARN_UNUSED
 rmw_event_t
 rmw_get_zero_initialized_event(void);
 
+/// Return a default event structure.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_event_t
+rmw_get_default_event(void);
+
 /// Initialize a rmw publisher event.
 /**
  * \param[inout] rmw_event to initialize
--- a/rmw/include/rmw/init.h
+++ b/rmw/include/rmw/init.h
@@ -55,6 +55,12 @@ RMW_WARN_UNUSED
 rmw_context_t
 rmw_get_zero_initialized_context(void);
 
+/// Return a default context structure.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_context_t
+rmw_get_default_context(void);
+
 /// Initialize the middleware with the given options, and yielding an context.
 /**
  * Context is filled with middleware specific data upon success of this function.
--- a/rmw/include/rmw/init_options.h
+++ b/rmw/include/rmw/init_options.h
@@ -71,6 +71,12 @@ RMW_WARN_UNUSED
 rmw_init_options_t
 rmw_get_zero_initialized_init_options(void);
 
+/// Return a default init options structure.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_init_options_t
+rmw_get_default_init_options(void);
+
 /// Initialize given init options with the default values and implementation specific values.
 /**
  * The given allocator is used, if required, during setup of the init options,
--- a/rmw/src/event.c
+++ b/rmw/src/event.c
@@ -24,11 +24,15 @@ extern "C" {
 rmw_event_t
 rmw_get_zero_initialized_event(void)
 {
-  // TODO(@fujitatomoya): This is not exatly zero initialized structure.
-  /// We should introduce xxx_get_default_event to return the default values.
+  // All members are initialized to 0 or NULL by C99 6.7.8/10.
+  static const rmw_event_t event;
+  return event;
+}
+
+rmw_event_t
+rmw_get_default_event(void)
+{
   static const rmw_event_t event = {
-    .implementation_identifier = NULL,
-    .data = NULL,
     .event_type = RMW_EVENT_INVALID
   };
   return event;
--- a/rmw/src/init.c
+++ b/rmw/src/init.c
@@ -26,10 +26,18 @@ extern "C"
 rmw_context_t
 rmw_get_zero_initialized_context(void)
 {
+  // All members are initialized to 0 or NULL by C99 6.7.8/10.
+  static const rmw_context_t context;
+  return context;
+}
+
+rmw_context_t
+rmw_get_default_context(void)
+{
   return (const rmw_context_t) {
            .instance_id = 0,
            .implementation_identifier = NULL,
-           .options = rmw_get_zero_initialized_init_options(),
+           .options = rmw_get_default_init_options(),
            .actual_domain_id = 0u,
            .impl = NULL
   };  // NOLINT(readability/braces): false positive
--- a/rmw/src/init_options.c
+++ b/rmw/src/init_options.c
@@ -25,8 +25,14 @@ extern "C"
 rmw_init_options_t
 rmw_get_zero_initialized_init_options(void)
 {
-  // TODO(@fujitatomoya): This is not exatly zero initialized structure.
-  /// We should introduce xxx_get_default_init_optionst to return the default values.
+  // All members are initialized to 0 or NULL by C99 6.7.8/10.
+  static const rmw_init_options_t init_option;
+  return init_option;
+}
+
+rmw_init_options_t
+rmw_get_default_init_options(void)
+{
   static const rmw_init_options_t init_option = {
     .domain_id = RMW_DEFAULT_DOMAIN_ID,
     .discovery_options = {RMW_AUTOMATIC_DISCOVERY_RANGE_NOT_SET, 0},
--- a/rmw/test/test_event.cpp
+++ b/rmw/test/test_event.cpp
@@ -20,6 +20,14 @@ TEST(rmw_event, get_zero_initialized_event)
   const rmw_event_t actual = rmw_get_zero_initialized_event();
   EXPECT_EQ(nullptr, actual.implementation_identifier);
   EXPECT_EQ(nullptr, actual.data);
+  EXPECT_EQ(0u, actual.event_type);
+}
+
+TEST(rmw_event, get_default_event)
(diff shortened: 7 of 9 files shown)


fujitatomoya:
Pulls: ros2/rmw#379
Gist: https://gist.githubusercontent.com/fujitatomoya/65e549d1cfbf47cfcfd10df7e86bd521/raw/5a23d57d187e0028fde4536c468462cbe0f56501/ros2.repos
BUILD args:  --packages-above-and-dependencies rmw rcl rmw_implementation rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_connextdds_common rmw_cyclonedds_cpp
TEST args:  --packages-above rmw rcl rmw_implementation rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_connextdds_common rmw_cyclonedds_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14581
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21858/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16128/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1364/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22569/)

clalancette:
Just as an FYI, I'm not entirely sure we need these `get_default` APIs.  I haven't looked at the RMW APIs, but we often have `init`-style APIs which tend to init them already; maybe we can lean more heavily on those?  I'd need to look closer at the RMW APIs to see if that makes sense here.

methylDragon (approved):
I'm not opposed in principle to these since this is an avenue for performance boosting, and I see in the associated PRs that they're already being used.

Pending green CI, of course.

fujitatomoya:
> but we often have init-style APIs which tend to init them already; maybe we can lean more heavily on those?

i thought that too. but i think zero-init is also useful for data copy and finalization. so i do not think those are exclusive APIs.

fujitatomoya:
CI is failure to build, i think i misunderstood the usage of https://github.com/ros-tooling/ros-github-scripts?tab=readme-ov-file#ros-ci-for-pr

fujitatomoya:
Pulls: ros2/rmw#379, ros2/rcl#1186, ros2/rmw_implementation#241, ros2/rmw_fastrtps#778, ros2/rmw_connextdds#160, ros2/rmw_cyclonedds#513
Gist: https://gist.githubusercontent.com/fujitatomoya/18370bf1e0640d41a76698d03ad74dee/raw/7c320f893c82a71b0be07d942e42f59cb7e1df08/ros2.repos
BUILD args:  --packages-above-and-dependencies rmw rcl rmw_implementation rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_connextdds_common rmw_cyclonedds_cpp
TEST args:  --packages-above rmw rcl rmw_implementation rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_connextdds_common rmw_cyclonedds_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14583
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21861/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16131/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1367/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22572/)

rmw/include/rmw/event.h
```diff
@@ -68,6 +68,12 @@ RMW_WARN_UNUSED
 rmw_event_t
 rmw_get_zero_initialized_event(void);
 
+/// Return a default event structure.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_event_t
+rmw_get_default_event(void);
```
clalancette, line 75: For instance, I don't believe we need this.  In all "real" situations (not tests) where `rmw_get_zero_initialized_event` is called, it is immediately followed by either `rmw_subscription_event_init` or `rmw_publisher_event_init`.  Because of that, I don't believe we need to add this API at all (though if we change `rmw_get_zero_initialized_event` to truly return all zeros, we may have to fix a couple of tests).

rmw/include/rmw/init.h
```diff
@@ -55,6 +55,12 @@ RMW_WARN_UNUSED
 rmw_context_t
 rmw_get_zero_initialized_context(void);
 
+/// Return a default context structure.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_context_t
+rmw_get_default_context(void);
```
clalancette, line 62: Same here; all places we call `rmw_get_zero_initialized_context` we are either cleaning up after a failure, doing fini on the context, or this is immediately followed by an `rmw_init` which initializes the context.  Again, we may have to fix some tests if we make `rmw_get_zero_initialized_context` actually return all zeros.

rmw/include/rmw/init_options.h
```diff
@@ -71,6 +71,12 @@ RMW_WARN_UNUSED
 rmw_init_options_t
 rmw_get_zero_initialized_init_options(void);
 
+/// Return a default init options structure.
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rmw_init_options_t
+rmw_get_default_init_options(void);
+
```
clalancette, line 79: And the same here.

clalancette (changes requested):
I've left some comments inline on why I don't think we actually need to add these functions to the RMW interface.  We can probably just get by with changing the `zero_initialized` functions to actually zero initialize, and maybe fix some tests here and there.

rmw/include/rmw/init_options.h
fujitatomoya: @clalancette i take this case as an example. `RMW_DEFAULT_DOMAIN_ID` is not zero, so i was wondering that it should be returned by `rmw_get_default_init_options` originally. if we truly return the zero initialized option, this will affect the user application. i think we can have some options,

- `get_zero_initialized_xxx` returns with default values. (almost current implementation) IMO this does not sound right, because what it does is not really what it sounds like... `initialized_xxx` would sound better.
- `get_zero_initialized_xxx` truly returns with zero initialized fields, and provides APIs to get default values too. (these PRs)
- `get_zero_initialized_xxx` truly returns with zero initialized fields, and caller is responsible to fill in default values. (this does not really make sense, this is exactly whey `get_default_xxx` suggested.)

clalancette: I definitely think that `get_zero_initialized_xxx` should return a zero-initialized structure.  So we are in agreement there.

But my main point is that all callers of `get_zero_initialized_xxx` (with the exception of maybe a few tests) immediately follow that up with a corresponding `init_xxx` call, which basically sets the defaults.  For instance, considering the `rmw_init_options_t` structure, we have `rmw_get_zero_initialized_init_options`, which I believe should return a zero'ed structure.  But all callers immediately follow that with a call to [rmw_init_options_init](https://github.com/ros2/rmw/blob/f4105b162277cdec8ad62e47cbf52aead073d277/rmw/include/rmw/init_options.h#L110).  That function is implemented by the various RMWs:

* [rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/blob/565bbc22357413aa7342e9dd0d6518b71f7f10a6/rmw_fastrtps_shared_cpp/src/rmw_init.cpp#L28-L48) - initializes the structure very much like your proposed change here
* [rmw_cyclonedds](https://github.com/ros2/rmw_cyclonedds/blob/2d99e22538329138755a03649ec1c63d5f161352/rmw_cyclonedds_cpp/src/rmw_node.cpp#L759-L778) - initializes the structure very much like your proposed change here
* [rmw_connextdds](https://github.com/ros2/rmw_connextdds/blob/500baaa0d9032a51335e349023b363c093680ca0/rmw_connextdds_common/src/common/rmw_context.cpp#L925-L945) - initializes the structure very much like your proposed change here
* [rmw_zenoh](https://github.com/ros2/rmw_zenoh/blob/25e51f2729129d11c77d531cfebf893cd8d03a6b/rmw_zenoh_cpp/src/rmw_init_options.cpp#L34-L54) - initializes the structure very much like your proposed change here

Thus, it doesn't seem like the new APIs are necessary, as we already have all of the functionality we need here.

To maybe ask this in a different way: in what situations are you expecting that `rmw_get_default_init_options` will be used?

fujitatomoya: > I definitely think that get_zero_initialized_xxx should return a zero-initialized structure. So we are in agreement there.

:+1: 

> To maybe ask this in a different way: in what situations are you expecting that rmw_get_default_init_options will be used?

ah, i see now. i was thinking `rmw_get_default_init_options` returns implementation agnostic default option from ROS 2 RMW (e.g ROS 2 default domain ID), and `rmw_init_options_init` initializes the field with implementation specific values. but your point is that is not how it is implemented, so calling implementation `rmw_init_options_init` also initializes with default values.

i can check the related APIs to make sure, and adjust the PR accordingly. thanks!

fujitatomoya:
after the discussion, i will go ahead to close all these PRs. and reopen only rmw one to change `get_zero_initialized_xxx` returns a zero-initialized structure. (test might need to be updated, and rmw implementation needs to initialize the default value, but we do not need to have `get_default_xxx` for these.)

PR #419, opened 2026-06-04
Add rosidl buffer backend metadata to topic endpoint info

nvcyc:
## Description

Add rosidl buffer backend metadata to `rmw_topic_endpoint_info_t` so graph introspection APIs can carry backend support information for publishers and subscriptions.

### Is this user-facing behavior change?

Yes. This enables higher-level ROS graph tools to expose buffer backend support information for topic endpoints.

### Did you use Generative AI?

Yes. GPT-5.5 in Cursor was used to help draft changes in this pull request.

### Additional Information

This PR only adds the RMW-level data model.
RMW implementations and client libraries need separate changes to populate and expose the metadata.

### DIFF
--- a/rmw/include/rmw/topic_endpoint_info.h
+++ b/rmw/include/rmw/topic_endpoint_info.h
@@ -21,6 +21,7 @@ extern "C"
 #endif
 
 #include "rcutils/allocator.h"
+#include "rcutils/types/string_map.h"
 #include "rosidl_runtime_c/type_hash.h"
 #include "rmw/types.h"
 #include "rmw/visibility_control.h"
@@ -44,6 +45,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_topic_endpoint_info_s
   uint8_t endpoint_gid[RMW_GID_STORAGE_SIZE];
   /// QoS profile of the endpoint
   rmw_qos_profile_t qos_profile;
+  /// Buffer backend metadata advertised for this endpoint
+  rcutils_string_map_t buffer_backend_metadata;
 } rmw_topic_endpoint_info_t;
 
 /// Return zero initialized topic endpoint info data structure.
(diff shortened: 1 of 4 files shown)


Yadunund:
@nvcyc is there a test in `ros2cli` that we need to update as well? 

See test failures across RMW impls: https://ci.ros2.org/job/ci_linux/29339/testReport/junit/ros2topic.ros2topic.test/test_cli/test_cli/

rmw/include/rmw/topic_endpoint_info.h
```diff
@@ -44,6 +45,8 @@ typedef struct RMW_PUBLIC_TYPE rmw_topic_endpoint_info_s
   uint8_t endpoint_gid[RMW_GID_STORAGE_SIZE];
   /// QoS profile of the endpoint
   rmw_qos_profile_t qos_profile;
+  /// Buffer backend metadata advertised for this endpoint
+  rcutils_string_map_t buffer_backend_metadata;
```
fujitatomoya: the struct previously held only const char * owning pointers, but it now holds a rcutils_string_map_t whose impl is a heap pointer with its own fini. any code path that copies an rmw_topic_endpoint_info_t by plain struct assignment or memcpy now aliases that impl between two structs.
the existing const char * members have the same hazard in principle, but the codebase already established a deep-copy discipline for strings.
i'd suggest a rmw_topic_endpoint_info deep-copy helper, or at minimum confirmation that no array/element copy in the current code and companion PRs.

nvcyc: Thanks for pointing this issue out.

I agreed that adding a field of type `rcutils_string_map_t` in `rmw_topic_endpoint_info_t` is exposing more possible failure surface and requires extra care when handling `rmw_topic_endpoint_info_t`.

For this reason, I'm updating `rmw_topic_endpoint_info_t` to host the newly added `buffer_backend_metadata` field with `const char *` type, which is consistent to what most other fields in `rmw_topic_endpoint_info_t` already use and hence not broadening the risk surface.

Corresponding serializer and parser functions are added to store and and get the rosidl buffer metadata.
Relevant PRs are also updated accordingly for this change.

